PR #2, opened 2017-03-29
Turn off logging

sdogruyol:
`route.cr` don't have logging and it's unfair to compare it Kemal which has logging on by default. This PR turns off Kemal's logging thus leads to more fair comparison. Also removed redundant status code


tbrand:
Thanks for these patches!
I've already thought about that and updated them.
Please wait next update:smile:

PR #3, opened 2017-03-29
Turn off logging and remove redundant code

sdogruyol:
route.cr don't have logging and it's unfair to compare it Kemal which has logging on by default. This PR turns off Kemal's logging thus leads to more fair comparison. Also removed redundant status code

### DIFF
--- a/src/server_kemal.cr
+++ b/src/server_kemal.cr
@@ -1,17 +1,16 @@
 require "kemal"
 
+logging false
+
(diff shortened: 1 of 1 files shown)


tbrand:
Actually I've already changed my local code like that. But I'm very happy about this contribute, so I merge this. :smile:

sdogruyol:
ありがとう

PR #4, opened 2017-04-02
Fix Sinatra

adam12:
The current Sinatra example uses Webrick for a web server and the development environment.


tbrand:
Thanks for this! I'll update the result soon :)

PR #10, opened 2017-05-01
Add release to shard build

faustinoaq:
Add `--release` to shards build like Rust/Cargo

shards allow to append custom options to build command

https://github.com/crystal-lang/shards/blob/master/src/commands/build.cr#L33


tbrand:
OMG...:scream:
I've drop the option at somewhere... thanks!

PR #12, opened 2017-05-02
Implement hyper

### LINKED ISSUE #9: Framework request: hyper 0.11 (with Tokio)
I'm filing an issue rather than sending a PR here because hyper 0.11 isn't _quite_ out, but is close.

Currently, all the Rust frameworks are based on synchronous I/O. Hyper 0.11 will include [Tokio](https://tokio.rs/) support, which should put up even better numbers.

I'm happy to actually send a PR, either if "using hyper master" is okay, or once it's released.

steveklabnik:
Fixes #9

I only verified that this compiles; I'm on Windows right now and so not only do I not have `make` installed, but I also don't have Crystal installed. Please triple check things before merging!


tbrand:
That's ok! I'll check this later. Thanks for this PR. :)

rust/hyper/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+                Response::new()
+                    .with_status(StatusCode::NotFound)
+            }
+        })
+    }
+}
+
+
+fn main() {
+    let addr = "127.0.0.1:1337".parse().unwrap();
```
tbrand: Please change the bind port to 3000.

rust/hyper/src/main.rs
```diff
@@ -0,0 +1,48 @@
...
+
+    let server = Http::new().bind(&addr, || Ok(Echo)).unwrap();
+    println!("Listening on http://{} with 1 thread.", server.local_addr().unwrap());
```
tbrand: Please remove this line.

tbrand:
Sorry, it seems that hyper has no function to get path parameter.
I don't like this. https://github.com/tbrand/which_is_the_fastest/pull/12/files#diff-f9f7b6a88d7f0e2b3a931f034325167aR27
So I can't accept this PR.

If there is another way to get path parameters, please tell me :)

steveklabnik:
> If there is another way to get path parameters, please tell me :)

Hyper is an HTTP library, it has no concept of paths or routing, because neither does HTTP. Frameworks built on top of it might, though.

You should add this to the list of requirements, then 😄

tbrand:
Ok, I will. Thanks. :smile:

PR #19, opened 2017-05-07
Add rust/rocket

kiliankoe:
This PR adds [Rocket](https://rocket.rs) to the list of Rust frameworks.

Running the client seems to work fine. I haven't run the benchmarker though.

Only downside is that Rocket currently requires nightly rust to be built because of feature flags. Not sure if this is a deal-breaker for you, but it works just fine otherwise.


kiliankoe:
If you want it might also be an option to specify the rust toolchain via the make and neph targets. It would probably rely on the user having rust installed via rustup, but since that's how most people install rust that should be a given.

README.md
```diff
@@ -20,6 +20,7 @@
    - [IRON](https://github.com/iron/iron)
    - [nickel.rs](https://github.com/nickel-org/nickel.rs)
  - node
+   - [Rocket](https://rocket.rs)
```
tbrand: Please fix here.

kiliankoe: Oh whoops, how did that happen? o.O

tbrand:
Thanks for this!!:tada:

kiliankoe:
Thanks for the merge 😊

PR #22, opened 2017-05-07
Add Swift frameworks

kiliankoe:
Hope you don't mind I just threw all three together into a single PR. Vapor, Perfect and Kitura. All build and run fine. 

I also added make and neph build targets and threw everything into the benchmarker as well.

### DIFF
--- a/swift/vapor/Sources/main.swift
+++ b/swift/vapor/Sources/main.swift
@@ -1,0 +1,19 @@
+import Vapor
+
+let drop = Droplet(environment: .production)
+
+drop.get("/") { _ in
+    return ""
+}
+
+drop.get("user", String.self) { _, userID in
+    return userID
+}
+
+drop.post("user") { _ in
+    return ""
+}
+
+drop.run(servers: [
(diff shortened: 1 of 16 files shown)


tbrand:
> Hope you don't mind I just threw all three together into a single PR

NP :smile:
I'll review this, thanks!!

swift/vapor/Sources/main.swift
```diff
@@ -0,0 +1,19 @@
+import Vapor
+
+let drop = Droplet(environment: .production)
```
kiliankoe: On looking over this again I'm pretty sure stating the environment here is unnecessary. As far as I can tell this only applies to multiple config environments being present.

kiliankoe: Nevermind, without this vapor falls back to default logging, which it shouldn't for this use case.

tbrand:
It works very fine!!! Perfect!!!

kiliankoe:
Thanks 😊

PR #26, opened 2017-05-09
Don't needlessly allocate in Rocket code

SergioBenitez:


SergioBenitez:
Just wanted to show the potential change here. Feel free to implement it if you'd like. :)

kiliankoe:
Makes sense! 🙈

Any possible improvements should definitely be added, I'm all for adding this change.

tbrand:
> Any possible improvements should definitely be added

Absolutely💯
Thanks @SergioBenitez !!!

SergioBenitez:
Have the benchmarks been rerun with this? I've run the benchmarks locally, and I see timing on the order of Iron/Nickel, not the 2.5x worse value I see now.

kiliankoe:
[Others](https://github.com/tbrand/which_is_the_fastest/pull/32#issuecomment-300272507) have noticed a certain discrepancy regarding rocket's stats as well, but I don't think anybody knows why it differs. On my machine it's definitely comparable to iron and nickel as well.

PR #28, opened 2017-05-09
Optimize rocket

kiliankoe:
Pulling over @SergioBenitez's code from #26 (for easy one-click-merging^^) and updating Rocket's dependencies along the way.


tbrand:
Great!!

PR #30, opened 2017-05-09
add python/sanic server

sharkspeed:
add a python server : sanic

### DIFF
--- a/python/sanic/server_python_sanic.py
+++ b/python/sanic/server_python_sanic.py
@@ -1,0 +1,25 @@
+#!/usr/bin/env python3.5
+# -*- coding: utf-8 -*-
+
+from sanic import Sanic 
+from sanic.response import text
+
+app = Sanic()
+
+@app.route("/")
+async def index(request):
+    return text('')
+
+
+@app.route("/user/<id:int>", methods=['GET'])
+async def user_info(request, id):
+    return text(str(id))
+
+
+@app.route("/user", methods=['POST'])
+async def user(request):
+    return  text('')
+
+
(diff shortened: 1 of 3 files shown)


tbrand:
Awesome, I'll try this! Thanks!

python/sanic/server_python_sanic.py
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env python3.5
```
tbrand: Is the python 3.x required for sanic?

python/sanic/server_python_sanic.py
```diff
@@ -0,0 +1,25 @@
...
+# -*- coding: utf-8 -*-
+
+from sanic import Sanic
```
tbrand: When I try this, I got following error.
May be you should put requirement.txt at the directory and install sanic in Makefile?
```
ImportError: No module named 'sanic'
```

python/sanic/server_python_sanic.py
sharkspeed: yeah. Python 3.5+ is required.

python/sanic/server_python_sanic.py
sharkspeed: Ok, I will add it.

tbrand:
Great! It works!!:tada:

tbrand:
As I mentioned here https://github.com/tbrand/which_is_the_fastest/pull/34, is there any way to disable all loggings?

tbrand:
Sorry I've reverted this PR.

PR #36, opened 2017-05-10
disable etag generation in express

petermm:
the default hashing of the response and adding etag headers seems a bit too much.


tbrand:
Thanks!:+1:

PR #37, opened 2017-05-10
bring some sanity to ExecServer.kill

petermm:
right now it's killing random processes that doesn't even relate to the benchmarking, or leaving things behind. (eg random node processes)

also it doesn't handle fx cluster express which launches multiple node processes.

this fixes both issues, and cleans up the code a lot.

### DIFF
--- a/benchmarker/src/benchmarker.cr
+++ b/benchmarker/src/benchmarker.cr
@@ -57,34 +57,15 @@ class ExecServer
 
   # Kill the server process
   def kill
-    @process.not_nil!.kill
-
-    # Since ruby's frameworks are running on puma, we have to kill the independent process
-    if @target.name == "rails" ||
-       @target.name == "roda" ||
-       @target.name == "sinatra"
-      kill_proc("puma")
-    elsif @target.name == "express"
-      kill_proc("node")
-    elsif @target.name == "plug"
+    if @target.name == "plug"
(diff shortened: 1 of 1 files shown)


tbrand:
Seems good!
I'll try this. Thanks.

tbrand:
Oops, there are remaining servers.

```
taicsuzu 10:57:08 which_is_the_fastest/. > sudo lsof -i:3000
COMMAND    PID     USER   FD   TYPE             DEVICE SIZE/OFF NODE NAME
server_cr 1715 taicsuzu    6u  IPv4 0x4144180eabed3229      0t0  TCP *:hbci (LISTEN)
server_cr 1771 taicsuzu    6u  IPv4 0x4144180eabf11551      0t0  TCP localhost:hbci (LISTEN)
server_go 1817 taicsuzu    3u  IPv6 0x4144180e7aa5b8f9      0t0  TCP *:hbci (LISTEN)
```

petermm:
I see that.

Looks like those server_lang_framework that are not started from a bash script, don't get the kill signals (or they ignore them).

Can you try Process.run("pkill -9 -g $(ps -o pgid= #{@process.pid} | grep -o [0-9]*)", shell: true) instead? seems to do the job.

Tested here on kemal,express, clustered express and plug.

tbrand:
Thanks for this!!
I've fix as you said and merged this PR.:tada:

PR #51, opened 2017-05-19
Fix benchmarker killing self...

OvermindDL1:
As of a few PR's ago it seems the benchmarker keeps killing itself at the end of every test before it prints anything:
```shell
╭─overminddl1@snip ~/tmp/which_is_the_fastest  ‹asptest›
╰─➤  ./bin/benchmarker csharp
Language (Runtime)        Framework (Middleware)          Max [sec]       Min [sec]       Ave [sec]
------------------------- ------------------------- --------------- --------------- ---------------
[1]    29091 killed     ./bin/benchmarker csharp
```
And that is because they are `pkill -g`'ing the entire process group after getting the process group leader of the client process, which isthe benchmarker as it started the spawning of everything...

Should be killing the process children of just the process, this fixes that by killing the process children then the process itself, it works on the few I've tried it on so far without things staying running, but it will not work if anything double-forks (the prior method would not have worked on them either) but I'm not seeing anything doing that.

### DIFF
--- a/benchmarker/src/benchmarker.cr
+++ b/benchmarker/src/benchmarker.cr
@@ -74,7 +74,8 @@ class ExecServer
       Process.run("bash #{path} stop", shell: true)
     end
     # kill the server_lang_framework @process AND it's children
-    Process.run("pkill -9 -g $(ps -o pgid= #{@process.pid} | grep -o [0-9]*)", shell: true)
+    Process.run("pkill -9 -p #{@process.pid}", shell: true)
+    Process.run("kill -9 #{@process.pid}", shell: true)
   end
 end
 


tbrand:
It seems that there are remaining processes.
```bash
ruby    30931 taicsuzu    9u  IPv4 0x5a12ea7038d96ae7      0t0  TCP *:hbci (LISTEN)
node    31523 taicsuzu   11u  IPv6 0x5a12ea7026a7d4d7      0t0  TCP *:hbci (LISTEN)
dotnet  31968 taicsuzu  102u  IPv4 0x5a12ea70389551ef      0t0  TCP localhost:hbci (LISTEN)
dotnet  31968 taicsuzu  114u  IPv6 0x5a12ea701d71aa57      0t0  TCP localhost:hbci (LISTEN)
```

tbrand:
I've fixed here. Could you try this? https://github.com/tbrand/which_is_the_fastest/pull/53

OvermindDL1:
I branched 53 and ran kemal and got this:
```shell
╭─overminddl1@snip ~/tmp/which_is_the_fastest  ‹test53›
╰─➤  ./bin/benchmarker kemal
Language (Runtime)        Framework (Middleware)          Max [sec]       Min [sec]       Ave [sec]
------------------------- ------------------------- --------------- --------------- ---------------
kill: No such process (Errno)
0x41a64e: ??? at ??
0x45dd94: kill at /opt/crystal/src/process.cr 43:7
0x4126bc: ??? at /home/overminddl1/tmp/which_is_the_fastest/benchmarker/src/benchmarker.cr 72:5
0x44501b: main at /opt/crystal/src/main.cr 12:15
0x7f9ab7e5b830: __libc_start_main at ??
0x40dfa9: _start at ??
0x0: ??? at ??
```
Same result as without my PR here, with 51 it works.

I'm curious, maybe it would be better to structure the tests so there are startup scripts and shutdown scripts.  Things that can daemonize (like erlang) can do that.  Things that don't can use PID files and such.  Any specific application can have any specific startup and shutdown process that they want then.

tbrand:
Agree but the error itself doesn't seem "benchmarker killing self" now.
It seems that kemal is not successfully started or failed to find correct pid.

How about adding 'killing self' entry point for each server?
It may be hard work though...

OvermindDL1:
> Agree but the error itself doesn't seem "benchmarker killing self" now.

It is still, tracing it shows that the benchmarker process received the kill code that it sent out.

> It seems that kemal is not successfully started or failed to find correct pid.

It was running, tested and it responded during the benchmarker test and all.

> How about adding 'killing self' entry point for each server?

Hmm, that could be an interesting test too, but I know of some servers that cannot kill themselves, especially when they are spawned and killed on each request.

What are the down-sides to both startup and shutdown scripts?

tbrand:
> What are the down-sides to both startup and shutdown scripts?

There is no downside actually, but I cannot imagine what you are assuming.
Could you give an example script?

OvermindDL1:
Well for something like Elixir (since that is my main domain here):

server_elixir_plug_startup.sh (should also have a corresponding server_elixir_plug_startup.cmd for windows runs probably), I'd say have the benchmarker set the base directory of each call to the environment variable `BASE_DIR` or something for ease of use, but that is entirely optional:
```sh
${BASE_DIR}/_build/prod/rel/my_plug/bin/my_plug start
```
server_elixir_plug_shutdown.sh (and corresponding server_elixir_plug_shutdown.cmd for windows runs):
```sh
${BASE_DIR}/_build/prod/rel/my_plug/bin/my_plug stop
```
Things like rust and crystal and so forth could use PID files (start up the app and write the PID to a file, the shutdown script reads that file and kills it, or the app could implement a kill endpoint or whatever, tons of options).

And the benchmarker just runs each startup script as necessary, runs the client, then runs the shutdown script to stop it.  This is the usual unix way.  :-)

tbrand:
I don't face this problem recently and will close this.
If you will face the same problem, reopen or post new one.

PR #57, opened 2017-05-29
Perfect recent version

kjessup:
The Package.swift file for the Swift/Perfect app had the version locked to "2.0.x". I removed the minor version so that the newer packages will be utilized.


tbrand:
Great!!:tada: Thanks!!

PR #67, opened 2017-09-04
Remove kemal static handler

krypton97:
Removed static handler


krypton97:
@tbrand  the static handler is not disabled by default..

tbrand:
Thanks!

krypton97:
@did it affect the performance?

tbrand:
@krypton97 Please wait for the next result:smile:

PR #76, opened 2017-10-18
Dockerfile for rust/iron

Guillaume-Mayer:
Similar Dockerfile should work for Nickel and Rocket, I think.
Just needed to modify the binding to 0.0.0.0 in the main.rs file.


tbrand:
Thanks for sending PR!
Could you edit docker-compose.yml as well?

PR #80, opened 2017-10-26
add fastify as a target

thekemkid:


tbrand:
Will check this! thanks!

tbrand:
Failed to pass the benchmark since the response contains double quotes.
May be this issue https://github.com/fastify/fastify/issues/207 related to the problem.
Could you fix?

Thanks!

thekemkid:
Nice Catch! Updated 👍

tbrand:
Hmm... it's still failed for the POST request with empty body...
https://github.com/fastify/fastify/issues/297

tbrand:
Will close this since fastify is cannot be added under the rules.
If you can fix it, please re-post new PR.
Thanks!

PR #90, opened 2018-01-10
Run aspnetcore in release mode

mattzink:
The current code is running in debug mode (optimizations off)


tbrand:
Thanks!! :tada:

PR #95, opened 2018-01-14
Adds Nim's Jester

amscotti:
Not sure if there is something else needed to add a new language, but I think I have updated everything else.


amscotti:
@tbrand Sorry about that, didn't mean to check it in. The binary should be removed now.

tbrand:
Awesome!! New language is very welcome!! Thanks!! :tada:

PR #97, opened 2018-01-16
Avoid locking ruby / gem versions

waghanza:
Hi,

As for `ruby` benchmarks 
~~~sh
make ruby
~~~
fails with
~~~ruby
Your Ruby version is 2.4.2, but your `Gemfile` specified 2.4.1
~~~
since my  **version** is `2.4.2`.

I think (for ruby) :
+ version **SHOULD NOT** be precised in `Gemfile`
+  `Gemfile.lock` **SHOULD NOT** be versionned

Regards,


tbrand:
That's good idea! Thanks!!

PR #111, opened 2018-02-13
Dynamic calculation of benchmarking threads

waghanza:
Hi,

Thread number are fixed **16**.

I could be useful to compute threads (for `http/client`) depending on (logical) cpu counts.

The purpose behind that is to run benchmark on multiple platforms.

Regards,

### DIFF
--- a/tools/src/benchmarker.cr
+++ b/tools/src/benchmarker.cr
@@ -1,4 +1,5 @@
 require "benchmark"
+require "option_parser"
 
 # Prefix of pathes for each executable
 PATH_PREFIX = "../../../bin/"
@@ -129,10 +130,12 @@ class ExecServer
 end
 
 # Running client and returning span
-def client
+# threads : number of thread to launch simultaneously
+# requests : numbers of request per thread
+def client(threads, requests)
   s = Time.now
   t = System.cpu_count + 1
-  `#{CLIENT} -t #{t.to_i} -r 5000`
+  `#{CLIENT} #{threads} -r #{requests}`
   e = Time.now
   (e - s).to_f
 end
@@ -140,7 +143,9 @@ end
 # Benchmark
 # server : server context
 # count  : number of samples
-def benchmark(server, count) : BenchResult
+# threads : number of thread to launch simultaneously
+# requests : numbers of request per thread
+def benchmark(server, count, threads, requests) : BenchResult
   max : Float64 = -1.0
   min : Float64 = 100000.0
   ave : Float64 = 0.0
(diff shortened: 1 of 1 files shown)


tbrand:
Almost agree with your proposal, on the other hand, fix it as number of cores is not good for the flexible benchmarking.
If you should fix it to get number of threads from option, it's ideal. :+1:
Default number of threads can be number of cores.

waghanza:
Hi,

I agree with you.

For me, optimum bencharmks run on a per-thread basis, but this thread depends on the system capabilities. The optimum way **SHOULD** be to get user input, but it should be also a **default** value (if no input given).

The code define threads by :

~~~
Logical CPU count + 1
~~~

I'm based on https://stackoverflow.com/questions/16128436/setting-ideal-size-of-thread-pool#16128493

tbrand:
Yes! Same opinion as I said :+1:
> Default number of threads can be number of cores.

waghanza:
@tbrand Glad we are one the same line 

threads / requests option added ;-)

tools/src/benchmarker.cr
```diff
@@ -129,17 +130,22 @@ class ExecServer
 end
 
 # Running client and returning span
-def client
+# threads : number of thread to launch simultaneously
+# requests : numbers of request per thread
+def client(threads, requests)
   s = Time.now
-  `#{CLIENT} -t 16 -r 5000`
+  t = System.cpu_count + 1
```
tbrand: Please remove the redundant line :pray:

tools/src/benchmarker.cr
```diff
@@ -187,6 +193,26 @@ def puts_markdown(line, m_lines = nil, m = false)
...
 end
 
+####################
+### DEFAULT VALUES #
+####################
+threads = (System.cpu_count + 1).to_i
+requests = 10000
+record = false
+
+OptionParser.parse! do |parser|
```
tbrand: Doesn't this will crash when I run the benchmark like?
```
./bin/benchmarker --threads=8 kemal amber
```

tools/src/benchmarker.cr
waghanza: done :+1: thanks

tools/src/benchmarker.cr
waghanza: sure, any options

+ `./bin/benchmarker --threads=8 kemal amber÷`
+ `./bin/benchmarker kemal amber -t 8`
+ `./bin/benchmarker kemal amber --threads=8`

will lead to 
~~~crystal
threads = 8
all_frameworks = ['kemal','amber']
~~~

tbrand: awesome :+1:

tbrand:
Thanks for this!! :tada:

PR #116, opened 2018-02-14
Re-enable lucky

waghanza:
Hi,

[Lucky](https://luckyframework.org) could not compile due to non existant `manifest`.

The **manifest** file list static assets.

Regards,

PS : I added an empty one


waghanza:
oops, seems this `PR` was **early** merged ;-)

I have not seen that dependencies are a little bit outdated

PR #119, opened 2018-02-15
Update lucky

waghanza:
Hi,

It seems that [Lucky](https://luckyframework.org) was outdated.

Regards,


paulcsmith:
It should have been auto-generated before, but requires the Lucky CLI or bin/setup will fail. It *might* be good to keep the manifest so even if bin/setup fails, the manifest is always there, but the best solution may be to merge this PR and make sure Lucky CLI is installed.

waghanza:
@paulcsmith **lucky_cli** is required ;-)

This project has l

~~~yaml
dependencies:
  lucky_migrator:
    github: luckyframework/lucky_migrator
    version: "~> 0.4.0"
~~~

so **lucky_cli** will be properly setup

paulcsmith:
@waghanza You're right that that will include the Lucky CLI *shard*, but there is the Lucky CLI *executable* that needs to be installed to run the tasks: https://luckyframework.org/guides/installing/#install-lucky

paulcsmith:
BTW, thanks for updating this. I really appreciate it and I'm excited to see how it does :)

waghanza:
@paulcsmith seems to be quick, but not as quick as [Amber](https://github.com/amberframework) :trollface: 

But code is not exactlt the same between frameworks

paulcsmith:
Yeah, Lucky and Amber share very little code so that makes sense that there is a difference. I haven't done much perf tuning so I'm sure there's a lot to be improved, but I'm not planning on focusing on perf improvements for awhile. It's plenty fast for what I've needed :)

tbrand:
Thanks!

PR #122, opened 2018-02-15
Avoid versioning npm locks

waghanza:


tbrand:
Thanks!

thiamsantos:
Why the locks were removed? The pakage-lock.json should always be committed. [More information](https://docs.npmjs.com/files/package-locks).

Quoting:
> It is highly recommended you commit the generated package lock to source control: this will allow anyone else on your team, your deployments, your CI/continuous integration, and anyone else who runs npm install in your package source to get the exact same dependency tree that you were developing on.

thiamsantos:
I was looking, all the lockfiles, even for other package managers are been removed. You are experiencing any issues with the lockfiles?

waghanza:
Hi @thiamsantos,

`lock` files are **removed** for this project for `node` but also for `crystal` and `ruby`. 

More, I'd say that `lock` files are **recommended** for **production-ready** `apps` or `library`, which is not the case here.

Since this **tool** is creating environments on-the-fly, keeping `lock` files are **irrelevant** since the environment and host could differs (language version, for example).

Regards,

PR #125, opened 2018-02-20
Add Vert.X Web

AGutan:
How about some Java Web frameworks? 
Personally using Vert.X for some time. Interested how it stacks up against Rust and Go which I'm also interested in :) 

Will work on adding Vert.X.


waghanza:
Hi,

Thanks for this.

However, this readme is dynamically written (from computed result).

To add a framework please your project on a separated directory `java/vertx` in  your case and follow the rule here https://github.com/tbrand/which_is_the_fastest#the-rule.

Your effort is very appreciated :heart:

PR #128, opened 2018-02-22
Add rack-routing

georgeu2000:
I was able to get it to work with `bin/server_ruby_rack-routing` and `time bin/client`. 

`bin/benchmarker` wasn't working on my system...

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,7 +39,7 @@ criollo:
 
 
 # --- Ruby ---
-ruby: rails sinatra roda
+ruby: rails sinatra roda rack-routing
 
 # Rails
 rails:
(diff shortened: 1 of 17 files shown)


georgeu2000:
The original one belongs to my previous employer. It will be shut down eventually. I wanted to make sure that it stays online. There is no difference between them.

Besides that, I was the only person who worked on this gem and I wanted it to show up on my GitHub.

georgeu2000:
Thank you. Do you have any other projects that I can help with?

waghanza:
@georgeu2000 I don't known about `rack-routing`. Just a question : why using **rackup** ? (instead of for example **puma**)

PR #130, opened 2018-02-23
Fix lucky compilation

waghanza:
Hi,

[Lucky](https://luckyframework.org/) compilation fails due to static asset compilation, no need in our use case.

@paulcsmith How can we disable it ?

@tbrand This solution works but only using `make`, don't now about `neph`

Regards,

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,6 +74,7 @@ raze:
 
 # Lucky
 lucky:
+	mkdir -p crystal/lucky/public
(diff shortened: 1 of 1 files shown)


paulcsmith:
@waghanza I can add an option to Lucky to disable asset support. I have some other priorities but I'll open a PR here once it is added!

waghanza:
@paulcsmith :heart:

PR #147, opened 2018-02-28
Upgrade actix to version 0.4

fafhrd91:


waghanza:
@fafhrd91 good catch, but could you remove version checking -> not revelant in our case
I don't know about `rust` but probably a _*_ will install the **latest** stable

fafhrd91:
do you mean remove version from actix project?

waghanza:
@fafhrd91 probably, I can fix it upgrading to latest stable. 

The **goal** of this project is only about comparison, but it could be accurate to compare last stable version, na ?

fafhrd91:
But for sure, you want result be reproducible and in this case it is just easier to pin versions

fafhrd91:
btw when do you guys plan to run bench again?

waghanza:
@fafhrd91 not an idea :stuck_out_tongue_closed_eyes: in fact @tbrand display the result

=> `rust` is not **working** for me

fafhrd91:
I don’t understand your comment

waghanza:
@fafhrd91 I mean, we split work
---
@tbrand display the result and I help maintaining, so I can give you any date

@tbrand will answer :-)

fafhrd91:
ok, thanks

Issue #151, opened 2018-03-01
Crash on node js benchmarks


waghanza:
Hi,

`Node` **benchmarks** crashes.

The cause is  a misunderstanding on background process to kill (which launch server) on `kill_proc` method.

@tbrand Does `grep -E` exists on **OS X** ? I've found a workaround by getting _pid_ from ps + grepping with a regexp

Regards,

Issue #123, opened 2018-02-16
Debating about an official build tool ...


waghanza:
Hi,

I can see that their are 3 ways for building :

+ `make`
+ `neph`
+ `docker`

I personnaly using `make` (cause it's easy and I'm lazy).

@tbrand I can see that are building using `neph`

Shouldn't we build using the same thing ?

Behind that, there is the idea of a cross-platform solution -> building `objectif c`, `dot net`, and **else** using the same toolchain / isolation level.

For me `docker` is a good idea, but for **linux** enable tools (no `asp` no `swift` ...).

I suggest we use `Vagrant` (cause there is **OS X**, **Windows** ...).

@tbrand @OvermindDL1 what do you think about that ?

Regards,

OvermindDL1:
I'd probably just use `make` or shell scripts myself in general...?

waghanza:
@OvermindDL1 its what I do but ... I do not run `asp` or `swift` benchmarks ;-)

there is not many solution that could be used to create various (OS) targets

OvermindDL1:
> @OvermindDL1 its what I do but ... I do not run asp or swift benchmarks ;-)

Ditto.  ^.^;

Probably docker that is front-managed by a `make` file or shell script would probably be best.

waghanza:
@OvermindDL1 the problem will be the same ...

`docker` is `linux` specific

OvermindDL1:
> docker is linux specific

Uh... no?  I'm pretty sure it runs on mac, and it *definitely* runs on windows (we are using it on windows at work)...  o.O?

waghanza:
@OvermindDL1 ;-)

sure

`docker` work on **Windows**

but you can not create **Windows** `targets`, can you ?

OvermindDL1:
> but you can not create Windows targets, can you ?

In a VM setup, but then again, screw windows?  ^.^

waghanza:
@OvermindDL1 Not a **windows*** fan either, but as a performance comparison, I think its a mistake too not show `asp` performance ..

OvermindDL1:
> @OvermindDL1 Not a windows* fan either, but as a performance comparison, I think its a mistake too not show asp performance ..

But asp works just fine on linux (we are running an ASP site with the official Microsoft .NET Core on RedHat7 at work)?

waghanza:
@OvermindDL1 I do not know that such language is officially supported by **RedHat** 👍 

[...]

PR #152, opened 2018-03-02
use regexp to kill background process on node

waghanza:
Hi,

`Node` **benchmarks** crashes due because `nodejs` (process) could not be kill after each run.

**Node** is such a common word, I use a regexp to fix this, @see #151.

@tbrand If it's OK for you, I merge this `PR` -> I am not sure if this works on **OS X**

Regards,

### DIFF
--- a/tools/src/benchmarker.cr
+++ b/tools/src/benchmarker.cr
@@ -98,7 +98,7 @@ class ExecServer
     elsif @target.lang == "python"
       kill_proc("gunicorn")
     elsif @target.lang == "node"
-      kill_proc("node")
+      kill_proc("node.*js$")
     elsif @target.name == "plug"
       path = File.expand_path("../../../elixir/plug/_build/prod/rel/my_plug/bin/my_plug", __FILE__)
       Process.run("bash #{path} stop", shell: true)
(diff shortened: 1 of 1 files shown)


waghanza:
@tbrand so who can you identify the running process (for `node`) ?

waghanza:
@tbrand it could be a *solution* to move to `docker` as official tool -> in order **NOT** to deal with processes

#123

tbrand:
(Re-)implement docker to all of frameworks gonna be hard work.
But it is acceptable for me as I said in #123 .
On the other hand, as I said in #123 as well, cannot deprecate neph for some reasons.

waghanza:
@tbrand 
> (Re-)implement docker to all of frameworks gonna be hard work.
Sure, but, in my opinion, difficulty is less important from reliability (be closest as **production-ready** as it can)
> On the other hand, as I said in #123 as well, cannot deprecate neph for some reasons.
I probably can use `neph` to build docker containers, and the actual tool to benchmark it

PS : My idea is to go to `docker` on **development** and `ecs` (or alternative) on **production**
PSS : For me when we will use `v1.x`, we must run on a non local environment

tbrand:
So you mean you just want **build** a binary on docker? not for executing?

waghanza:
@tbrand no **both** :stuck_out_tongue_closed_eyes: 

but I'm starting to run from `docker` on #167

PR #158, opened 2018-03-05
[WIP] add spider-gazelle

stakach:
Please let me know if there is anything else I should do to add this framework.
I've started on documenting the project here too https://spider-gazelle.net/

Would love to see how it stacks up against the competition


kalicki:
@stakach What a strange name for a framework?! hahaha, looks promising...

waghanza:
@stakach Thanks for this  `PR`, I've added #160 to run it

PR #171, opened 2018-03-27
bumped versions

Ch4s3:


waghanza:
Hi @Ch4s3,

Thanks for update. 

I have no skills in `exilir`. 

Will it be accurate (e.g : avoid manually update version) not to use any version ? Will `mix` use the latest stable version in case of no version specified for dependencies ?

Regards,

OvermindDL1:
@waghanza 

The `elixir: "~> 1.6.4",` change means that is the minimum Elixir version supported (newer minor/patch is fine).

The `{:phoenix, "~> 1.3.0"},` means this is the minimum phoenix version supported (newer minor/patch is fine).

The `{:plug, "~> 1.5"},` means this is the minimum plug version supported (newer minor/patch is fine).

:-)

waghanza:
@OvermindDL1 okay, so I assume the version condition are similar to `ruby` / `Gemfile` 
I mean that **~> 1.6.5** mean **>= 1.6.5 && < 1.7.0**, is that right ?

In `Gemfile` (https://github.com/tbrand/which_is_the_fastest/blob/master/ruby/roda/Gemfile for example) I've remove version number to force to the last (published gem) release, it may differ in elixir / compiled lang

OvermindDL1:
I don't use ruby, but `~> 1.6.5` means 1.6.* and `~> 1.6` means `1.*.*`.  There is a lock file to lock at a specific version but you can upgrade past that by either specifying a new version of just running `mix deps.upgrade` or so to grab the latest version that still matches the given semver.  Leaving out the `~>` and just having a number in the mix.exs file will permanently only allow that version and no other.

On an aside, you can also have it use a file system location, a git with specific commit/tag, and a few other things (in addition to whatever else you want as it is a pluggable system).

OvermindDL1:
> In Gemfile (/ruby/roda/Gemfile@master for example) I've remove version number to force to the last (published gem) release, it may differ in elixir / compiled lang

This sounds absolutely horrifying to me...  The latest version could have an API change that makes it no longer work according to semver bounds, so why on earth would anyone ever do that ever?!  o.O

waghanza:
> This sounds absolutely horrifying to me
Sure, for the kind of application that will run **for** some users.

I mean, if there is an API change, we (maintainers) could handle it :stuck_out_tongue_closed_eyes:

PR #194, opened 2018-04-28
[DNM] Add Fastify (node)

thiamsantos:


thiamsantos:
Conflicts resolved.

kalicki:
Brazuca! Mandou bem =)

waghanza:
Awesome, I have just a question @thiamsantos. 

I use to come from Ruby world and I use to have the equivalent of **~** in `package.json` instead of **^**.

Is it node recommanded to use **~** since it keeps minor version (keep stability)

thiamsantos:
> Is it node recommanded to use ~ since it keeps minor version (keep stability)

Yeah, you are completely. Generally we don't worry about that because the lockfiles ensure this stability. But maybe a should update to use a tilde range.

PR #218, opened 2018-05-15
must use compile option "-d:release"

2vg:
If "- d: release" is not used, it will be built in debug mode.
It has considerable overhead and performance gets worse.
I want you to redo the benchmark right now...

### DIFF
--- a/nim/jester/Dockerfile
+++ b/nim/jester/Dockerfile
@@ -5,6 +5,6 @@ WORKDIR /usr/src/app
 COPY server_nim_jester.nim server_nim_jester.nim.cfg server_nim_jester.nimble ./
 
 RUN nimble install -y
-RUN nim c server_nim_jester.nim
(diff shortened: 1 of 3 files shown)


waghanza:
Awesome, thanks

I will wait until we use `wrk` to recompute results

OvermindDL1:
@2vg Wow, my thing in https://github.com/tbrand/which_is_the_fastest/issues/101 was testing mofuw when it is *not* in release build?  It was already one of the absolute fastest out!

Do note, the benchmarks shown in the readme are absolute garbage, the testing tool is single threaded so it cannot saturate hardly any of the servers at all.

I *still* don't know why the benchmarks were done with a custom single-core tool...  >.>

waghanza:
It will be updated @OvermindDL1 ... but feel free to `PR` (a `crystal` frontend that use `wrk`)

PR #223, opened 2018-06-01
Use wrk as a proper sieger

waghanza:
Hi,

This `PR` enable `wrk`. Siege process is handled by `wrk` and a **custom** `lua` script, write results in a `csv` file.

The `crystal ` part parse the `csv` and compute the rank from the average latency.

I'm not quiet sure of how to define this rank (what metric should be in).

@OvermindDL1 Any idea ?

Regards,

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,107 +8,113 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
 ## Result
 
 <!-- Result from here -->
-Last update: 2018-05-12
+Last update: 2018-06-06
 ```
-OS: Linux (version: 4.16.7-200.fc27.x86_64, arch: x86_64)
-CPU Cores: 4
+OS: Linux (version: 4.16.11-100.fc26.x86_64, arch: x86_64)
+CPU Cores: 8
 ```
 
 ### Ranking (Framework)
 
-1. [router.cr](https://github.com/tbrand/router.cr) (crystal)
-2. [raze](https://github.com/samueleaton/raze) (crystal)
-3. [spider-gazelle](https://github.com/spider-gazelle/spider-gazelle) (crystal)
-4. [lucky](https://github.com/luckyframework/lucky) (crystal)
-5. [japronto](https://github.com/squeaky-pl/japronto) (python)
-6. [amber](https://github.com/amberframework/amber) (crystal)
-7. [actix-web](https://github.com/actix/actix-web) (rust)
-8. [kemal](https://github.com/kemalcr/kemal) (crystal)
-9. [fasthttprouter](https://github.com/buaazp/fasthttprouter) (go)
-10. [nickel](https://github.com/nickel-org/nickel.rs) (rust)
-11. [iron](https://github.com/iron/iron) (rust)
-12. [rocket](https://github.com/SergioBenitez/Rocket) (rust)
-13. [iris](https://github.com/kataras/iris) (go)
-14. [echo](https://github.com/labstack/echo) (go)
-15. [gorilla-mux](https://github.com/gorilla/mux) (go)
-16. [polka](https://github.com/lukeed/polka) (node)
-17. [fastify](https://github.com/fastify/fastify) (node)
-18. [aspnetcore](https://github.com/aspnet/Home) (csharp)
-19. [perfect](https://github.com/PerfectlySoft/Perfect) (swift)
-20. [express](https://github.com/expressjs/express) (node)
-21. [vapor](https://github.com/vapor/vapor) (swift)
-22. [akkahttp](https://github.com/akka/akka-http) (scala)
-23. [gin](https://github.com/gin-gonic/gin) (go)
-24. [kitura](https://github.com/IBM-Swift/Kitura) (swift)
-25. [roda](https://github.com/jeremyevans/roda) (ruby)
-26. [rack-routing](https://github.com/iAmPlus/rack-routing) (ruby)
-27. [sanic](https://github.com/channelcat/sanic) (python)
-28. [mofuw](https://github.com/2vg/mofuw) (nim)
-29. [plug](https://github.com/elixir-lang/plug) (elixir)
-30. [phoenix](https://github.com/phoenixframework/phoenix) (elixir)
-31. [flask](https://github.com/pallets/flask) (python)
-32. [flame](https://github.com/AlexWayfer/flame) (ruby)
+1. [fasthttprouter](https://github.com/buaazp/fasthttprouter) (go)
+2. [rocket](https://github.com/SergioBenitez/Rocket) (rust)
+3. [actix-web](https://github.com/actix/actix-web) (rust)
+4. [evhtp](https://github.com/criticalstack/libevhtp) (cpp)
+5. [nickel](https://github.com/nickel-org/nickel.rs) (rust)
+6. [japronto](https://github.com/squeaky-pl/japronto) (python)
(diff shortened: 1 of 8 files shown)


waghanza:
@OvermindDL1 interesting ... and is the **percentile** as proper information ?

OvermindDL1:
Percentile is part of the deviation yep.

waghanza:
@OvermindDL1 sorry I do not understand how those metrics are calculated.
could you explain how those metrics (those defined [here](https://github.com/wg/wrk/blob/master/SCRIPTING#L90)) are calculated (and what their are representing) ?

OvermindDL1:
@waghanza You can get the full deviation via that `latency:percentile(float)` function it looks like, so probably this to get the full statistical deviation:
```lua
   for _, p in pairs({ 50, 90, 99, 99.999 }) do
      n = latency:percentile(p)
      io.write(string.format("%g%%,%d\n", p, n))
   end
```
Which of course you can output to `dot` format to generate pretty charts or output to csv or whatever.  :-)

And of course the `summary.requests/summary.duration` should get you the requests/second I think.

Oh, it is also good to report errors, only if non-0 though, those signify major issues and that test needs to be checked on if that happens, so we definitely don't want to ignore those!

OvermindDL1:
In addition to the other latency values of course*

OvermindDL1:
All of that information could be put on a whisker-chart or a deviation chart to show how things relate in more detail.

For an overall ranking number, I'd probably just use requests-per-second, maybe with a poor weighting for ones with high `latency.stdev` values (as those indicate wildly fluctuating latencies).

waghanza:
I've pushed a new `README`, with some results from `wrk`

Actually I only use the first rule (GET on /) to compute results (so this `PR` is not **mergeable**).

The reason I pushed the new **results** are to find the best way to display them.

@tbrand Do you find the results understandable ? (Results are based on a **5 minutes** `wrk` siege)

waghanza:
@tbrand I have updated the `README`, I think structure is **clearer**

> **DISCLAIMER** results are computed with the first rule : GET on /

@OvermindDL1 the results seems more realistic :stuck_out_tongue_winking_eye:

tbrand:
:+1:

README.md
```diff
@@ -8,158 +8,115 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
 ## Result
 
 <!-- Result from here -->
-Last update: 2018-05-12
+Last update: 2018-06-08
 ```
-OS: Linux (version: 4.16.7-200.fc27.x86_64, arch: x86_64)
-CPU Cores: 4
+OS: Linux (version: 4.16.11-100.fc26.x86_64, arch: x86_64)
+CPU Cores: 8
```
OvermindDL1: Is this the literal number of cores, or is it the number of hyper-threaded cores?  I showed both values in my benchmarker as it is an important value to state and I'd think it should be distinguished here as well.

README.md
```diff
@@ -8,158 +8,115 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
...
 32. [flame](https://github.com/AlexWayfer/flame) (ruby)
-33. [sinatra](https://github.com/sinatra/sinatra) (ruby)
+33. [flask](https://github.com/pallets/flask) (python)
 34. [jester](https://github.com/dom96/jester) (nim)
-35. [django](https://github.com/django/django) (python)
-36. [tornado](https://github.com/tornadoweb/tornado) (python)
-37. [rails](https://github.com/rails/rails) (ruby)
+35. [sinatra](https://github.com/sinatra/sinatra) (ruby)
+36. [django](https://github.com/django/django) (python)
+37. [akkahttp](https://github.com/akka/akka-http) (scala)
```
OvermindDL1: Wow, scala's akka one should not be that slow!  I wonder if the author forget to enable threading or so...  o.O

README.md
```diff
@@ -8,158 +8,115 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
...
-| scala                     | akkahttp                  |       64.114466 |       62.593498 |       63.369870 |
-| csharp                    | aspnetcore                |       51.437905 |       50.897969 |       51.071856 |
-| python                    | sanic                     |      118.590948 |       79.125047 |      101.660033 |
-| python                    | japronto                  |       28.014322 |       27.667655 |       27.879403 |
-| python                    | flask                     |      219.014852 |      114.585625 |      145.897795 |
-| python                    | django                    |      478.203936 |      194.889338 |      271.425040 |
-| python                    | tornado                   |      726.349087 |      706.362335 |      718.087107 |
-| nim                       | jester                    |      246.613563 |      246.253752 |      246.379956 |
-| nim                       | mofuw                     |      147.627614 |       83.375723 |      108.593340 |
+| ruby                      | rails                     |               0 |          5244.0 |          0.88MB |
```
OvermindDL1: Hmm, should this list be sorted by language name, or should it be sorted by req/s or so since the above is not listing the req/s on the language?

README.md
```diff
@@ -8,158 +8,115 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
...
-Last update: 2018-05-12
+Last update: 2018-06-08
 ```
-OS: Linux (version: 4.16.7-200.fc27.x86_64, arch: x86_64)
-CPU Cores: 4
+OS: Linux (version: 4.16.11-100.fc26.x86_64, arch: x86_64)
+CPU Cores: 8
 ```
 
 ### Ranking (Framework)
```
OvermindDL1: Should the req/s and maybe throutput be listed with these as well so people can see the differences in the ranking, not just the positional ranking?  Perhaps it should be a markdown table too?

tools/src/benchmarker.cr
```diff
@@ -126,42 +128,64 @@ def frameworks : Array(Target)
...
+end
 
-  sleep 5
+class Percentile
+  JSON.mapping(
+    fifty: String,
+    seventy_five: String,
+    ninety: String,
+    ninety_nine: String
+  )
```
OvermindDL1: Showing the 4 or 5 9's percentile stat is also useful and standard as it shows the extreme outliers.

README.md
waghanza: it the number of **physical** cores -> my proc is `AMD FX-8320E Eight-Core Processor`
in fact, I'm running on different computers / kernels

README.md
waghanza: yes, I think results are wrong

README.md
waghanza: req/s, but a think at a time ... (I'll sort after results stabilization)

waghanza:
@OvermindDL1 I have published other results on https://github.com/tbrand/which_is_the_fastest#ranking-framework
+ 16 threads
+ 15 seconds
+ 1000 connections

but still `GET` on `/` => next step is to combine all routes

README.md
```diff
@@ -8,158 +8,115 @@ Any idea is :heart:, let discuss about it on [![Join the chat at https://gitter.
 ## Result
 
 <!-- Result from here -->
-Last update: 2018-05-12
+Last update: 2018-06-08
 ```
-OS: Linux (version: 4.16.7-200.fc27.x86_64, arch: x86_64)
+OS: Linux (version: 4.16.13-300.fc28.x86_64, arch: x86_64)
 CPU Cores: 4
```
OvermindDL1: 4 core is no where near enough for a good test!  o.O

PR #240, opened 2018-06-09
Add rayo.js to the tests.

aichholzer:
@waghanza -As mentioned [here](https://github.com/tbrand/which_is_the_fastest/issues/237)
Thank you.


node/rayo/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "author": {
+    "name": "Stefan Aichholzer",
+    "email": "theaichholzer@gmail.com",
+    "url": "https://github.com/aichholzer"
+  },
+  "dependencies": {
+    "rayo": "^1.0.4"
```
waghanza: I prefer using `~` instead of `^`

aichholzer: Noted, thank you.

aichholzer:
@waghanza -How are you running the other Nodejs frameworks (express, fastify, polka)? -None of them implements a `HEAD` endpoint (which is what `curl -vIL` is doing) and they all return `404` on that request.

waghanza:
Yes. I'm confusing myself. Juste wating to add some tests

aichholzer:
@waghanza -Let me know if there is anything I can help you with.

waghanza:
curl -vIL could be used to display http headers

aichholzer:
@waghanza -Yes, but the server needs to implement a `HEAD` endpoint, for instance:
```js
const rayo = require('rayo');
rayo({ port: 3000 })
  .head('/', (req, res) => res.end(''))
  .start();
```

`curl -vIL` basically makes a `HEAD` request to the server.

PR #251, opened 2018-06-14
Update link to rack-routing

georgeu2000:


waghanza:
Thanks for update, but the change has to be made directly in the source code https://github.com/tbrand/which_is_the_fastest/blob/master/tools/src/benchmarker.cr#L51 (it will be overwritten otherwise)

waghanza:
@georgeu2000 I can see that the link (leading to github) to the source code is dead, based on https://rubygems.org/gems/rack-routing

georgeu2000:
Oh, that's right. I forgot that the change has to be made directly in the source code. Thank you for updating it.

I'll see if I can fix rubygems.

georgeu2000:
OK, updated. Thanks for letting me know about rubygems link.

georgeu2000:
How do I fix the conflict? Do I need a new branch and new PR?

Issue #140, opened 2018-02-26
Add Hanami (Ruby)


kalicki:
Hanami http://hanamirb.org/

ping @jodosha

PR #273, opened 2018-06-22
Add hanami (ruby)

waghanza:
Hi,

This `PR` add **hanami**

@see #140 

Regards,

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -51,6 +51,10 @@ roda:
 rack-routing:
 	docker build -t rack-routing ruby/rack-routing
 
+# Hanami
+hanami:
+	docker build -t hanami ruby/hanami
+
 # Flame
 flame:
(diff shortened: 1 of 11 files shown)


jodosha:
@waghanza Thanks for adding Hanami. Please check if it's working correctly and in production mode (check `Hanami.env?(:production)`). I don't see routes definition, so it could be it's hitting 404 or raising internal exceptions. Thanks.

waghanza:
Thanks for checking this @jodosha

Routes are defined here =>
https://github.com/the-benchmarker/web-frameworks/blob/master/ruby/hanami/apps/web/config/routes.rb

In fact, I've done multiples `PR` to add `hanami`.

It appears that I forget to configure it as production

jodosha:
@waghanza Thanks for getting back. 👍

Issue #238, opened 2018-06-08
How to rank frameworks


waghanza:
Hi,

As a standard tool, we decide to go on `wrk`. 

This **benchmarking** tool _give_ us a lot of useful informations.

+ Global (per scenario / route to test)
   + Duration
   + Number of requests
   + Total bytes received
   + Number of error (socket / read / write / http / timeout)
   + Number of
+ Latency
   + Minimum
   + Maximum
   + Average
   + Standard deviation
+ Percentile
   + 50
   + 90
   + 99
   + 99.999

**ALL** informations **SHOULD** be displayed, but **COULD** be used to determine ranks (I mean display is **OK**, but to keep it simple now we **SHOULD** only use `1` metric to rank).

@OvermindDL1 What do you think about talking only the number of `requests-per-second` to `rank` ?

Regards,

waghanza:
sure. I think those informations **are** useful. However, they **SHOULD** be displayed, not imperatively on the main table, but somewhere ...

In order to keep simple, I suggest choosing 3 or 4 metrics to display on main table :
+ Duration
+ Requests
+ Errors
+ Latency (average or deviation, not sure of what)

OvermindDL1:
Hmm, for active display, perhaps:

* req/s
* average response latency
* 99%ile response latency
* throughput

Honestly errors probably don't need to be shown, if it's anything but `0` ever then a test is broken and it shouldn't be posted anyway

waghanza:
@tbrand To simplify reading, I think we should display one **ONE** table of results (sorted by rank)

aichholzer:
It's interesting to see how these results can change, last time [this file](https://github.com/waghanza/which_is_the_fastest/commit/1f81ad7d22730fc2c4345cee31c6d5ad57230676) was updated, `rayo` was doing ~87K reqs/s and `polka` around ~86K reqs/s. Now they have both dropped by +10K reqs/s each and the order has shifted.

What kind of consistency are you providing/considering for these tests?
Consistently, in all my tests `rayo` is still the faster.

waghanza:
Hi @aichholzer, We are adjusting tests parameters, duration, request number ...
I hope to release **stable** results soon

OvermindDL1:
@waghanza I'm not sure that adjusting such things should cause such shifting results, that implies to me the server it is running on is either running other code, the network is chatty, it's a cloud machine or VM, etc... etc...  My results were fairly stable, less than 1% variance between runs.

[...]

Issue #252, opened 2018-06-15
Migrate as an org


waghanza:
Hi @tbrand,

Some services (which are in the marketplace) are more usable (due to github permissions) to use / enable on some group project.

For example, using https://dependabot.com (to manage dependency) **COULD** only be done either on my fork or on a group project.

Do you mind if I create a group in order to migrate this project ?

It will be very helpful for **maintainers**.

Regards,

tbrand:
OK, I accept the migration.

waghanza:
Then we have to find a name

For me, the **main** idea behind that `performance` battle :stuck_out_tongue_closed_eyes: is some king of promotion, so name could be about that

tbrand:
Specifically?

waghanza:
late take some time to think about that ;-)

at least I would have some ideas in 2 days ;-)

@OvermindDL1 any ideas ?

tbrand:
How about `which-is`/`the-fastest`? lol

OvermindDL1:
Hah, I like @tbrand's suggestion, leaves it open for more repo's with more tests later too.  ^.^

waghanza:
So why not a more **generalistic** word like `agon`, then it could be `agon`/`backend`

https://en.wikipedia.org/wiki/Agon

waghanza:
@tbrand @OvermindDL1 why not 
[agon](https://en.wikipedia.org/wiki/Agon)/[actium](https://en.wikipedia.org/wiki/Battle_of_Actium)
for this repository, but with a focus on performance for **backend** apps (web)

leaves open for other benchmarks with a focus on performance for **fontend** apps (`node` things ...) or why not asking @lukeed and @aichholzer to join this adventure :heart:

tbrand:
@waghanza Sorry I can't agree the name.
For me, who don't have enolugh vocabulary, dont' understand what it means without googling.
`which_is_the_fastest` is verfy simple, understandable and catchy for every users on Github.
Even we drop the name of `which_is_the_fastest`, it have to be `simple, understandable and catchy` at least.

waghanza:
@tbrand `which_is_the_fastest` is surely understandable, by whom have some technical skills
what I mean is that :

+ `fastest` **IS** subjective (depending on use case) -> a telecommunication app and a web app are not considered **fastest** the same way
+ `is` **COULD** be restraining for some people, according that anyone has enough knowledge to say that he has **the** solution

In short :
[...]

PR #281, opened 2018-06-25
Clarify results

waghanza:
Hi,

Following the #238, it **COULD** be `awesome` to report :
+ Some result ranked by `req/s`
+ Some result ranked by `latency`

It also **COULD** help `clarify` results. May developers argue that comparing so languages such as `node` and some such as `elixir` (which targets are completely different) are not very relevant.

This `PR` show results order by `req/s` **AND** by `latency`

Regards,


waghanza:
@OvermindDL1 Should not we use percentile, instead of average latency to rank by `latency` ?
it will be more accurate no ?

OvermindDL1:
50th percentile, 99.99th percentile, and standard deviation would be the best to show as it shows the latency range and how it grows.

waghanza:
@OvermindDL1 my question is want metric (one ?) use to order `latency` tab ?

is it accurate to differentiate  both table ?

OvermindDL1:
I'd include all information on both tables, but I'd use two table, one sorted on `req/s`, the other sorted on 50th percentile latency.

If only Github had the ability to mark a table as sort-by-click-on-header.  Could always make a dedicated HTML for that though (github pages or so).

waghanza:
Don't you think tables will be **huge** ?

My final purpose is to compute result, store them in a `BigQuery` or else, and having a website displaying them

@tbrand @OvermindDL1 What do you think ?

OvermindDL1:
Eh, the tables would be long sure, but that seems fine to me (especially if you can clamp the headers to the top as it scrolls).

waghanza:
@OvermindDL1  On a **race** between `ruby` /  `node` / `php` / `python`, I have different results (ranks) for `latency` :
+ average
  1. roda
  2. rack-routing
  3. flame
  4. hanami
  5. sinatra
  6. polka
  7. rayo
  8. fastify
  9. japronto
  10. restify
+ 50th percentile
  1. roda
  2. rack-routing
  3. flame
  4. sinatra
  5. hanami
  6. polka
  7. rayo
  8. fastify
  9. koa
  10. restify
+ 90th percentile
  1. roda
  2. rack-routing
  3. flame
  4. polka
  5. sinatra
  6. rayo
  7. hanami
  8. japronto
  9. restify
  10. fastify
+ 99th percentile
  1. roda
  2. rack-routing
  3. flame
  4. hanami
  5. sinatra
  6. rayo
  7. japronto
  8. polka
  9. restify
  10. fastify

I suggest then to **keep** `existing` version (sorting by `average`) and **adjust** if after using a **cloud**  to compute results

OvermindDL1:
Low latencies at high percentiles most likely means they generate little to no garbage in comparison.  :-)

waghanza:
@OvermindDL1 not understood

OvermindDL1:
Just me guessing at what might cause 50 and 99+ percentile differences, ignore.  ^.^

For note, this is a good chart for how to display the overarching information:

Add in `req/sec` as well and it would be a good detail showing for each framework.

image

tbrand:
Could you remove the column which is not related to each ranking?
For example, remove `Requests / s` for `Ranked by latency` ranking.

Other things are acceptable for me :+1:

waghanza:
I've done a little change to the `benchmarker` code base, i know use a `redis` to store result (`redis` was picked because of simplicity but it add a dependency).

We need to refactor using an other key/value store (to avoid adding dependency) between :

* https://github.com/kostya/kyotocabinet
* https://github.com/CodeSteak/Nuummite
* https://github.com/crystal-community/kiwi

OvermindDL1:
> (redis was picked because of simplicity but it add a dependency)

As long as it's ran via docker or so and not needed to be installed.  Why not just use leveldb/rockdb if it's just a KV store needed or sqlite for querying?

waghanza:
@OvermindDL1 I mean a `redis` installed on   `sieger` machines, not targets. I need a k/v store to store all results as json (at least in order to let us decide what informations to render in each tables)

OvermindDL1:
> I need a k/v store to store all results as json (at least in order to let us decide what informations to render in each tables)

Definitely no point in using something as crazy-heavy as redis, just embed leveldb or rocksdb, I use both rather excessively, and they will far *far* outperform redis.

waghanza:
@OvermindDL1 sure, it was I have said :stuck_out_tongue: 
the list above was essentially for me picking an idea ^^

'cause it was late last night (and I was not very productive), I take a random tool -> I'll rewrite this part

waghanza:
@OvermindDL1 The idea with this modification was to decide what to show in tables, I have made for this a sample here => https://github.com/waghanza/which_is_the_fastest/blob/clarify_results/README.md#results but I'm not very convinced

@tbrand What do you think ?

PS : I have taken advantage of this, by adding a **warm-up** call :stuck_out_tongue:

waghanza:
@OvermindDL1 for me `ranking by latency` **SHOULD** be ranked by 50th perc. (more reliable -> ex : `rack-routing` **SHOULD**  the **5th** cause of 50th perc., it is actually the **2th** 'cause of average)

@georgeu2000 do you observe same results ?

OvermindDL1:
@waghanza yep, 50% makes sense, for someone else then 99% may make sense, etc... etc...  I really wish github had sortable column by headers, maybe need to set up a github pages so they become clickable.  ^.^;

waghanza:
@OvermindDL1 I have find that it's possible, but very hard to use / maintain (sortable table). my idea is to push results is `BigQuery` or else and create a website for this (github's page when #252 :stuck_out_tongue: )

OvermindDL1:
No need for BigQuery, just generate an html file via a template and use some progressively enhancing javascript to make the table sortable via column.

waghanza:
@OvermindDL1 I know it could be done like that, but I find it mor flexible to separate compuation from layout 

e.g -> use `travis` to compute, and a website to display results :stuck_out_tongue_winking_eye: 

moreover, compute and generate markdown on `travis` need to push from `travis` to `github`, no a big fan of this kind of solution, but I suggest we open an other issue for this question :stuck_out_tongue_closed_eyes:

waghanza:
@tbrand Just after merging this, I'll release :100:

PR #288, opened 2018-07-04
Downgrade python

waghanza:
Hi,

This  `PR`  downgrade **python**, `3.6` instead of `3.7`, as `uvloop` is not compatible with `3.7` (yet ?)
https://github.com/MagicStack/uvloop/issues/178

@frnkvieira As far a I understand workers (at least for app I design) **SHOULD** be 1 per core. Why  `vibora` spread on `cpu_count() + 2` by default

@see https://github.com/vibora-io/vibora/blob/master/vibora/server.py#L282

@greed2411 I can see that on the implementation, you specified headers. 
Is there any idea behind that ? (there is no header in doc examples).

Regards,

### DIFF
--- a/python/japronto/Dockerfile
+++ b/python/japronto/Dockerfile
@@ -1,4 +1,4 @@
-FROM python
+FROM python:3.6
 
 WORKDIR /usr/src/app
 
--- a/python/japronto/requirements.txt
+++ b/python/japronto/requirements.txt
@@ -1,6 +1,1 @@
-appdirs==1.4.3
 japronto==0.1.1
-packaging==17.1
-pyparsing==2.2.0
-six==1.11.0
-uvloop==0.8.1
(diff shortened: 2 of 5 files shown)


frnkvieira:
@waghanza From my tests, this is the best setup for Vibora if you take the average of different benchmarks. There is no science on this and never will, the optimal number of workers is always dependent on the application workload. In my own benchmarks I setup 1 worker per core for each framework to make it fair as you should keep doing. Hope it makes sense :)

waghanza:
@frnkvieira I understand you use 1 worker per core on you're benchmarks, but should not it be the default ? I mean, at least having a huge workload it is enough no ?

frnkvieira:
@waghanza there is a difference between "enough" and "optimal", as I said, this is a personal choice based on my experience but usually throwing a little more workers than CPU cores leads to better results. If you look at Gunicorn they also seems to throw one "extra" worker on their formula. (http://docs.gunicorn.org/en/stable/design.html#how-many-workers).

Issue #296, opened 2018-07-18
Fix jester


waghanza:
Hi,

`jester` is failing. A mid (short) workaround is to use a specific commit as in https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Nim/jester/techempower.nimble

`jester` target the next `nim` release, hoppefully comming in August :-)

Regards,

dom96:
because that image only offers the latest stable Nim version AFAIK

PR #297, opened 2018-07-18
fix jester

waghanza:
Hi,

This `PR` fix `jester`.

This a temporary patch, since jester latest is now incompatible with nim (I use a specific commit).

@see #296 

Regards,


waghanza:
@dom96 How can `jester` listen to  `0.0.0.0` ?

PR #305, opened 2018-07-25
Fix jester

waghanza:
Hi,

This `PR` fix `jester` implementation

@see https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Nim/jester

@dom96 Could you validate this code ?

Regards,


dom96:
You need that .nimble file. You can't remove it.

waghanza:
@dom96 Even if option are passed through the compiler ?

dom96:
You're calling `nimble` in the docker file and the .nimble file specifies dependencies for the script, so yes, you need the .nimble file.

PR #312, opened 2018-07-31
remove version specification on elixir impl

waghanza:
Hi,

This `PR` remove version specification in `mix.exs`.

This originally fix CI, cause on **travis** `elixir` is `1.7.1` and `1.6.4` on **docker**.

Btw, version is specified on https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Elixir/phoenix/mix.exs

@kelvinst Is it a best-practice or absolutely a horror to do ?

Regards,

PS : I'm an _interpreted_ language guy :stuck_out_tongue_winking_eye:

### DIFF
--- a/elixir/phoenix/mix.exs
+++ b/elixir/phoenix/mix.exs
@@ -4,7 +4,6 @@ defmodule MyPhoenix.Mixfile do
   def project do
     [app: :my_phoenix,
      version: "0.0.1",
-     elixir: "~> 1.6.4",
      elixirc_paths: elixirc_paths(Mix.env),
      compilers: [:phoenix, :gettext] ++ Mix.compilers,
      build_embedded: Mix.env == :prod,
--- a/elixir/plug/mix.exs
+++ b/elixir/plug/mix.exs
@@ -4,7 +4,6 @@ defmodule MyPlug.Mixfile do
   def project do
     [app: :my_plug,
      version: "0.1.0",
-     elixir: "~> 1.6.4",
      build_embedded: Mix.env == :prod,
      start_permanent: Mix.env == :prod,
      deps: deps()]


OvermindDL1:
That lines only defines the version range that it can work with, something like `"~> 1.6.4"` means to work with `1.6.4 <= ver < 1.7`, what it probably should be is `elixir: "~> 1.6",` as that means work with `1.6 <= ver < 2.0`.  There is rarely a reason to put a patch specifier in that version field.

waghanza:
I see this useful for _dynamic_ languages, but as a _pseudo_ compiled language it's difficult for me to find any **realworld** advantages / disavantages of using this.

also this impl is so simple

OvermindDL1:
Elixir is a dynamic language, it has a full interpreter in it (and you can run through the interpreter just fine), it can also compile to 'beam' object files (bytecode, think python's pyc files) but even those can be swapped in an out for hot code reloading.  It's as dynamic as Python is (technically even more so since you can run multiple 'versions' of code at once).

waghanza:
Ok, I'll force a specific version then `1.6.6`.

I'll be interesting to display version.

OvermindDL1:
Make sure OTP is set to at minimum version 21, the BEAM VM had some useful performance enhancements.

waghanza:
How Can I do that ?

OvermindDL1:
OTP21 would not be specified in the mix.exs file but rather just make sure that version of the VM is installed in docker.  :-)

Definitely don't use the package distribution's.  If you want to use a deb then add the erlangsolutions ppa as that is always updated, though honestly should just use the asdf language manager (it can handle a *lot* more than just erlang/elixir if that is useful elsewhere) and install the latest erlang and elixir that way (in that order, the elixir builds based on the installed erlang/otp version).

waghanza:
What do you mean the by _asdf language manager_ ?

OvermindDL1:
You've never used [`asdf`](https://github.com/asdf-vm/asdf) before?!  o.O

It's pretty big in a few areas.  It let's you install any published or git version of a supported language (and it's easy to make a plugin to add support for a language, consequently it supports a *lot* of languages) and you can set a specific version as 'global' default and you can even have subdirectory specific versions to use different ones and all.  It's a fantastic little manager.  :-)

OvermindDL1:
For note, the 'Officially Supported Plugin' list is at:
https://github.com/asdf-vm/asdf-plugins

waghanza:
I was using [https://www.softwarecollections.org](software collection) as a  `redhat` user (`centos`, `rhel`, `fedora`, `coreos`) and since `amazonlinux` is based on  `rhel`

waghanza:
as for now, I will merge #315 

this is a temporary solution since next step (fixing version of each implementations)

PR #319, opened 2018-08-02
Use CI for each implementations

waghanza:
Hi,

`CI` is not very useful, because we use a different tool (to build impl) + tests failure do not lead to build failure

- [x] Run CI for each implementation
- [x] Use `crystal spec` to test per implementation

Regards,


waghanza:
@tbrand @OvermindDL1 this  `PR` purpose is to have a **green** status on `CI` to validate `PR`
https://travis-ci.com/the-benchmarker/web-frameworks/builds/83650036

@OvermindDL1 could you remind me want are / are not standard http header, that I can assert that headers exists in every implementations here

Issue #318, opened 2018-08-02
Difficult to understand the ranks


tbrand:
I love the expandable tables, however, the rankings are difficult to understand to me.
As a first impression, "average" is higher priority than "50% percentile" since it is located left of it.
But the ranks are sorted by latter one. It is very confusing.

Also I prefer to add numbers of ranks even if it is a bit verbose.
Like this.

No. 1 👑 framework name
No. 2 🌟the other framework
No. 3 ⭐️ another one

I would like to use the table as an evidence, not ranking itself.
So the above ranking should be detached from the evidence table.

What we want to tell?
I would like to keep a `which_is_the_fastest` concept. (Not be a generic benchmark I hope.)

tbrand:
Additionally, current tables are not ranking but just sorting results.

waghanza:
@tbrand I understand what you mean when you speak abount columns, as lettre columns should have more valable data. Lets list columns that we want to display (that's thé easy part ^^^)

However, I do not understand the differences you make between ranking and sorting.  The first element of each table is on top of the podium for me

tbrand:
See the old commit.
https://github.com/the-benchmarker/web-frameworks/tree/c890d049a878c108214fe185c4e623712e8d4dff

The ranking is very simple, detached from the detailed table and numbered.
It's quite understandable for me.

Currently, there are too many frameworks to be listed all of them.
So it might be good to list only top 10 frameworks.

aichholzer:
In overall, results are very inconsistent and mostly wrong...

tbrand:
We'll try to improve the stability.
Let me focus on how to show the result here.

tbrand:
@waghanza

You should understand not only the rankings but also core value of this repository.
People don't put stars for the profiling but for the rankings at first.

> However ranking differs depending on point of view (what matters depending on usage)

We can fix the point of view. And don't have to show the language rankings now.

> Also, performance of a real-world application will also vary depending on architecture in use. For me a so simple ranking is counter-productive, because of how we scale.

Then we just put a description for it on README.md.
It's completely contradictory if we cannot make a rankings for a server. Then the profiling table means nothing as well.

Also please open another issue for the discussion of correctness if you want.

waghanza:
@tbrand I know figures are not correct and can not be yet, I'll open an issue later if necessary :heart:

Honestly, I can not correlate the fact that people are putting stars with any conclusion, just that they appreciate our work

[...]

PR #327, opened 2018-08-27
[WIP] Show ranks, top 5

waghanza:
Hi,

As suggested in https://github.com/the-benchmarker/web-frameworks/issues/318, a **ranking** will be useful to display.

This `PR` add a top 5 ranking (as for now only per latency).

@tbrand I've added a sample results test on https://github.com/waghanza/http-benchmark/blob/ranking/README.md

Regards,

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ flame:
 	docker build -t flame ruby/flame
 
 # --- Crystal ---
-crystal: kemal router.cr lucky amber spider-gazelle
+crystal: kemal router.cr amber spider-gazelle # lucky
 
 # Kemal
 kemal:
--- a/Makefile
+++ b/Makefile
@@ -69,7 +69,7 @@ flame:
 	docker build -t flame ruby/flame
 
 # --- Crystal ---
-crystal: kemal router.cr amber spider-gazelle # lucky
+crystal: kemal router.cr amber spider-gazelle lucky raze
 
 # Kemal
 kemal:
--- a/README.md
+++ b/README.md
@@ -64,117 +64,109 @@ bin/benchmarker [tools]
 ## Results
 
 <!-- Result from here -->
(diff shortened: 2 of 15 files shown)


tbrand:
Cool! :+1::+1::+1:

I prefer more showy result by using emojis.

For example.

- :one: abc
- :two: def
- :three: :ghi:
- :four: jkl
- :five: mno

or

- :crown: abc
- :trophy: def
- :gem: ghi
- :ring: jkl
- :crystal_ball: mno

or what ever you like.

---

Optional. Don't have to be implemented in this PR.
If we can refer a previous result, the changes are interesting as well.

Like this.

- :chart_with_upwards_trend: abc
- :chart_with_downwards_trend: def
- ghi
- :chart_with_upwards_trend: jkl
- mno

or

- :arrow_upper_right: abc
- :arrow_lower_right: def
- :arrow_right: ghi
- :arrow_upper_right: jkl
- :arrow_right: mno

waghanza:
@tbrand yeah, emojis could be easier to read for somes :stuck_out_tongue: 

However, having an up (or down) arrow showing rank increase (or decrease) for a framework requires to track history

I think there is 2 different approaches for this :
+ Storing results in a database (for example [BigQuery](https://cloud.google.com/bigquery)) and display them with a specialized tool (for example [Tableau Software](https://www.tableau.com)) like **github** do for `opensource` archives
+ Storing results in a local database and push them in a remote storage like [space](https://www.digitalocean.com/docs/spaces)

I think the first approach is more **communicative** than the second, but requires to have **public** results

It could be interesting to implement switch for those approaches 

What do you think ?

tbrand:
The result witch have to refer the previous one is optional for me.
So I think we don't have to discuss about it here.

Anyway, both approaches are fine for me.

Issue #221, opened 2018-06-01
Question about the status code in the rule


greenlaw110:
In the rules section it says:

* GET '/' return status code 200 with empty body
* GET '/user/:id' return status code 200 with the id
* POST '/user' return status code 200 with empty body

As per HTTP standard, empty body should have a 204 response, and POST success normally indicates a create success, thus it might be good to return 201 Created.

tbrand:
Agree but it requires a lot of efforts to fix. lol

OvermindDL1:
Rules are good, the more detailed the better.

Like not just saying things like GET and 204 with empty body or so, but even more detailed so that the implementations remain good comparisons, things like routing must be performed a certain way or validation must be done on x-and-x, etc...

waghanza:
@tbrand @OvermindDL1 digging for any benchmarks, and how developers use it / implement it
it appears that on _most_ **case**, there is only **1** endpoint.

What do you think about that ?

I think it **COULD** be helpful to decrease our rule set to only one endpoint implementation, for example, it could be a `GET` to `/hello/:name` to ensure  `routing` and `param parsing`.

The advantage I see :

+ simpler to implement -> to increase community
+ can focus on standard -> HTTP header, `SO_REUSEPORT` ...
+ simplify benchmarker code
+ less time to run (or we can publish more ofter with **cloud** results)

PR #352, opened 2018-09-07
Patch meiheld

waghanza:
Hi,

Following https://github.com/bottlepy/bottle/issues/1091, this `PR` patch `meinheld`.

Thanks @agalera :tada: 

Regards,

### DIFF
--- a/python/bottle/server.py
+++ b/python/bottle/server.py
@@ -1,3 +1,6 @@
+from meinheld import patch
+patch.patch_all()
+
 from bottle import Bottle, run
 
 app = Bottle()
--- a/python/flask/server.py
+++ b/python/flask/server.py
@@ -1,3 +1,6 @@
+from meinheld import patch
+patch.patch_all()
+
 from flask import Flask
 
 app = Flask(__name__)


waghanza:
I use this feature also for `flask`

| Framework (Middleware)    |         Average |  50% percentile |  90% percentile |  99% percentile | 99.9% percentile | Standard deviation |Requests / s |      Throughput |
|---------------------------|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|---------:|
| bottle (unpatched)                   | 22.18 ms | 18.03 ms | 37.88 ms | 75.21 ms | 324.32 ms | 15796.33 |48441.00 | 119.39 MB |
| flask  (unpatched)                   | 52.62 ms | 37.67 ms | 110.96 ms | 158.00 ms | 632.17 ms | 41189.67 | 20414.67 | 50.18 MB |
| bottle (patched)                   | 21.00 ms | 13.69 ms | 41.86 ms | 90.58 ms | 282.19 ms | 18187.67 | 53832.67 | 132.67 MB |
| flask  (patched)                   | 54.46 ms | 38.40 ms | 109.36 ms | 166.15 ms | 394.67 ms | 36173.33 |19572.00 | 48.09 MB |

For `bottle` this patch is :tada: but not sure if `flask` should be **patched**

waghanza:
@agalera @benoitc any idea about the performance diff ? any help will be :heart:

agalera:
in bottle should not be patched for errors in the response.
Bottle is much faster than Flask, this is normal.

If you want to improve the performance by returning answers in json format, you can install ```ujson``` is an optional requirement that makes bottle serialize much faster.

@defnull maybe I can contribute something more

defnull:
Bottle is a lot faster than flask, this is also visible in other benchmarks. And yes, `ujson` is automatically picked up, if present, and used for faster json processing. That's a cheap boost for json-heavy APIs.

I'm not sure if this would be fair in a benchmark, though. Bottle is advertised as a zero-dependency framework, after all. And already faster than most ;)

waghanza:
@defnull Why ? `Flask` and `bottle` are **python** framework and their are running **both** under `gunicorn` and `meinheld`

waghanza:
@agalera In fact, we wan this **benchmark** as _simple_ as possible, as for now we are just using 3 rules, see https://github.com/the-benchmarker/web-frameworks/issues/221#issue-328459520

defnull:
Look at `Bottle._cast` for example: https://github.com/bottlepy/bottle/blob/master/bottle.py#L1028
It's a mess, way too long, complicated and full of hacks, but it's fast.
In Flask/Werkzeug, the same functionality is probably spread across multiple modules, and requires dozens of classes, method calls, wrappers and computed properties. 

Or `cached_property()`: https://github.com/bottlepy/bottle/blob/master/bottle.py#L234
It's a computed property that replaces itself with the value, once computed, so Python can find it in `__dict__` and won't bother executing the property again. It's a small detail, but when you try to process 3k requests per second, not having to parse a header multiple times pays off.

waghanza:
I understand the performance gap, but it's always useful to have some figures.

PR #353, opened 2018-09-07
scala/akkahttp: Running outside of sbt

whiplash:
I don't understand why `sbt run` is used though the uberjar is created via the `assembly` command.

### DIFF
--- a/scala/akkahttp/Dockerfile
+++ b/scala/akkahttp/Dockerfile
@@ -10,4 +10,4 @@ RUN sbt assembly
 
 EXPOSE 3000
 
-CMD sbt run
+CMD java -jar target/scala-2.12/server_scala_akkahttp


waghanza:
I did not have `scala` skills ... but I will help :stuck_out_tongue_closed_eyes:

waghanza:
@whiplash What will be the **advantage** to run outside of `sbt` ?

whiplash:
`sbt` is a java process itself. It uses some memory and cpu. When developing it is useful to do `sbt run` after compilation and see that your project runs ok, but in production you would create an executable `jar` and run it on your jvm by itself.

Hope this helps.

waghanza:
`crystal` clear for me :stuck_out_tongue_winking_eye: 

4ill merge if test passes ;-)

fyi you can run test on your `workstation`, just need `crystal` and `docker`
~~~sh
shards install
shards build --release
export FRAMEWORK=akkahttp
bin/neph ${FRAMEWORK}
crystal spec
~~~

waghanza:
| |         Average |  50th percentile |  90th percentile |  99th percentile | 99.9th percentile | Standard deviation |Requests / s |      Throughput |
|----|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|---------:|
| sbt | 184.77 ms | 8.43 ms | 23.10 ms | 4605.44 ms | 7911.87 ms | 786570.67 | 63576.67 | 136.58 MB |
| java | 188.79 ms | 8.33 ms | 25.79 ms | 4688.89 ms | 7940.66 ms | 803354.67 | 61694.33 | 132.45 MB |

:tada: 

As the **benchmark** are running on  `docker` (this will **change**) only **50th percentile** is _significant_

whiplash:
Oh that's unfortunate. I was expecting better performance. Oh well. I guess one can also tweak JVM parameters a little bit (for example more memory or something).

Thanks anyways.

waghanza:
@whiplash I understand, but as `docker` is used (for **flexibility**, not **performance**) we could **NOT** take any definitive conclusion.
The only understandable metric is that  **java** **COULD** increase performance

OvermindDL1:
Speaking of, we still need a way to run on bare metal for actual performance testing.

waghanza:
@OvermindDL1 sure, this is planned ... I think that there will be enough fixes at the end of septembre, then I plan to add support to run on **non-development** environments
+ public cloud (`digitalocean` for the official way, but helpers for `aws`)
+ private cloud (`cloudstack` / `openstack`)
+ bare metal (but **harder**, since creation should be **sequential**, e.g : the main advantage of cloud is that we can parallelize creation)

PR #364, opened 2018-09-10
Update nephp to 0.2.6

tbrand:
I've fixed the broken neph configuration.
I just make it works but I'll add few functions after merging this.
For example, specifying multiple jobs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -58,7 +58,7 @@ env:
 
 script:
   - shards build
-  - bin/neph ${FRAMEWORK}
+  - bin/neph ${FRAMEWORK} --mode=CI
   - crystal spec
 
 notifications:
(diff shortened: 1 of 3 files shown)


tbrand:
@waghanza ping

waghanza:
I still have
~~~crystal
Only one job is supported yet.
~~~
when **building** with
~~~sh
shards update
bin/neph sinatra flame
~~~

waghanza:
but I actually have **output** on my `console`  :+1:

waghanza:
seems to `mess-up` the **results**, please :warning: do not merge

tbrand:
@waghanza Yeah that's the problem. I'll fix it but not now.
I would like to merge this if it works fine without the function.

I'll work on fixing neph and send a patch to update it in the future.

waghanza:
@tbrand I think I find a solution #365

waghanza:
@tbrand I notice that **build** time increase comparing to `neph` `0.2.3`

PR #394, opened 2018-09-26
correct url rocket.rs

kocoten1992:


waghanza:
Thanks for notification @kocoten1992. I've see the change and I've fixed it
@see https://github.com/the-benchmarker/web-frameworks/blob/2018.9.26/FRAMEWORKS.yml#L140

Issue #400, opened 2018-10-05
Marking flame as experimental


waghanza:
Hi @AlexWayfer,

The actual version of `flame` (on **rubygems**) is `4.18.1`.

I prefer to show only framework that officially available.

I'll then mark `flame` as **experimental** (not show by default). 

Be sure that when `5` will be release, I'll re-enable it on results.

OK for you ?

Regards,

waghanza:
@AlexWayfer sure, but using `gem search -r` or https://rubygems.org/gems/flame display `4.18.1` by **default**, and in term of representation display result of `flame` (because of `5.0.0rc5`) could lead to misunderstanding (if read is not a `ruby` dev or a **non-technical** person)

I'll remove flame (from results) and re-enable after release :stuck_out_tongue: 

Thanks for understanding

AlexWayfer:
> I'll remove flame (from results) and re-enable after release stuck_out_tongue

Or, you can use Flame 4, and we will update version after 5 release. It's not too bad, just performance problems for thousands of routes.

waghanza:
Sure, but it needs to downgrade the implementation.

If you want to downgrade it, I can give invite you on my fork to collaborate :heart:

AlexWayfer:
> If you want to downgrade it, I can give invite you on my fork to collaborate heart

I just can make a PR, I guess. To this repo or to your fork, as you want.

waghanza:
Ok. On this repo then :+1:

PR #407, opened 2018-10-05
mark flame as experimental

waghanza:
Hi,

Following #400, I prefer marking `flame` as **experimental**, since non-default when doing a `gem search`.

I'll re-activate when release **5.0** will be out :tada: 

Regards,


waghanza:
rejected by of #410

Issue #404, opened 2018-10-05
Update http4s


waghanza:
Hi @mfirry,

We are using `0.0.1-SNAPSHOT` for [http4s](http://http4s.org).

Do you mind opening a `PR` to use `0.18` ?

Regards,

mfirry:
Hi @waghanza,
actually you're using `0.18.17` as you can see from the very first line of the `build.sbt` file.
I know it might look weird but `sbt` lets you define variables and use them later in the build file (for example `"org.http4s" %% "http4s-dsl" % Http4sVersion`.

So you're good to go, I think.

waghanza:
Ah ok, I was on the wrong version number 🤣

waghanza:
But there is still #401 if you want

waghanza:
And the version referenced
https://github.com/the-benchmarker/web-frameworks/blob/master/FRAMEWORKS.yml#L159
is wrong

PR #413, opened 2018-10-07
Addresses #404

mfirry:
Just changing http4s version number.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ bin/benchmarker [tools]
 ## Results
 
 <!-- Result from here -->
-Last update: 2018-09-25
+Last update: 2018-09-26
 ```
 OS: Linux (version: 4.16.11-100.fc26.x86_64, arch: x86_64)
 CPU Cores: 8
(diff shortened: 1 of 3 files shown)


FRAMEWORKS.yml
```diff
@@ -156,7 +156,7 @@ scala:
     version: "10.0"
   http4s:
     website: http4s.org
-    version: "0.0"
+    version: "0.18.17"
```
waghanza: Could you please specify X.Y format (e.g : 0.18)

It is more flexible, since modifying this file (and regenerate results) after each patch upgrade, will be hard to maintain

FRAMEWORKS.yml
```diff
@@ -156,7 +156,7 @@ scala:
     version: "10.0"
```
waghanza: You have upgrade to 10.1, thanks, but don't forget to upgrade conf

waghanza:
Thanks @mfirry

mfirry:
Sure. Done.

waghanza:
I've probably done a mistake

Issue #331, opened 2018-08-29
Jester failure


waghanza:
Hi @dom96,

`jester` is failing on our last run, https://travis-ci.com/the-benchmarker/web-frameworks/jobs/142574924

Could you get me a quick feedback on the error message on this line https://travis-ci.com/the-benchmarker/web-frameworks/jobs/142574924#L754 ?

Regards,

dom96:
That's really strange. All I can say is: the dockerfile for techempower is working fine. Try checking what the differences are?

waghanza:
The only diff I found (on `Dockerfile`) is that techempower copy all file in dir, and we only pick somes
+ server_nim_jester.nim
+ server_nim_jester.nimble

I've not enough skill to know what files are needed :stuck_out_tongue: 

The error is :

~~~nim
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
~~~

dom96:
Yeah, that error isn't very helpful sadly. Try using techempower's dockerfile exactly?

PR #429, opened 2018-10-10
update jester to 0.4.1

waghanza:
Hi @dom96,

`jester` was de-activated here.

This `PR` fix also #331

I also use `0.4.1` inspired by https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Nim/jester

Regards,

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -99,9 +99,9 @@ crystal:
     website: spider-gazelle.net
     version: "1.1"
 nim:
-  # jester:
-  #   github: dom96/jester
-  #   version: "0.4"
+  jester:
+    github: dom96/jester
+    version: "0.4"
   mofuw:
     github: 2vg/mofuw
     version: "2.0"
@@ -147,13 +147,13 @@ rust:
   rocket:
     website: rocket.rs
     version: "0.3"
-# elixir:
-#   phoenix:
-#     website: phoenixframework.org
-#     version: "1.3"
-#   plug:
-#     github: elixir-plug/plug
-#     version: "1.6"
+#elixir:
+#  phoenix:
+#    website: phoenixframework.org
+#    version: "1.3"
+#  plug:
+#    github: elixir-plug/plug
+#    version: "1.6"
 cpp:
   evhtp:
     github: criticalstack/libevhtp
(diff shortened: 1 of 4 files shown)


dom96:
Cool. Is it working for you now?

nim/jester/Dockerfile
```diff
@@ -1,17 +1,10 @@
-FROM gcc:latest
-
-RUN apt update && apt install -y libgc-dev
-
-ENV CHOOSENIM_NO_ANALYTICS 1
-ENV CHOOSENIM_CHOOSE_VERSION #0c683d28bbd5
-RUN curl https://nim-lang.org/choosenim/init.sh -sSf | sh -s -- -y
-ENV PATH $PATH:/root/.nimble/bin
+FROM nimlang/nim
```
dom96: You should pin to Nim 0.19.0 in case something breaks in the future.

FRAMEWORKS.yml
```diff
@@ -99,9 +99,9 @@ crystal:
     website: spider-gazelle.net
     version: "1.1"
 nim:
-  # jester:
-  #   github: dom96/jester
-  #   version: "0.4"
+  jester:
+    github: dom96/jester
+    version: "0.4"
```
dom96: *0.4.1

waghanza: sure, but in fact this is use on `README` (to display version of each framework).
To clarify/simplify maintainance, I only display them on X.Y format

waghanza:
@dom96 we plan to show only some frameworks by default (`native`, `stable`, with a certains features ...)

this mean `jester` (and `httpbeast`) will **not** be show by default (use `--experimental` to show results)

what do you think bout that ?

we are talking about that here https://github.com/orgs/the-benchmarker/teams/web-frameworks/discussions/2

dom96:
I can understand httpbeast, but why would jester be excluded?

waghanza:
@dom96 because of `version`, <1. and it's **not** an exclusion

waghanza:
@dom96 thinking about it, you're right.
Not showing some framework because of a version contraint will disable many here.

We have plently of frameworks un <1, and their are used in production.

So I've changed my mind, OK to keep `jester` and OK for `httpbeast`

dom96:
Awesome :)

waghanza:
@dom96 I've update the results https://github.com/the-benchmarker/web-frameworks#full-table with `jester`

PR #442, opened 2018-10-18
use nim 0.19 on mofuw

waghanza:


2vg:
i can't see issues list on this repo.
already bug fixed, so this change is unnecessary.

waghanza:
Sorry, my mistake.

I've disabled issue (not on purpose).

The issue is #443

Issue #440, opened 2018-10-17
Display language version


waghanza:
Hi,

Version of each **frameworks** is now displayed on `result` _table_.

It could be interesting to display also the version of each **language** in use.

Actually `docker` (**development**) is `used`, so pining a version is easy (tag in `Dockerfile`).

However, each `provider` (in fact `os` behind) has a specific version.

For example, **ruby** is available on :
+ `2.5` for `docker`
+ `2.4` for `amazonlinux`
+ `2.3` for `debian`
+ `2.4` for `rhel` and `centos`

@the-benchmarker/web-frameworks What do you think ?

Regards,

PR #444, opened 2018-10-18
Pin all containers

waghanza:
Hi,

This  `PR` pins all **containers** (`docker`) versions

It add ability to display language version on `README`

On production we will **HAVE TO** display version

Related to #440 

Regards,


OvermindDL1:
> @OvermindDL1 how can I specify any version (of cpp) for evhtp ?

It's already 'pin'ned.  In the CMake script the HunterGate call?  That version and hash value locks the versions, it will not nor can it ever change until that itself is changed.

waghanza:
Sorry, I think I am unclear

We already have specified version used for `evhtp`

I want now to specify version of  `C++` we use

OvermindDL1:
> I want now to specify version of C++ we use

That is specified in the CMake toolchain via options passed by the CMakeLists.txt file at:  https://github.com/the-benchmarker/web-frameworks/blob/master/cpp/evhtp/CMakeLists.txt#L14-L16
```cmake
set(CMAKE_CXX_STANDARD 11)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
set(CMAKE_CXX_EXTENSIONS OFF)
```

ras0219-msft:
>It is to ensure completely reproducible builds unlike other broken dependency management things (*cough*vcpkg*cough*).

If you keep using the same vcpkg commit, you should get identical builds -- if you found a case where this doesn't happen, please let us know because that's a serious bug!

OvermindDL1:
> If you keep using the same vcpkg commit, you should get identical builds -- if you found a case where this doesn't happen, please let us know because that's a serious bug!

The issue with vcpkg from the last time I touched it (a while back, entirely feasible this is fixed) is that it is managed outside of the source tree, thus different computers could have different versions of dependencies, where Hunter ensures it is identical everywhere while still globally caching it based on the version, build flags, etc... etc... to prevent repeated rebuilding.  In addition, hunter requires nothing else *but* the build system, no other weird software to install since if you have the build system (cmake, can pretty well be considered 'the' standard now) then everything 'just works', in addition across toolchains (last I tried vcpkg it also had issues with compiling for the desktop then trying to compile for android for example, again, that might be fixed now).

waghanza:
@OvermindDL1 would it help (in term of performances) to enable it as `C++17` ?
https://stackoverflow.com/questions/45688522/how-to-enable-c17-in-cmake

OvermindDL1:
@waghanza Only if such things as Move-Constructors or so were used, which they are not, so I wouldn't personally see it doing so.  However it wouldn't hurt at all (backwards compatibility and all) if you wanted to do so, just replace the `11` with `17`.

OvermindDL1:
Oh also, your SO link is out of date, CMake 3.10 I think it was, added `17` as a fully supported flag with the appropriate feature tests.  Once a standard is 'standardized' then CMake bakes it in, when it's not 'standard' yet (like 20) then you need to feature test individual features you want instead of relying on a (still-changing) version.

PR #449, opened 2018-10-22
Specify language for kore (quickfix)

waghanza:
Hi,

Now that we display language version in results, the `language ` key` in `FRAMEWORKS`.yml` is **required**

This `PR` add this to avoid build failures

@xieyuheng @OvermindDL1 which "version" of `C` are we using here ?

Regards,

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -246,3 +246,4 @@ c:
   kore:
     website: kore.io
     version: "3.1"
+    language: "??"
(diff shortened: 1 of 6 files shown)


OvermindDL1:
I looks like just C99, although setting a future version wouldn't hurt either (backwards compatibility and all).  Newer C versions don't change performance, just adds features (that if used might change performance).  C11 didn't really change much of anything about C99 other than just standardizing a few idioms, so calling in C11 is 'safe' I'd say.  C18 is too new to be in heavy use yet.

OvermindDL1:
Ah, just found it actually, Kore is built for C99.

Issue #483, opened 2018-10-23
Crystal not leveraging the 8 cores


drujensen:
Crystal doesn't have threads so its not leveraging the 8 cores. 

For Amber you can enable `port_reuse` in the settings:
```
  # Port Reuse: Amber supports clustering mode which allows to spin
  # multiple app instances per core. This setting allows to bind the different
  # instances to the same port. Default this setting to true if the number or process
  # is greater than 1.
  #
  # > Read more about Linux PORT REUSE https://lwn.net/Articles/542629/
  #
  settings.port_reuse = true
  #
  #
  # Process Count: This will enable Amber to be used in cluster mode,
  # spinning an instance for each number of process specified here.
  # Rule of thumb, always leave at least 1 core available for system processes/resources.
  #
  settings.process_count = 7
```

drujensen:
@waghanza Did the benchmarks get worse for Amber after changing the number of processes?  Is there some logs that I can find out what's going on?

waghanza:
@drujensen here I have results on my 4 cores

|  | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
|------------|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|
| **master** | 23.63 ms | 23.84 ms | 29.86 ms | 31.38 ms | 235.06 ms | 5656.33 |
| 8de479f4815a044157ddf3ac15cd2b17da70f060 (optimization) | 23.37 ms | 24.04 ms | 29.04 ms | 29.68 ms | 232.26 ms | 5438.67 |
| cbe94adc84b5f28e4bbf3e4ce979660a76a4bb7e (before optimization) | 23.28 ms | 22.83 ms | 29.52 ms | 30.42 ms | 175.41 ms | 5593.33 |

according to those metrics the global performance decrease then, but we can not take any assumption on them;  results are computed on local docke. i'm working on runing each **framework** on `droplet` in my branch https://github.com/waghanza/http-benchmark/tree/cloudify, after I've finished we could take any conclusion

I'm quite sure run on `droplets` will change this **benchmark** completely. for example, `laravel` is consider faster than `amber` which, I think, is impossible

I'm sure `amber` will climb :heart:

PR #284, opened 2018-06-28
[WIP] Implement SO_REUSEPORT

waghanza:
Hi,

This `PR` implements **SO_REUSEPORT**.

+ java
   + [x] act
+ nim
   + [ ] jester @dom96
+ c
   + [ ] kore @jorisvink
   + [ ] agoo-c @ohler55
+ rust
   + [ ] nickel @nickel-org
   + [ ] iron @iron
   + [ ] rocket @SergioBenitez
   + [ ] actix-web @actix
   + [ ] gotham @gotham-rs
+ ruby
   + [x] sinatra
   + [x] rack-routing
   + [x] hanami
   + [x] agoo
   + [x] roda
   + [x] flame
   + [x] rails
+ objc
   + [ ] criollo @thecatalinstan
+ node
   + [x] foxify
   + [x] turbo_polka
   + [x] fastify
   + [x] hapi
   + [x] muneem
   + [x] restify
   + [x] polka
   + [x] rayo
   + [x] koa
   + [x] restana
   + [x] express
+ scala
   + [ ] akkahttp @akka
   + [ ] http4s
+ cpp
   + [x] evhtp
+ go
   + [ ] muxie
   + [ ] iris
   + [ ] gorilla-mux
   + [ ] gin
   + [ ] beego
   + [ ] echo
   + [ ] chi
   + [ ] fasthttprouter
   + [ ] gf
+ csharp
   + [ ] aspnetcore
+ elixir
   + [x] plug
   + [x] phoenix
+ swift
   + [ ] perfect
   + [ ] kitura
   + [ ] vapor
+ php
   + [ ] zend-framework
   + [ ] zend-expressive
   + [ ] slim
   + [ ] symfony
   + [ ] laravel
   + [ ] lumen
+ crystal
   + [x] onyx
   + [x] spider-gazelle
   + [ ] raze
   + [ ] amber
   + [x] kemal
   + [x] router.cr
   + [x] orion
   + [x] lucky
+ python
   + [x] japronto
   + [x] flask
   + [x] falcon
   + [x] responder
   + [x] cyclone
   + [x] starlette
   + [x] django
   + [x] sanic
   + [x] bottle
   + [x] tornado
   + [x] aiohttp
   + [x] vibora
   + [x] quart
   + [ ] quart

@OvermindDL1 This feature **COULD** be enabled on each framework https://github.com/tbrand/which_is_the_fastest/issues/69

Regards,

PS : Only `kemal` as in **WIP**

[...]

PR #484, opened 2018-10-23
optimize amber

drujensen:
This does several things to provide a better comparison:
- disable logging
- removes all unnecessary middleware
- enables port_reuse
- sets number of processes to 7 to leverage the 8 cores

This addresses issue #483

### DIFF
--- a/crystal/amber/Dockerfile
+++ b/crystal/amber/Dockerfile
@@ -6,8 +6,6 @@ COPY shard.yml ./
 COPY src src
 COPY config config
 
-ENV PROCESS_COUNT $(nproc)
-
 RUN shards build --release --no-debug
 
 CMD bin/server
--- a/crystal/amber/Dockerfile
+++ b/crystal/amber/Dockerfile
@@ -6,6 +6,8 @@ COPY shard.yml ./
 COPY src src
 COPY config config
 
+ENV PROCESS_COUNT $(nproc)
+
 RUN shards build --release --no-debug
 
 CMD bin/server
--- a/crystal/amber/config/application.cr
+++ b/crystal/amber/config/application.cr
@@ -75,7 +75,7 @@ Amber::Server.configure do |settings|
   # spinning an instance for each number of process specified here.
   # Rule of thumb, always leave at least 1 core available for system processes/resources.
   #
-  settings.process_count = 7
+  settings.process_count = ENV["PROCESS_COUNT"].to_i if ENV["PROCESS_COUNT"]?
   #
   #
   # PORT: This is the port that you're application will run on. Examples would be (80, 443, 3000, 8080)
--- a/crystal/amber/config/application.cr
+++ b/crystal/amber/config/application.cr
@@ -75,7 +75,7 @@ Amber::Server.configure do |settings|
   # spinning an instance for each number of process specified here.
   # Rule of thumb, always leave at least 1 core available for system processes/resources.
   #
-  settings.process_count = ENV["PROCESS_COUNT"].to_i if ENV["PROCESS_COUNT"]?
+  settings.process_count = System.cpu_count.to_i32
   #
   #
   # PORT: This is the port that you're application will run on. Examples would be (80, 443, 3000, 8080)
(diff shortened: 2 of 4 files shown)


waghanza:
I was trying to enable `reuse_port` on https://github.com/the-benchmarker/web-frameworks/pull/284, but seems to take time

I'm glad you did :heart:

waghanza:
@proyb6 the `amber` implementation on **tfb** use `1 CPU (core)` 

@drujensen we can not write a `static` number as `amber` process count. **official** rounds will run on 8 CPU env, but devs could want to run on 1 CPU. I've set process count **dynamic** (in `env`)

waghanza:
|   | Framework (Middleware) | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
| here | [amber](http://amberframework.org) (0.9) | 36.49 ms | 35.10 ms | 47.14 ms | 76.63 ms | 409.67 ms | 21775.00 |
| master| [amber](http://amberframework.org) (0.9) | 36.11 ms | 30.58 ms | 48.70 ms | 145.97 ms | 774.06 ms | 34394.67 | 

@drujensen For me this is more performant, since **Standard deviation** decrease. But to be sure we have to wait until running on `cloud`

PS : Those data are not calculated from the same environment as on `README`

drujensen:
@proyb6 The TechEmpower benchmarks is using a `run.sh` that all of the crystal projects are using.  I created that before we added the feature to Amber.  We just never went back and changed it to use this built-in feature.

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/amber/run.sh

drujensen:
@waghanza Thanks for merging.  Rust Nickel is bare bones.  Have you considered doing a Crystal Standard library version to match?  Or does this benchmark require a router?

proyb6:
@drujensen Is there an alternative libraries you would prefer to choose HTTP in C instead? As we all can use the standard libraries, we need a universal routers for Crystal web frameworks.

drujensen:
> As we all can use the standard libraries, we need a universal routers for Crystal web frameworks.

@proyb6 right.  Just wondering what this benchmark is trying to achieve. 

> However, when youy say that nickel is barebone, could you give me more details about that ?

@waghanza What I mean is that rust/nickel is using a very simple middleware to handle routing.  Just wondering what the requirements for this benchmark are.

For example, you could do something like this:
```crystal
require "http/server"

server = HTTP::Server.new do |context|
   context.response.print ""
end

server.listen("0.0.0.0", 8080, reuse_port: true)
```
 
I think this will pass this benchmark since there is no criteria for what should be returned from each path.   To leverage the `reuse_port`, you would need to launch multiple processes like this: https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/run.sh

You could also add some conditional logic based on path/method to provide a different response if that was part of the criteria.  This is the simplest way to perform routing.  You can see an example here:

https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/server.cr#L16

TechEmpower asks you if you are a fullstack/micro/platform to make it clear what's included when you perform comparisons between frameworks.

waghanza:
@drujensen  `router.cr` is also very simple, as `orion` ...

The goal here is :
+ To show performance (on a **real** usage)
+ To show how to write (according to each `language` best practices)
+ To show details (ressources comsumption, implementation choice ...)

proyb6:
In additional, adding Parallel.cr will parallelize if you like to add.
```
require "parallel"

pspawn do
   get "/" ...
end
```

waghanza:
@proyb6 seems to specific for me :stuck_out_tongue_closed_eyes: 

I mean parallelism is a `core` task (not a `framework` one), and I'm sure this will be on `crystal` `1.0`

drujensen:
@waghanza yeah, regarding router.cr and orion, both are very simple probably more on par with the nickel example.

Regarding goals, so there is no goal to have an apples-to-apples comparison? The goal is to compare real usage performance?  

Obviously, each will have more or less features that will impact the performance.  Amber out the box tries to do everything for you.  There are 10+ middleware included.  https://github.com/amberframework/amber/tree/master/src/amber/pipes 

I removed them, but maybe we should put them back because in a real usage situation, you would have logging, csrf, error, flash, session, static, and possibly cors if your building an API.  WDYT?

waghanza:
@drujensen Actually, we have 3 rules (endpoints needed).

We want to test frameworks on this basis, to have a fair (most we can) comparison, and then avoid apple-to-orange comparison

PR #486, opened 2018-10-23
Fix kore build

waghanza:
Hi @xieyuheng @zhangkaizhao,

I've merge too early some `PR`, `kore` was then failing.

Reverting change (not upgrade `docker` **image**) fix it

Is there any  reason `ubuntu` image was picked, instead of https://store.docker.com/images/gcc (for **example**)

@OvermindDL1 What do you think ?

Regards,


OvermindDL1:
Considering kore uses it's own very funky build system instead of something standard like make or cmake, it's hard to tell without seeing the build logs.  Do you have a link to the build logs for the failing build so I can take a look?

waghanza:
@OvermindDL1 sure => https://travis-ci.com/the-benchmarker/web-frameworks/jobs/153548670

I think the failure is cause by an upgrade to **ubuntu**  `18.10` => **gcc** `8.1`

I've pushed a update to **gcc** `8.2` and everythin is ok :tada:

OvermindDL1:
Huh, yeah that error just means a missing type, which is super weird, wonder if an 8.1 bug, I tend to use clang myself anymore...

waghanza:
I do not have skills on `C` to define which is better 😎 

I'll merge this then ans accept furthur modification

PR #492, opened 2018-10-24
Neph v0.2.8

tbrand:
I've fixed the neph.yaml for updating neph itself.
The latest version of neph support '--seq' option for sequential execution.
So I recommend you to run like this in CI.
```bash
neph --seq --mode=CI
```

It's also support `before` and `after` command execution.
Like this. https://github.com/tbrand/neph/blob/master/sample/neph.yaml#L13-L26

Thanks


waghanza:
@tbrand `foxify` build is failing

what is unexpected is that with `neph` `0.2.6`
~~~
shards install 
shards build --release
bin/neph foxify
bin/benchmarker foxify
~~~
works, but not with `0.2.8`

waghanza:
@tbrand I think I'll have to `revert` this (and re-merge later), even on a 8 core droplet this is failing, I checkout on the commit just before the merge, and it was OK.

PR #496, opened 2018-10-24
(re) enable plug

### LINKED ISSUE #494: Re-enable plug

waghanza:
Hi,

This `PR` fix #494 

`plug` has been **disabled** because of failure

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,7 @@ env:
     - FRAMEWORK=rayo
     - FRAMEWORK=turbo_polka
     - FRAMEWORK=kore
+    - FRAMEWORK=plug
 
 script:
   - shards build
--- a/.travis.yml
+++ b/.travis.yml
@@ -67,6 +67,7 @@ env:
     - FRAMEWORK=plug
 
 script:
+  - docker -v
   - shards build
   - bin/neph ${FRAMEWORK} --mode=CI
   - crystal spec
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -213,15 +213,15 @@ rust:
     website: rocket.rs
     version: "0.3"
     language: "nightly"
-#elixir:
+elixir:
 #  phoenix:
 #    website: phoenixframework.org
 #    version: "1.3"
 #    language: "1.7"
-#  plug:
-#    github: elixir-plug/plug
-#    version: "1.6"
-#    language: "1.7"
+  plug:
+    github: elixir-plug/plug
+    version: "1.6"
+    language: "1.7"
 cpp:
   evhtp:
     github: criticalstack/libevhtp
--- a/neph.yaml
+++ b/neph.yaml
@@ -8,7 +8,7 @@ main:
     - swift
     - rust
     - node
-    #- elixir
+    - elixir
     - cpp
     - scala
     - csharp
@@ -91,7 +91,7 @@ node:
 elixir:
   depends_on:
     - plug
-    - phoenix
+    # - phoenix
 
 cpp:
   depends_on:


OvermindDL1:
Do you know what error it got?  Travis doesn't seem to be running for it and I've been waiting quite a while...  ^.^;

For note, plug's version tends to upgrade in lockstep with elixir's version, so if elixir was upgraded but plug was not or vice-versa, that would cause an issue.  Plug is a base library put out by the Elixir dev's that upgrades in sync with Elixir itself.  It's essentially a standard library in Elixir but since it's web specific it's just not included in the base distribution.

waghanza:
Honestly, I remember, I've **disabled** `plug` a while ago since it was failing, so break CI

This `PR` is to re-eable it :heart: as disabling is **never** a reliable solution

I've also re-enable it on CI, it will automatically run at https://travis-ci.com/the-benchmarker/web-frameworks/jobs/154015721

All I can say is that I have 
~~~elixir
22:42:48.902 [info]  Application my_plug exited: exited in: MyPlug.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) function Plug.Cowboy.child_spec/4 is undefined or private
            (plug_cowboy) Plug.Cowboy.child_spec(:http, MyPlug.Router, [], [port: 3000, protocol_options: [max_keepalive: 5000000]])
            (my_plug) lib/plug/application.ex:15: MyPlug.Application.start/2
            (kernel) application_master.erl:277: :application_master.start_it_old/4
{"Kernel pid terminated",application_controller,"{application_start_failure,my_plug,{bad_return,{{'Elixir.MyPlug.Application',start,[normal,[]]},{'EXIT',{undef,[{'Elixir.Plug.Cowboy',child_spec,[http,'Elixir.MyPlug.Router',[],[{port,3000},{protocol_options,[{max_keepalive,5000000}]}]],[]},{'Elixir.MyPlug.Application',start,2,[{file,\"lib/plug/application.ex\"},{line,15}]},{application_master,start_it_old,4,[{file,\"application_master.erl\"},{line,277}]}]}}}}}"}
Kernel pid terminated (application_controller) ({application_start_failure,my_plug,{bad_return,{{'Elixir.MyPlug.Application',start,[normal,[]]},{'EXIT',{undef,[{'Elixir.Plug.Cowboy',child_spec,[http,'

Crash dump is being written to: erl_crash.dump...done
~~~
using  **latest** `plug`

OvermindDL1:
For note, for Elixir 1.7.x (1.7.4 as of right now) then Plug 1.7.x (1.7.1 as of right now) should be used.

OvermindDL1:
> ** (UndefinedFunctionError) function Plug.Cowboy.child_spec/4 is undefined or private

Whoa, this implies old code, that's the old child_spec format...

waghanza:
ok, but there is no `docker` image for `1.7.4` yet

OvermindDL1:
So in the Elixir plug Dockerfile:
https://github.com/the-benchmarker/web-frameworks/blob/master/elixir/plug/Dockerfile#L1
```Dockerfile
FROM elixir:1.7.3
```
It's using elixir 1.7.3 (1.7.4 is current if it matters), and the version of plug being used, specified at:
https://github.com/the-benchmarker/web-frameworks/blob/master/elixir/plug/mix.exs#L33
```elixir
{:plug, "~> 1.6.1"},
```
Is for Elixir 1.6, that line needs to be changed to `{:plug, "~> 1.7.1"},`, somehow someone updated Elixir without updating the core dependencies (newer deps generally work on older Elixir's to 2 or 3 major versions behind, but new elixir's require new dependencies).

waghanza:
@OvermindDL1 feel free to give a try :heart:

OvermindDL1:
> ok, but there is no docker image for 1.7.4 yet

Lol, I wonder who manages that, it's not the official Elixir group, 1.7.4 has been out for a while now.  Honestly should just use an `asdf` image and grab any specific elixir version (like `1.7.4-otp-21`) in it.  ^.^;

> @OvermindDL1 feel free to give a try heart

Should be able to just change the dependency in the mix.exs file to fix it I'd bet, I can't get to `git` right now, crazy busy at work...  ^.^;

waghanza:
stuck is `1.7.3`, not sure who update docker images
https://store.docker.com/images/elixir

since all tests pass, I'll merge, but I don't have enough skills in `elixir` to understand error messages, and to fix them :stuck_out_tongue_closed_eyes:

OvermindDL1:
> since all tests pass, I'll merge, but I don't have enough skills in elixir to understand error messages, and to fix them stuck_out_tongue_closed_eyes

Heh, the BEAM VM's error messages are certainly archaic but they are *very* detailed, elixir tries to pretty them up a lot (and it really does!), but there is still excess fluff to get to so it's just a learning to read it over time, but once learned it's usually very easy to figure out about any issue.  ^.^;

waghanza:
yeah, I also tried to learn `haskell` once, but as I more working on dynamic languages, it seems odd to me :stuck_out_tongue:

OvermindDL1:
> yeah, I also tried to learn haskell once, but as I more working on dynamic languages, it seems odd to me stuck_out_tongue

Heh, well the BEAM VM is entirely dynamic, every type, every call.  To compare it to, oh, Ruby, or Crystal, or something else OOP'ish, think of it as the BEAM VM has static function calls that anything can call, every function call can only operate on it's inputs and return outputs, everything side-effecting is a message (hidden behind function calls 99% of the time).  An instanced 'class' like thing on the BEAM is the 'process'/actor, everything is such an actor, just like with OOP classes you can instance millions of them if you want, except that BEAM actors all run concurrently and in parallel with each other, able to scale arbitrarily not just on a single node/computer with many *many* cores but even across systems, making it scale very well both horizontally and vertically.  The memory model is that messages are copied between actors, each actor's memory is entirely segregated from all other actors, and since most actors (like a web request) are short lived then a garbage collection just never runs, and for long-living actors then the GC algorithm is a multi-tiered combination of generational, mark-and-sweep, and a couple of others all run based on various conditions, and there is no such thing as stop-the-world collection, at worst a single actor will be stopped for nanoseconds to perform a collection, and since all data can only reference older data, there are no memory cycles unlike in OOP languages, than that means that garbage collections don't have to worry about cycles and collecting is *extremely* fast compared to about any other language out.

The BEAM itself has two JIT-ish engines, the primary one just called the BEAM is optimized for IO through-put, it is designed for pushing IO data faster than any other language can do with the same reliability (a beam VM should never *ever* fail, unlike, say, a C or go node or so), the second one is HiPE, an optimizing compiler that performs IO slower but great if you have something math heavy (although slaving out such functionality to a NIF or Port is usually better as then you can use *any* language that can be optimized for whatever functionality you want, the BEAM VM is a fantastic 'Glue' language, designed to interact with any other language whatsoever in one form or another).  Even each individual module file (collection of static functions) can use whichever JIT'ish thing they want, though calling function cross-backend (BEAM<->HiPE) does incur overhead as well so you only want to do it when it is worth it (pretty much everyone sticks to the BEAM since NIF's are so popular via Rust or C++ or Ports with Python or so).  The BEAM VM is, in essence, perfect for the web, originally designed for telephone switching systems where uptime and throughput is paramount, so it is designed to never *ever* fail (which is why BEAM users have the 'Let it Fail' philosophy, since the system recovers safely and quickly), so the reliability combined with the throughput ended up being a perfect fit for the web (even though the BEAM predates the web).

There are quite a number of languages on the BEAM VM (it is *really* easy to compile to), first-class languages like erlang (prolog'y), elixir (ruby'ish), lfe (lisp'y), along with a *huge* host of community languages such as one based on Elm, a customized lua variant designed for the intense concurrency/parallel the BEAM supports, and about 40 others, in addition to some being made like for Idris and OCaml backends (both incomplete, mostly test projects that someone should finish up sometime).  Erlang was the original language on the beam, Elixir has recently taken over as the most used.  All BEAM languages have interoperability between each other and you can mix and match as you wish.  The packaging and build systems are all solid and everything just tends to 'work'.

The BEAM is not good for things that are not long-living, it has some spool-up time and it is designed to essentially act as a kind of mini-OS on its own (though it doesn't touch anything outside of it, just in the reliability and scheduling it does), so it's not good to act as something quick like a python or shell script (although it still can, it just takes a second to spool up first).

It's a fascinating system to learn however!  As long as Crystal supports the C FFI standards than it can be used as a NIF into the system (a requirement for NIF's however is *NO* GC pauses in the NIF, so you need to ensure that somehow if you do), or just use a Port, which works with everything though has more overhead than a NIF (a NIF is almost as fast as a built-in native code call), any language would work to interact with the BEAM as a Port.

PR #508, opened 2018-10-28
Run on (public) cloud

waghanza:
Hi @proyb6 @doanguyen,

I'm opening this `PR` a a starting point to **cloudification**

The need here is to create `cloud` `instances` to replace `docker` (**local**) `containers` 
@OvermindDL1 You can notice, I have designed this a `cli` tools, the **goal** behind is that we can extend this with `aws`, `bare-metal` ...

The `cli` tool (**bin/droplet**) is just a wrapper on https://github.com/digitalocean/doctl and configuration are based on https://cloudinit.readthedocs.io/en/latest/

@proyb6 @doanguyen if you want to contribute,  it will be a pleasure

Regards,

### DIFF
--- a/ruby/config.yml
+++ b/ruby/config.yml
@@ -1,0 +1,60 @@
+providers:
+  aws:
+    version: 2.4
+    user: ec2-user
+    # amazon linux 2 use cloud-init 18.2
+    config:
+      runcmd:
+        - amazon-linux-extras install ruby2.4
+        - yum --assumeyes install ruby-devel
+        - yum --assumeyes install @development
+        - yum --assumeyes install rubygems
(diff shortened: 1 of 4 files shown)


waghanza:
Nop. Because sieger is also on digitalocean 😊

OvermindDL1:
With a fiber interlink?  You are certain they are in the same rack?  Even being in different racks can add substantial limitations.  In addition that most setups like that are designed for low-connection count with high throughput (think database access) unlike many *many* short-lived connections as this test is doing.

waghanza:
Nop. I don't think they have, but it's better thanks a local Docker 😊

I'll also investiguate on AWS ...

PR #513, opened 2018-10-28
Switch to go modules

waghanza:
Hi @appleboy,

I've followed you advice and use [go modules](https://github.com/golang/go/wiki/Modules)

@kataras I have not been successful in using  [go modules](https://github.com/golang/go/wiki/Modules) in `iris` (when doing so `iris` require `blackfriday). any ideas ?

Regards,


kataras:
Hello @waghanza , try to put `v11.0.2+incompatible` as shown below:

incop

Issue #292, opened 2018-07-16
Choose fastest mode (python)


waghanza:
Hi,

There is **plenty** methods for `python`  _deployments_ .

As this project aim to display the fastest, it's accurate to check which way is the fastest (at least for python frameworks) :

* [ ] Built-in (or recommanded) webserver
* [ ] [Meinheld](http://meinheld.org/)
* [ ] [asyncio](http://asyncio.readthedocs.io)
* [ ] [uvloop](http://uvloop.readthedocs.io)

The purpose of this `PR` is to get the performance factor between this way for all `python` **frameworks**.

Regards,

waghanza:
### django (implementations 3443ea3)

| Language (Runtime)        | Framework (Middleware)    |    Requests / s |         Latency |   99 percentile |      Throughput |
|---------------------------|---------------------------|----------------:|----------------:|----------------:|-----------:|
| python                    | django                    | 852.00 | 169411.00 | 2167552.33 | 0.41 MB |
| python                    | django_meinheld           | 11440.67 | 90586.00 | 244103.00 | 10.85 MB |
| python                    | django_gevent             | 5718.67 | 137400.67 | 3113876.33 | 5.22 MB |
| python                    | django_asyncio            | 1511.67 | 610000.00 | 1512050.67 | 1.33 MB |
| python                    | django_waitress           | 1444.00 | 69772.00 | 126216.67 | 0.81 MB |

Conclusion : `meinheld` is the `fastest`

PS : As `django` is a **WSGI** framework, `uvloop` is not available

waghanza:
Closing in favour of https://github.com/the-benchmarker/web-frameworks/issues/1481

PR #545, opened 2018-10-31
Adding Responder

greed2411:

   + [x] `travis.yml`, for CI
+ [x] A `Dockerfile` exists ?
+ [ ] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~

I don't have crystal with me,  so I can't promise anything about all test cases.
@waghanza , can you see if this is working?


waghanza:
@greed2411 Thanks for your **awesome** work :tada: 

What I've learn during this competition, is that there is a dozen (at least) of appservers in `python`

I've used `gunicorn` to avoid too much diff between implementations (and as **tfb**) is doing, but #292 is still open and I'll then try to pick the right appserver for a maxium of implementations (I've just pick `gunicorn` because it's what I know best)

greed2411:
Idk which one is the recommended appserver for `responder`, but I think it uses something called `uvicorn`. Also it's a `ASGI` framework. So I think it might do better on `uvloop` than on `gunicorn`. Try it out and let me know please.

Also I've never written apps that go past localhost or heroku.  Therefore thanks for making these scripts production ready.😁

tomchristie:
You’ve got the right deployment here, it’s just missing `-w 4` to increase the number of worker processes. (Probably between 1-2 workers per core would be a good number.)

waghanza:
@ greed2411 in fact, I use `gunicorn`, but the worker inside is `uvicorn`

@ tomchristie  we use 1 worker personnes core on all `python` **frameworks** (with reuse_port ...)

The problem here is that I have a `null` instead or an **empty string** on GET /

waghanza:
@greed2411 I've fixed null issue; `responder` use `marshmallow`, so I have to create a custom type based on `marshmallow` definition

tomchristie:
Dockerfile here needs to use nproc workers when running gunicorn.

waghanza:
Yep forger it. Thanks for reminder

tomchristie:
Cool. Last thing is setting the log level to warning on all gunicorn cases. Python’s logging is a little slow, so should make a bit of a difference. 😇

waghanza:
I made `gunicorn` confs uniform, fix for #549
* listening 0.0.0.0:3000
* using `SO_REUSEPORT`
* use 1 work per core

waghanza:
@tomchristie I'm currently computing results, but I had to disable `responder` -> seems working on CI (docker 17.09) but not on digitalocean (docker 18.06)

having **null** instead of **empty strings**

waghanza:
@greed2411 I had to disable `responder`  in the last update, do you found any fix for the above error ?

greed2411:
Not really. I don't know how to fix the `null` thing. Was hesitant about raising issue to kennethreitz about it. If you wish, we can raise a ticket about it in his repository. What do you say?

waghanza:
@greed2411 sure, go ahead, you shouldn't be afraid to do so :stuck_out_tongue_closed_eyes:

waghanza:
@greed2411 this issue was fixed by @kennethreitz

PR #554, opened 2018-11-02
be more strict on python versionning

waghanza:
Hi,

This `PR` specify strict versions of each dependencies for `python` **based** _frameworks_

This **COULD** be done because of http://pyup.io and it **COULD** avoid some side-effects
https://pyup.io/account/repos/github/the-benchmarker/web-frameworks/

Regards,

### DIFF
--- a/crystal/kemal/src/server.cr
+++ b/crystal/kemal/src/server.cr
@@ -18,4 +18,7 @@ post "/user" do |env|
 end
 
 Kemal.config.env = "production"
-Kemal.run
+Kemal.run do |config|
+  server = config.server.not_nil!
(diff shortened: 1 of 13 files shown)


greed2411:
so we are sticking to py36 and all the frameworks tested with the versions they were built with? That's equal treatment. 

But it may be punishing towards new async frameworks which leverage few things from py37,so it's not necessarily fair.

waghanza:
not at all @greed2411 

in fact, I've found this service https://pyup.io that will stick to X.Y.Z `python` **packages**

anythin to do with `python` (the **language**) version

PR #557, opened 2018-11-04
update router.cr for improved performance

stakach:
See: https://github.com/tbrand/router.cr/pull/29

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -140,7 +140,7 @@ crystal:
   router.cr:
     github: tbrand/router.cr
     version: "0.2"
-    language: "0.26.1"
+    language: "0.27.0"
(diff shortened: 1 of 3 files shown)


proyb6:
version.cr also need update from 0.2.5.

waghanza:
@proyb6 I don't understand

proyb6:
> @proyb6 I don't understand

@waghanza Sorry, I was referring to this need to be update as well as ```shards install``` notified ```0.2.5 at 0.2.6```
https://github.com/tbrand/router.cr/blob/master/src/router/version.cr#L2

tbrand:
Oh that's my bad. I'll fix it.

tbrand:
It's fixed on 0.2.7.

Issue #604, opened 2018-11-08
Foxify is faling on route 2


waghanza:
Hi @ardalanamini,

`foxify` return a **404** on https://github.com/the-benchmarker/web-frameworks/blob/master/node/foxify/app.js#L13
@see https://travis-ci.com/the-benchmarker/web-frameworks/jobs/157156855

Regards,

ardalanamini:
@waghanza what is the request url for this test?

waghanza:
@ardalanamini `/user/0` on test suite, using `0` as URL param

waghanza:
@ardalanamini  unfortunately, this is not working
I am on :
+ node  `10.12.0`
+ npm `6.4.1`
+ foxify `0.10.8`

it return **404** on `/user/0` using `.get("/user/:id", (req, res) => res.send(req.params.id))`

PR #605, opened 2018-11-08
Update Foxify

ardalanamini:


waghanza:
Fix #604

waghanza:
@ardalanamini You are awesome :heart: 

However, could you made updates `PR` , from branches

I can, then modify them :stuck_out_tongue: in case of other modification to do

ardalanamini:
Fixed #604

ardalanamini:
Foxify by default creates clusters according to number of cpu cores
so `.set("workers", 1)` says not to that
this option is useful when using pm2

PR #606, opened 2018-11-08
Fix responder (temporary fix)

waghanza:
Hi @kennethreitz,

I can see that line 16 of `api.py` importing **debug** module of **starlette** was failing my tests
https://github.com/kennethreitz/responder/blob/v1.1.1/responder/api.py#L16

I can see that this line in not present anymore in **master**

I've the switch to **master** (even if this is temporary), do have a working `CI`

I'll revert this until `responder` will be pushed on pypy

Regards,


greed2411:
Thanks for asking him! @waghanza

PR #614, opened 2018-11-11
Better Agoo settings

ohler55:
These changes along with some changes in Agoo itself should improve Agoo performance. Comments were also added to the app to make it clear why the setting were selected.

---------------------

   + [ ] `neph.yml`, for [neph](http://tbrand.github.io/neph) job processing
   + [ ] `FRAMEWORKS.yml`, for **frameworks** list
   + [ ] `travis.yml`, for CI
+ [ ] A `Dockerfile` exists ?
+ [ ] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~


ruby/agoo/Dockerfile
```diff
@@ -2,8 +2,9 @@ FROM ruby:2.5.3
 
 WORKDIR /usr/src/app
 
-COPY Gemfile app.rb ./
+COPY Gemfile app.rb index.html ./
```
waghanza: why adding an `index.html` ?

ohler55: It is an empty file. Since the `/` path returns a blank page it is really a static asset or can be considered as one. Agoo handles static assets directly without involving Ruby so is much faster.

waghanza: I understand, but it is kind of out ouf scope here, I think
What we wan to compare is **framework** performance -> with dynamic content not static

Can be ruby invoked in `/` ?

ohler55: I should mention that the previous version of Agoo considered an empty file a 404. The current does not so the `/` can be a static empty asset.

The app.rb explains some of this. I tried to make it clear.

waghanza: I understand, but all **frameworks** here invoke the `runtime` behind when calling `/`, it will not be a fair comparison, as the test protocol in to call `runtime`

ohler55: There is a comment in the app.rb file. By un-commenting the line that sets the route for `/` it will use Ruby. My apologies, I thought the inclusion of `/` was meant as a test for static assets. Apps usually have a mix so I thought it was normal.

ohler55: I wasn't sure what action I should take. Should I uncomment the offending line or were you planning to?

waghanza: > Should I uncomment the offending line or were you planning to?

Yes

> IMHO, it seems that servers with higher throughput are handicapped when calculating latency

Absolutely. That's way there is two tables :
+ latency
+ req/s

ohler55: Maybe I misunderstood. It looked like the latency was calculated on the same runs as the throughput which would mean latency was for different loads for each framework. If the latencies were calculated under the same load for each framework then my mistake. I did see the two table though.

Anyway, offending use of static assets has been disabled.

waghanza: All metrics are computed from the same runs. I have taken inspiration from https://github.com/TechEmpower/FrameworkBenchmarks

I'll have to explain this a bit more :stuck_out_tongue: now this process is quite stable, now

ohler55:
I noticed a checkin that is titled 'use all cores'. You may find that in this case where the processing load is light that not spawning multiple workers is better. Agoo already creates extra threads for read and writes so adding extra threads for processing only takes CPU time for context switching instead of for doing real work. Maybe you are just experimenting though.

ohler55:
Not surprised. Thats what happens when I added changes to help with single process performance for light weight ruby apps and then too many workers are created. I guess I'll revert the changes or make Agoo smarter so that it is harder to misconfigure it.

waghanza:
Yeah :+1:

This also a feature of this tool. Helping frameworks author to decide about code design.

In the near future, historical data will be stored on slicingdice, si we can check wheter version X is more or less performing than version Y

PR #624, opened 2018-11-13
Add GF(GoFrame) framework

johng-cn:
This PR adds another go framework `GF(GoFrame)` to this amazing project.

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -191,6 +191,10 @@ go:
     website: godoc.org/github.com/kataras/muxie
     version: "1.0"
     language: "1.11"
+  gf:
+    website: gfer.me
+    version: "1.1.994"
+    language: "1.11"
   gin:
     website: gin-gonic.github.io/gin
     version: "1.3"
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -193,7 +193,7 @@ go:
     language: "1.11"
   gf:
     website: gfer.me
-    version: "1.1.994"
+    version: "1.1"
     language: "1.11"
   gin:
     website: gin-gonic.github.io/gin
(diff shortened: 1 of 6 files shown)


FRAMEWORKS.yml
```diff
@@ -191,6 +191,10 @@ go:
     website: godoc.org/github.com/kataras/muxie
     version: "1.0"
     language: "1.11"
+  gf:
+    website: gfer.me
+    version: "1.1.994"
```
waghanza: could you use the format : X.Y -> it will be more readable for our audience :+1:

waghanza:
@johng-cn Could you use **go modules** ?

FRAMEWORKS.yml
johng-cn: ok, done

johng-cn:
@waghanza yes, it used go modules, you can see it in `/go/gf/go.mod`

waghanza:
@johng-cn should not `go.mod` have 2 requires ? as their is 2 imports in `main.go`

johng-cn:
@waghanza As `GF` for this test codes, only one `require` in `go.mod` is neccessary, although there're two `import`s in `main.go`, but they're in the same package. You can see this's similar of `beego` framework in `/go/beego/go.mod` and `/go/beego/main.go`.

waghanza:
@johng-cn I've update results https://github.com/the-benchmarker/web-frameworks#full-table :tada: 

You might be disapointed :stuck_out_tongue: results are not (yet) **production-ready**

johng-cn:
@waghanza Thanks , I've seen the results. That's great, it means `GF` do need some performance improvement. Comparison makes progress. 😄

johng-cn:
@waghanza Not yet, lots of work to do. I'll remember to tell you if such's done.

PR #626, opened 2018-11-14
Agoo c

ohler55:
Agoo-C uses the same core code as Agoo for Ruby.

------------------

   + [ ] `neph.yml`, for [neph](http://tbrand.github.io/neph) job processing
   + [ ] `FRAMEWORKS.yml`, for **frameworks** list
   + [ ] `travis.yml`, for CI
+ [ ] A `Dockerfile` exists ?
+ [ ] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~

### DIFF
--- a/c/agoo-c/simple.c
+++ b/c/agoo-c/simple.c
@@ -1,0 +1,50 @@
+// Copyright 2018 by Peter Ohler, All Rights Reserved
+
+#include <stdio.h>
+#include <stdlib.h>
+
+#include <agoo.h>
+#include <agoo/log.h>
+#include <agoo/page.h>
+#include <agoo/server.h>
+
+static int	user_off = 6;
+
+static void
+user_handler(Req req) {
+    agoo_respond(req, 200, req->path.start + user_off, req->path.len - user_off, NULL);
+}
+
+static void
+empty_handler(Req req) {
+    agoo_respond(req, 200, NULL, 0, NULL);
+}
+
+int
+main(int argc, char **argv) {
+    struct _Err	err = ERR_INIT;
+    
+    agoo_init("simple");
+
+    // Set the number of eval threads.
+    the_server.thread_cnt = 1;
+
+    pages_set_root(".");
+
+    if (ERR_OK != agoo_bind_port(&err, 3000)) {
+	printf("Failed to bind to port. %s\n", err.msg);
+	return err.code;
+    }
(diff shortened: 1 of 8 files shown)


c/agoo-c/simple.c
```diff
@@ -0,0 +1,50 @@
...
+}
+
+int
+main(int argc, char **argv) {
+    struct _Err	err = ERR_INIT;
+    
+    agoo_init("simple");
+
+    // Set the number of eval threads.
+    the_server.thread_cnt = 1;
```
OvermindDL1: Can't this be hardcoded to the number of cores on the system or so instead?

ohler55: With the light weight functions it doesn't make sense to set it to anything more than 1. If the functions were long running such as hitting a database then having more threads makes sense. You could set it to 100 if you wanted usually the developer tunes the thread count according to the application. It is not really related to the number of cores. As it is there are at least 5 threads running in Agoo-C by default when there are 1000 connections. At some point the `thread_cnt` option will be removed in favor of a self tuning thread creator. Its not there yet but if it was the count would be 1.

Long winded reply.

OvermindDL1: Long winded reply's are often good for the information they give though.  ^.^

So it does already use a thread pool for receiving network connections regardless?  That's all good then.  :-)

ohler55: Sort of. There is a single thread for accepting connections which get handed off to the read/write threads. If the read/write threads get too busy (exceed some managed connection limit) then a new read/write thread is created. There is a maximum number of threads imposed but that, for now is hard coded.

waghanza:
@ohler55 we can set thread num from the command line, if it's easy to do

c/agoo-c/simple.c
waghanza: it's kind of https://github.com/libuv/libuv then  ?

ohler55: Not nearly as expansive but could be doing some of the same things.

waghanza:
@ohler55 I've update results :tada: https://github.com/the-benchmarker/web-frameworks#full-table

ohler55:
Looks like I have some work to do to get the numbers up. This is helpful though. Maybe in a week I'll have some improvements. Thanks.

PR #664, opened 2018-11-28
Enable starlette / responder

waghanza:
Hi @tomchristie,

I have seen that `starlette` has been updated to **0.9.1**, then it might fix my issue.

I also open this `PR` as a **central** point to re-enable :
+ starlette
+ responder

`here `

I'm working on `python` **3.7** 

The following :

~~~
shards install
export FRAMEWORK=starlette
bin/neph ${FRAMEWORK} 
crystal spec
bin/benchmarker ${FRAMEWORK}
~~~

show that `starlette` is **OK** only when I comment lines https://github.com/the-benchmarker/web-frameworks/blob/master/tools/src/benchmarker.cr#L142-L152 wich lead to test only 1 endpoint (`/user` on `POST`)

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -119,14 +119,14 @@ python:
     language: "3.7"
     version: "1.4"
     website: falconframework.org
-  #starlette:
-  #  language: "3.7"
-  #  version: "0.8"
-  #  website: starlette.io
-  # responder:
-  #  language: "3.7"
-  #  version: "1.1"
-  #  github: kennethreitz/responder
+  starlette:
+    language: "3.7"
+    version: "0.9"
+    website: starlette.io
+  responder:
+    language: "3.7"
+    version: "1.1"
+    github: kennethreitz/responder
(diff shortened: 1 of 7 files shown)


waghanza:
@greed2411 seems then to be on our side

Following code (`python` translation from `crystal` code) is working :

~~~python
import subprocess

for i in ("/", "/user/0"):
  subprocess.run("bin/client --duration 15 --connections 1000 --threads 9 --url http://172.17.0.19:3000{0}".format(i), shell=True, check=True)

subprocess.run("bin/client --method POST --duration 15 --connections 1000 --threads 9 --url http://172.17.0.19:3000/user", shell=True, check=True)
~~~

greed2411:
Oh fine then. Thanks for the update!

waghanza:
@tomchristie I will re-enable `starlette` 

@kennethreitz I will re-enable `responder`

@greed2411 Probably a `gunicorn` / `uvicorn` update fix the issue I had

Issue #799, opened 2019-01-22
What's the version of crystal to run project locally?


AllanKlaus:
Whats the version of Crystal to run the project? I'm running 
```
Crystal 0.27.0 [c9d1eef8f] (2018-11-01)

LLVM: 4.0.0
Default target: x86_64
```

And I'm getting this error on every test

```
Unhandled exception: Invalid Float64: --: 1: --: wrk: not found
 (ArgumentError)
  from /usr/share/crystal/src/string.cr:665:56 in 'to_f64'
  from /usr/share/crystal/src/string.cr:629:5 in 'to_f'
  from /usr/share/crystal/src/string.cr:628:3 in 'to_f'
  from tools/src/client.cr:55:49 in 'run'
  from tools/src/client.cr:97:1 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:86:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???
Unhandled exception: Unexpected token: EOF at 1:1 (JSON::ParseException)
  from /usr/share/crystal/src/json/pull_parser.cr:518:5 in 'parse_exception'
  from /usr/share/crystal/src/json/pull_parser.cr:514:5 in 'unexpected_token'
  from /usr/share/crystal/src/json/pull_parser.cr:50:7 in 'initialize'
  from /usr/share/crystal/src/json/pull_parser.cr:12:3 in 'new'
  from /usr/share/crystal/src/json/from_json.cr:12:3 in 'from_json'
  from tools/src/benchmarker.cr:144:5 in 'benchmark'
  from tools/src/benchmarker.cr:213:12 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:97:5 in 'main_user_code'
  from /usr/share/crystal/src/crystal/main.cr:86:7 in 'main'
  from /usr/share/crystal/src/crystal/main.cr:106:3 in 'main'
  from __libc_start_main
  from _start
  from ???

**Comment by Blacksmoke16** [comment]
I'm pretty sure this is due to not having `wrk` installed locally.  That should probably be added to the documentation.

I also been having some trouble running this locally:
```
Benchmark running ...
--: 1: --: Syntax error: word unexpected (expecting ")")
```

[...]

PR #800, opened 2019-01-22
Update readme

AllanKlaus:
- Motivation: Was missing the dependency of `wrk` in the README file.

-Issue: #799


waghanza:
good catch, I'll regenerate readme (it is in fact a template) soon

waghanza:
Thanks again for your contribution :heart:

PR #896, opened 2019-02-06
Update spider-gazelle to support the latest version of action-controller

stakach:


waghanza:
@stakach I'm updating results, I'll push them directly in this **branch**

Issue #950, opened 2019-02-18
Introduce fastapi


waghanza:
+ Language : python
+ Repo : https://github.com/tiangolo/fastapi
+ Author : @tiangolo

PR #958, opened 2019-02-19
Introduce fastapi

waghanza:
Hi,

This `PR` addresses #950 

@tiangolo How can I produce `text/plain` instead of `application/json` ?

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,6 +43,7 @@ env:
     - FRAMEWORK=responder
     - FRAMEWORK=hug
     - FRAMEWORK=bocadillo
+    - FRAMEWORK=fastapi
     - FRAMEWORK=jester
     - FRAMEWORK=act
     - FRAMEWORK=rails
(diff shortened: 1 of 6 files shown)


waghanza:
@tiangolo found it :stuck_out_tongue:

tiangolo:
Great! :smile:

waghanza:
@tomchristie When using a `starlette` **Response**, `status_code` is `200` by default, even for `POST` method.

Should not `status_code` depends on HTTP method in use ? 201 for `POST`,  204 for `PUT` ...

tiangolo:
`201` is specifically for "Created" responses.

But not all `POST` operations necessarily terminate on a resource "created".

`200` in `PUT` or `POST`:

> The resource describing the result of the action is transmitted in the message body.

Taken from: https://developer.mozilla.org/en-US/docs/Web/HTTP/Status#Successful_responses

waghanza:
indeed this code are not very standard

I always find this odd having a  `200` for any `POST` requests

tiangolo:
Yep, and now with GraphQL that everything is on `POST` :joy:

BTW, thanks for the invitation to the org! :tada:

PR #989, opened 2019-02-26
Upgrade to kore 3.2

waghanza:
Hi @jorisvink,

We are using `kore` in this benchmarking project.

Could you quick check our implementation in https://github.com/the-benchmarker/web-frameworks/blob/master/c/kore/hello/src/hello.c

Regards,

### DIFF
--- a/c/kore/Dockerfile
+++ b/c/kore/Dockerfile
@@ -1,7 +1,7 @@
 FROM gcc:8.3.0
 
-RUN wget -c https://kore.io/releases/kore-3.1.0.tar.gz && tar xvf kore-3.1.0.tar.gz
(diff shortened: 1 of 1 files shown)


jorisvink:
Hi,

Neat. I was secretly looking forward to seeing this updated to latest 3.2 release.

So the user_details() seems incredibly complex for no reason. If you need to get the id specified after the /user path why not just use req->path (which is guaranteed to match /user/<id> due to the routing) + sizeof("/user") - 1.

I don't get it.

Is it suppose to just return the id part after /user?

jorisvink:
Also your diff didn't touch the tarball it is extracting.

waghanza:
@jorisvink Thanks for your feedback.  

It will be awesome if you (or anyone) provide a `PR`.

I have no `C` skillse

PR #1001, opened 2019-02-27
upgrade perfect

waghanza:
Hi,

This `PR` upgrade `Perfect` from `3.0` to `3.1`

@iamjono Would you mind if I ping you on each `Perfect` **updates** on our side ?

You could help us with `swift` :heart:

### DIFF
--- a/swift/perfect/Package.swift
+++ b/swift/perfect/Package.swift
@@ -4,9 +4,9 @@ import PackageDescription
 let package = Package(
     name: "server",
     dependencies: [
-        .package(url: "https://github.com/PerfectlySoft/Perfect-HTTPServer.git", "3.1.0"..<"3.2.0")
+        .package(url: "https://github.com/PerfectlySoft/Perfect", "3.1.0"..<"3.2.0")
     ],
     targets: [
-        .target(name: "server", dependencies: ["PerfectHTTPServer"], path: ".")
+        .target(name: "server", dependencies: ["Perfect"], path: ".")
     ]
 )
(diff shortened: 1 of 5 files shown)


waghanza:
cc @kiliankoe

swift/perfect/Package.swift
```diff
@@ -4,9 +4,9 @@ import PackageDescription
 let package = Package(
     name: "server",
     dependencies: [
-        .package(url: "https://github.com/PerfectlySoft/Perfect-HTTPServer.git", "3.0.0"..<"3.1.0")
+        .package(url: "https://github.com/PerfectlySoft/Perfect", "3.1.0"..<"3.2.0")
```
kiliankoe:
```suggestion
        .package(url: "https://github.com/PerfectlySoft/Perfect", .upToNextMinor(from: "3.1.0"))
```

The more idiomatic way of specifying version ranges for SwiftPM is using the available requirement type. Both work fine though 😊

swift/perfect/Package.swift
```diff
@@ -4,9 +4,9 @@ import PackageDescription
...
     ],
     targets: [
-        .target(name: "server", dependencies: ["PerfectHTTPServer"], path: ".")
+        .target(name: "server", dependencies: ["Perfect"], path: ".")
```
kiliankoe:
```suggestion
        .target(name: "server", dependencies: ["PerfectLib"], path: ".")
```

As far as I can tell Perfect calls their library package `PerfectLib` instead of `Perfect`. See https://github.com/PerfectlySoft/Perfect/blob/master/Package.swift#L25

PR #988, opened 2019-02-26
Intoduce kami

waghanza:
Hi @guregu,

This `PR` addresses #663

I've added `kami` in this benchmark.

Could you check our implementation ?

Results found (docker on 4 x CPU / 8gb)

| Language (Runtime) | Framework (Middleware) | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
|---------------------------|---------------------------|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|
| go (1.11) | [fasthttprouter](http://godoc.org/github.com/buaazp/fasthttprouter) (0.1) | 5.55 ms | 5.05 ms | 8.53 ms | 14.78 ms | 142.99 ms | 2863.00 | 
| go (1.11) | [chi](http://github.com/go-chi/chi) (4.0) | 13.00 ms | 9.39 ms | 21.24 ms | 118.09 ms | 427.88 ms | 21244.67 | 
| go (1.11) | [muxie](http://godoc.org/github.com/kataras/muxie) (1.0) | 11.41 ms | 9.59 ms | 15.76 ms | 73.15 ms | 335.53 ms | 14806.33 | 
| go (1.11) | [iris](http://iris-go.com) (11.1) | 12.04 ms | 9.93 ms | 16.93 ms | 68.31 ms | 523.91 ms | 19680.00 | 
| go (1.11) | [gin](http://gin-gonic.github.io/gin) (1.3) | 12.55 ms | 10.19 ms | 18.48 ms | 83.77 ms | 388.75 ms | 16719.67 | 
| go (1.11) | [echo](http://echo.labstack.com) (4.0) | 14.16 ms | 10.32 ms | 21.00 ms | 129.89 ms | 482.05 ms | 24880.33 | 
| go (1.11) | [kami](http://github.com/guregu/kami) (2.2) | 15.06 ms | 10.73 ms | 24.93 ms | 139.06 ms | 403.63 ms | 24584.00 | 
| go (1.11) | [beego](http://beego.me) (1.11) | 13.87 ms | 10.80 ms | 22.94 ms | 106.79 ms | 357.95 ms | 19435.00 | 
| go (1.11) | [gorilla-mux](http://www.gorillatoolkit.org/pkg/mux) (1.7) | 16.00 ms | 11.04 ms | 30.21 ms | 132.70 ms | 380.82 ms | 24787.33 | 
[...]

PR #1006, opened 2019-02-28
Fix go resolver issue

waghanza:
Hi,

This `PR` addresses #998 

Regards,

### DIFF
--- a/go/beego/Dockerfile
+++ b/go/beego/Dockerfile
@@ -4,6 +4,7 @@ WORKDIR /go/src/app
 COPY main.go go.mod ./
 
 ENV GO111MODULE=on
+RUN go get
 RUN go build .
 
 EXPOSE 3000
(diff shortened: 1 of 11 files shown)


hmarr:
I'd actually recommend running a `go get` in this directory - when I did so locally a couple of other things were added to the `go.mod`.

guregu:
I agree with @hmarr, it should really look this the comment I posted in the PR here: https://github.com/the-benchmarker/web-frameworks/pull/988#issuecomment-467777091 (which I just copy/pasted from what `go get` gave me).
I know it feels weird specifying the indirect dependencies but that's just how Go modules work, especially with projects that predate modules and are incompatible with it.

waghanza:
@hmarr @guregu As I understand, `go get` COULD automatically check dependency tree, as `bundle install` does in `ruby` ?

I don't know how `go` works for this, but if this is a **resolver**, I think it should not change file in `git` **repo**, but should be done just before build
+ on @dependabot[bot] side
+ in our local containers

waghanza:
@guregu the behavior described is not so weird, at least it's how ruby / bundler works :tada:

Issue #1016, opened 2019-03-03
Introduce zewo


waghanza:
+ Language : swift
+ Repo : https://github.com/Zewo/Zewo
+ Author : @paulofaria

PR #1017, opened 2019-03-03
Introduce zewo

waghanza:
Hi @paulofaria,

I'm trying to add https://github.com/Zewo/Zewo here.

Could you quick review the code

This addresses #1016 

Regards,

### DIFF
--- a/swift/zewo/Package.swift
+++ b/swift/zewo/Package.swift
@@ -1,0 +1,12 @@
+// swift-tools-version:4.0
+import PackageDescription
+
+let package = Package(
+    name: "server",
+    dependencies: [
+        .package(url: "https://github.com/Zewo/HTTPServer.git", .upToNextMinor(from: "0.14.0"))
+    ],
+    targets: [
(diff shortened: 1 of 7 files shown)


paulofaria:
This is all good. The only caveat is to launch as many processes as the number of logical cores in the test machine. Use the command below to get the number of logical processors:

```
nproc --all
```

Then just launch as many processes as the number the command above returns. For example, if it returns 8, launch 8 processes of the executable generated by SwiftPM.

waghanza:
Thanks @paulofaria 

An other question if I may, can I run `swiftlint` is `CLI` (to clean the codebase)

swift/zewo/Package.swift
```diff
@@ -0,0 +1,12 @@
+// swift-tools-version:4.0
+import PackageDescription
+
+let package = Package(
+    name: "server",
+    dependencies: [
+        .package(url: "https://github.com/Zewo/HTTPServer.git", .upToNextMinor(from: "0.14.0"))
```
waghanza: @paulofaria this line is not working, any idea ?

PR #1066, opened 2019-03-10
Using restana 2.10.0

jkyberneees:
Using new `restana` configuration option that allows to disable the `response` event if unused.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -100,7 +100,7 @@ env:
     - FRAMEWORK=restana
     - FRAMEWORK=hapi
     - FRAMEWORK=fastify
-    # - FRAMEWORK=foxify
+    - FRAMEWORK=foxify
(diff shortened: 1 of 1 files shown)


jkyberneees:
Hi @waghanza, I wanted to let you know that some how this merge was override during the merging. 
Changes were actually discarded. In this case it was more than updating to latest version, but also to disable an unused internal event to reach even more performance. 
More details here:
https://github.com/jkyberneees/ana/blob/master/specs/disable-response-event.test.js#L8

How can we proceed here?
Thanks in advance.

waghanza:
Hi @jkyberneees,

I think I've understood. This has been override because of `squash` (I think).

Could you redo a `PR`, I'll be careful doing this run.

Thanks for notification :tada: 

Regards,

PR #1147, opened 2019-03-31
Introduce violtear

waghanza:
Hi @nbari,

I've added [violetear](https://violetear.org) in here.

Could you review the code base ?

I've updated results in https://github.com/waghanza/http-benchmark/blob/add_violetear/README.md#results, however their are still on **development**

Regards,

### DIFF
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -249,7 +249,7 @@ go:
     language: "1.12"
   violetear:
     website: violetear.org
-    version: 6.0
+    version: 7.0
     language: 1.12
 swift:
   kitura:
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -249,7 +249,7 @@ go:
     language: "1.12"
   violetear:
     website: violetear.org
-    version: 5.0
+    version: 6.0
     language: 1.12
 swift:
   kitura:
--- a/FRAMEWORKS.yml
+++ b/FRAMEWORKS.yml
@@ -247,6 +247,10 @@ go:
     github: "guregu/kami"
     version: "2.2"
     language: "1.12"
+  violetear:
+    website: violetear.org
+    version: 5.0
+    language: 1.12
 swift:
   kitura:
     website: kitura.io
--- a/go/violetear/main.go
+++ b/go/violetear/main.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"fmt"
+	"log"
+	"net/http"
+	"time"
+
+	"github.com/nbari/violetear"
+)
+
+func showId(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, violetear.GetParam("id", r))
+}
+
+func showEmpty(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "")
+}
+
+func main() {
+	router := violetear.New()
+
+	router.AddRegex(":id", `\d`)
+
+	router.HandleFunc("/", showEmpty, "GET")
+	router.HandleFunc("/user/:id", showId, "GET")
+	router.HandleFunc("/user", showEmpty, "POST")
+
+	srv := &http.Server{
+		Addr:           ":3000",
+		Handler:        router,
+		ReadTimeout:    5 * time.Second,
+		WriteTimeout:   7 * time.Second,
+		MaxHeaderBytes: 1 << 20,
+	}
+	log.Fatal(srv.ListenAndServe())
+
+}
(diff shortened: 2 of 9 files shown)


waghanza:
Is the version listed right ? `go get` specify **v0.0.0** from the tag **5.0.1**

nbari:
Hi  @waghanza thanks, the latest version is 5.0.1

go/violetear/main.go
```diff
@@ -0,0 +1,38 @@
...
+}
+
+func showEmpty(w http.ResponseWriter, r *http.Request) {
+	fmt.Fprintf(w, "")
+}
+
+func main() {
+	router := violetear.New()
+
+	router.AddRegex(":id", `\d`)
```
nbari: replace `\d` with `[0-9]+` to match any number

FRAMEWORKS.yml
```diff
@@ -247,6 +247,10 @@ go:
     github: "guregu/kami"
     version: "2.2"
     language: "1.12"
+  violetear:
+    website: violetear.org
+    version: 5.0
```
nbari: please update version to `6.0`

waghanza:
Hi @nbari,

Sure, I'll update. I though that for the regexp `\d` and `[0-9]` was the same
https://en.wikipedia.org/wiki/Regular_expression#Character_classes

Btw, you're right I forget the `+`.

nbari:
@waghanza I think is safe to roll back and force, fix the regex to be `^<your regex>$` , dealing with regex is tricky and they could be matching othere handlers, for example if just using `\d` it could match:

    /users/foo123bar

When probably the desired behaviour was to match only:

    /users/123

The user could fix the regex and use `^\d+$` but that could take a while for him to find the issue.

what do you think, now that where doing a basic test?

waghanza:
I'll then at a spec that test this case

waghanza:
With a strict regexp `router.AddRegex(":id", `^[0-9]+$`)`, the status code given is `525`.

Should not it be 404 ?

nbari:
Yes,  please try version  `7.0.0`,

If you use:
    
    router.AddRegex(":id", [0-9]+)
    // or
    router.AddRegex(":id", \d+)

    router.HandleFunc("/user/:id", showId, "GET")

If will return `404` for `/users/foo123bar`

and `200` for `/users/123`

go/violetear/main.go
```diff
@@ -20,7 +20,7 @@ func showEmpty(w http.ResponseWriter, r *http.Request) {
 func main() {
 	router := violetear.New()
 
-	router.AddRegex(":id", `[0-9]+`)
+	router.AddRegex(":id", `^[0-9]+$`)
```
nbari: You could indeed just use `\d+`

waghanza:
@sebastienros How can we for `id` to be a numerical `id` in 
https://github.com/the-benchmarker/web-frameworks/blob/master/csharp/aspnetcore/Startup.cs#L23

nbari:
@waghanza thanks for testing and your time on this 👍, just for the record using regex is quite a heavy task, indeed are mainly supported for extream cases like the one briefly described here: https://violetear.org/post/context-named/ (Duplicated named parameters):

    uuid := `[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}`
    router.AddRegex(":uuid", uuid)
    router.HandleFunc("/test/:uuid/:uuid", handleUUID, "GET,HEAD")

A catch-all alocates less, for example:

     router.HandleFunc("/test/*", handler)

And to get the param:

    func showId(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, violetear.GetParam("*", r))
    }

Full example:

```go
package main

import (
	"fmt"
	"log"
	"net/http"
	"time"

	"github.com/nbari/violetear"
)

func showId(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, violetear.GetParam("*", r))
}

func showEmpty(w http.ResponseWriter, r *http.Request) {
	fmt.Fprintf(w, "")
}

func main() {
	router := violetear.New()
	router.LogRequests = true

	router.AddRegex(":id", `\d+`)

	router.HandleFunc("/", showEmpty, "HEAD, GET")
	router.HandleFunc("/user/*", showId, "HEAD, GET")
	router.HandleFunc("/user", showEmpty, "POST")

	srv := &http.Server{
		Addr:           ":3000",
		Handler:        router,
		ReadTimeout:    5 * time.Second,
		WriteTimeout:   7 * time.Second,
		MaxHeaderBytes: 1 << 20,
	}
	log.Fatal(srv.ListenAndServe())
}
```

waghanza:
@mfirry @iPomme how can we force `id` to be numerical in
https://github.com/the-benchmarker/web-frameworks/blob/master/scala/akkahttp/src/main/scala/Main.scala#L22

mfirry:
hey @waghanza do you still need help with this?

PR #1182, opened 2019-04-06
Freeze Roda app

adam12:


waghanza:
@adam12 What does `freeze` do `here` ?

adam12:
From the Roda README:

> Freezing the app makes modifying app-level settings raise an error, alerting you to possible thread-safety issues in your application. It is recommended to freeze the app in production and during testing. The .app is an optimization, which saves a few method calls for every request.

PR #1253, opened 2019-04-29
[WIP] [Python] Enforce unicorn for ASGI frameworks

ahopkins:
The existing implementation here for [Sanic](https://sanicframework.org/) uses `gunicorn` instead of its own internal server. This PR updates the benchmark to run against Sanic's own server thus **greatly** improving performance.

### DIFF
--- a/python/sanic/Dockerfile
+++ b/python/sanic/Dockerfile
@@ -9,4 +9,4 @@ COPY requirements.txt server.py ./
 
 EXPOSE 3000
 
-CMD gunicorn --log-level warning --bind 0.0.0.0:3000 --reuse-port --workers $(nproc) --worker-class sanic.worker.GunicornWorker server:app
+CMD python server.py $(nproc)
--- a/python/sanic/server.py
+++ b/python/sanic/server.py
@@ -1,19 +1,31 @@
 from sanic import Sanic
 from sanic.response import text
+import sys
 
 app = Sanic(log_config=None)
+WORKERS = int(sys.argv[-1])
 
 
 @app.route("/")
 async def index(request):
-    return text('')
+    return text("")
 
 
-@app.route("/user/<id:int>", methods=['GET'])
+@app.route("/user/<id:int>", methods=["GET"])
 async def user_info(request, id):
     return text(str(id))
 
 
-@app.route("/user", methods=['POST'])
+@app.route("/user", methods=["POST"])
 async def user(request):
-    return text('')
+    return text("")
+
+
+if __name__ == "__main__":
+    app.run(
+        debug=False,
+        access_log=False,
+        workers=WORKERS,
+        port=3000,
+        host="0.0.0.0",
+    )


waghanza:
Thanks @ahopkins, I understand that changing server could increase performance a lot.

I have already changed the server behind this implementation, this is not about performance here,  but make all python implementations uniform here.

I prefer keeping`gunicorn` since **all** (except `japronto`) use it.

If there is a worker for `sanic`, I'll be glad to upgrade the actual implementation.

The purpose is to homogenize implementations to ensure a representative benchmark.

Regards,

ahopkins:
I think the problem is that `gunicorn` support does not exactly have first-class citizen status in Sanic. The difference in performance (on my machine) is not even close.

**Using `gunicorn`**
```
    Last update: 2019-04-30
    ```
    OS: Linux (version: 5.0.4-arch1-1-ARCH, arch: x86_64)
    CPU Cores: 12
    threads: 13, connections: 1000
    ```
    Benchmark running ...
    Done. <- sanic

    ### Latency


    #### Ranking (top 5)


    :one: sanic (python)


    #### Full table

    | Language (Runtime) | Framework (Middleware) | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
    | python (3.7) | [sanic](http://github.com/huge-success/sanic) (19.3) | 52.45 ms | 44.11 ms | 97.69 ms | 163.76 ms | 305.50 ms | 33171.00 | 

    ### Requests per seconds


    #### Ranking (top 5)


    :one: (sanic) (python)


    #### Full table

    | Language (Runtime) | Framework (Middleware) | Requests / s | Throughput |
    |---------------------------|---------------------------|----------------:|---------:|
    | python (3.7) | [sanic](http://github.com/huge-success/sanic) (19.3) | 24544.33 | 46.19 MB |
```

**Using Sanic.run**
```
    Last update: 2019-04-30
    ```
    OS: Linux (version: 5.0.4-arch1-1-ARCH, arch: x86_64)
    CPU Cores: 12
    threads: 13, connections: 1000
    ```
    Benchmark running ...
    Done. <- sanic

    ### Latency


    #### Ranking (top 5)


    :one: sanic (python)


    #### Full table

    | Language (Runtime) | Framework (Middleware) | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
    | python (3.7) | [sanic](http://github.com/huge-success/sanic) (19.3) | 30.23 ms | 26.00 ms | 53.04 ms | 90.53 ms | 169.41 ms | 17376.33 | 

    ### Requests per seconds


    #### Ranking (top 5)


    :one: (sanic) (python)


    #### Full table

    | Language (Runtime) | Framework (Middleware) | Requests / s | Throughput |
    |---------------------------|---------------------------|----------------:|---------:|
    | python (3.7) | [sanic](http://github.com/huge-success/sanic) (19.3) | 67305.33 | 123.61 MB |

```
The standard way to deploy Sanic is not using `gunicorn`, rather the worker is there for the benefit of the developer to transition their application. I do understand the attempt to homogenize the results, but it just seems like these results are not representative of the expected performance. In fact, we are working on lowering the emphasis on `gunicorn` even more in favor of ASGI going forward. While I cannot say that the `GunicornWorker` (for compatibility purposes), it certainly is being pushed aside. And, unless there is some specific feature that the `gunicorn` server provides that Sanic.run does not, the default deployment strategy is **not** to use it.

A more fair comparison of apples to apples would be to use `Sanic.run` similar to your tests for `japronto` and `vibora` that also come with their own servers. This allows the framework to operate closer to its intent and provide a more accurate picture.

waghanza:
Hi @ahopkins, 

The actual benchmark way use `docker`, the results are not very accurate, so we can not say anything about them until I cloudify them (use one remote isolated container/VM/VPS instead of local ones).

Btw, the results are misleading by around 40%, but I'm sure **sanic** `built-in` server has more performance that `gunicorn`.

Using `gunicorn` is not gravel is stone 😗 and if you have a better way to run apple to apple comparison for each python framework here, be my guest

ahopkins:
I apologize if you know all this already, but just so that we know we are speaking the same language ...

Frameworks in Python generally fall into one of two buckets: sync and async.

The first category includes the heavy hitters like `django`, `flask`, and `pyramid`. When the language added native `async`/`await` keywords, a whole new group of frameworks popped up: `Sanic`, `aiohttp`, `quart`, etc.

The sync frameworks do not come with a built-in webserver, but are merely WSGI compliant. Something like `gunicorn` is required to run them.

But, that leaves the question of how to run `async` frameworks since they generally are **NOT** WSGI compliant. WSGI does not play nice with async. A request comes in, it pauses until a response goes out. 

There are three approaches to this problem:

1. Run your own server (`japronto`, `vibora`, `sanic`)
1. Port the application to be compatible with WSGI and give up on `async` (`aiohttp`, `sanic`, etc)
1. Run ASGI and make the framework compatible with that (`quart`, `starlette`, etc)

Using something like `gunicorn` to run an async framework can work. But it does not actually provide a representative benchmark for that framework. I propose that **all** of the `async` python frameworks be run in a manner that allows them to leverage `async`/`await`, which means not using `gunicorn`. 

For those projects where `gunicorn` is standard practice, then by all means, that makes sense. But it is not for the newer `async` projects. I would suggest (where possible) provide preferential treatment in the following order:

1. If ASGI compliant: `uvicorn`, `hypercorn`, or `daphne` (pick one and stick with it)
1. built in server
1. `gunicorn`

All of the `async` frameworks will fall into either 1 or 2. Option 3 is really a last resort.

waghanza:
Thanks for this clear explanation. We are using `uvicorn` to run **async** like `quart`

If you find a way to run `uvicorn` for **all** **async** frameworks here, it will be ok for me.

I personally prefer `uvicorn` since it could be used alongside `gunicorn`

Btw, any ASGI solution is fine for me

waghanza:
Hi @ahopkins, time for this `PR` ?

ahopkins:
Yes. I will probably do it this coming week in the next several days.

PR #1267, opened 2019-05-02
Add ktor 1.1.5

norac89:

   + [x] `travis.yml`, for CI
+ [x] A `Dockerfile` exists ?
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~

Adds [Ktor](https://ktor.io/)


waghanza:
Awesome :tada: the first framework in `kotlin`

A pleasure to **review**

FRAMEWORKS.yml
```diff
@@ -323,3 +323,8 @@ c:
     website: github.com/ohler55/agoo-c
     version: "0.4"
     language: "11"
+kotlin:
+  ktor:
+    website: ktor.io
+    version: "1.1.5"
```
waghanza: Could you use `X.Y` format here ? No needs to be **string**

FRAMEWORKS.yml
```diff
@@ -323,3 +323,8 @@ c:
...
+    language: "1.3.31"
```
waghanza: Could you use `X.Y` format here ? No needs to be **string**

kotlin/ktor/Dockerfile
```diff
@@ -0,0 +1,22 @@
+FROM gradle:4.10-jdk8
```
waghanza: Why this version ? The latest available is not the `5` ?

kotlin/ktor/resources/logback.xml
```diff
@@ -0,0 +1,12 @@
+<configuration>
+    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
+        <encoder>
+            <pattern>%d{YYYY-MM-dd HH:mm:ss.SSS} [%thread] %-5level %logger{36} - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="trace">
```
waghanza: Could it be disabled ?
All frameworks has lot disabled here (or at least **SHOULD**)

kotlin/ktor/Dockerfile
norac89: Had problem in a more complex previous project updating to gradle 5. So just out of lazyness.
Made the change here and everything went smoothly.

waghanza:
@norac89 I can see that https://hub.docker.com/_/gradle use `jdk` **8** and https://hub.docker.com/r/zenika/kotlin `jdk` **12**

Any reason to choose the first ? 

It seems that `kotlin` **COULD** target `jdk` **12**

* https://github.com/keeganwitt/docker-gradle/issues/91
* https://github.com/docker-library/openjdk/issues/302

norac89:
From https://blog.jetbrains.com/kotlin/2019/04/kotlin-1-3-30-released/

> If you run the code under JVM versions 9, 10, 11, or 12, you can now set the corresponding jvmTarget. This setting will affect the version of the generated classfiles, and the resulting bytecode won’t run on any lower version of JVM. Note that so far newer versions don’t add any bytecode optimizations or features beyond the ones that exist in lower versions, but that is going to change in the future.

I think that the code is compile to JVM 8 bytecode by default. Haven't tested targeting other JVM version yet. But it seems there would be no gain doing so for now.

I must say I am no Kotlin expert. Came from python and started playing with it recently. So I may be wrong...

Anyway I have updated the docker image to jdk11 (the latest version from the official gradle images)

kotlin/ktor/resources/application.conf
```diff
@@ -0,0 +1,9 @@
+ktor {
+    deployment {
+        port = 3000
+        port = ${?PORT}
```
waghanza: Not a duplicate ?

FRAMEWORKS.yml
```diff
@@ -323,3 +323,8 @@ c:
     website: github.com/ohler55/agoo-c
     version: "0.4"
     language: "11"
+kotlin:
+  ktor:
+    website: ktor.io
+    version: 1.1.5
```
waghanza: Could you use only `X.Y` ? I mean `1.1`, this is more readable for audience

FRAMEWORKS.yml
```diff
@@ -323,3 +323,8 @@ c:
...
+    language: 1.3.31
```
waghanza: Same as above

kotlin/ktor/src/Application.kt
```diff
@@ -0,0 +1,26 @@
...
+
+fun main(args: Array<String>): Unit = io.ktor.server.netty.EngineMain.main(args)
+
+@Suppress("unused") // Referenced in application.conf
+fun Application.module() {
+    routing {
+        get("/") {
+            call.respondText("", contentType = ContentType.Text.Plain, status = HttpStatusCode.OK)
+        }
+        get("/user/{id}") {
```
waghanza: Could this route be constrains ?
I mean can we for `id` to be numeric ?

waghanza:
No skills in `kotlin` either -> will have to test ;-)

waghanza:
Here is a table of perf recorded on my 4 CPU x 8 Gb (what seems to say that performance decrease for `jdk` **11**).

Let's wait for https://github.com/keeganwitt/docker-gradle/pull/98 to be release ;-) to add `jdk` **12** here

|  | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation | Req / s | Throughput |
|----------------------------------------|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|-----------:|----------------:| 
| jdk8 | 26.48 ms | 21.60 ms | 46.48 ms | 89.58 ms | 331.00 ms | 18494.00 | 38843.67 | 60.59 MB |
| jdk11 | 38.63 ms | 28.73 ms | 70.78 ms | 183.35 ms | 709.34 ms | 39578.67 | 28423.00 | 44.30 MB |

norac89:
Interesting results. Did not think it would be that slower on jdk 11. 
But kotlin code is still compiled to JVM 8 bytecode even if the jdk 11 is used. There is a `jvmTarget` setting that allows kotlin to be compile to specific JVM version. But haven't play with it yet. And I don't think I will have the time, at least in the next weeks.

I think it would be better to keep everything running on a jdk 8 image which correspond to the generated bytecode version. Also, documentation on ktor is using jdk 8.

kotlin/ktor/resources/application.conf
norac89: It allows to override the port if a environment variable PORT is present.

kotlin/ktor/src/Application.kt
norac89: 3817050

FRAMEWORKS.yml
norac89: 9faf6c1

FRAMEWORKS.yml
norac89: 9faf6c1

kotlin/ktor/resources/application.conf
waghanza: ok

waghanza:
@norac89 I've updated results with `jdk` **11**, will consider to use `jdk` **8**

waghanza:
@fracaron do you have any recommendation for `kotlin` lint ? I'm try to use community standard for all frameworks here in https://github.com/the-benchmarker/web-frameworks/pull/1011

fracaron:
I haven't used any lint tools yet except for what is included in Intellij IDE. 
With a quick search, I found https://ktlint.github.io/.

waghanza:
Yeah that's the one I've used in https://github.com/waghanza/http-benchmark/blob/community_standard/.ci/lint.sh#L170

Issue #1279, opened 2019-05-09
Swift 5.0 Kitura-NIO  


shial4:
You already have a benchmark for Kitura. However, there is an option to lunch kitura under Swift-NIO. It is a totally different implementation. I am curious how would results look like under this environment. 

With Kitura, to run on top of Kitura-NIO (instead of Kitura-Net) all you need to do is set an environment variable called KITURA_NIO before building your Kitura application.

+ Language : Swift 5.0
+ Repo : https://github.com/IBM-Swift/Kitura-NIO
+ Author : IBM-Swift

waghanza:
Interesting, two variants of `kitura`. What does this `swift-nio` offer ?

shial4:
The core comprising of HTTPServer, ClientRequest/ClientResponse and TLS support have been implemented using SwiftNIO. Kitura-NIO uses NIOSSL for TLS support.

SwiftNIO is a cross-platform asynchronous event-driven network application framework for rapid development of maintainable high-performance protocol servers & clients.

For example swift vapor (3+) is built on top of it as well. That's why I want to see those numbers. How do they compare

waghanza:
Hi @shial4,

Thanks for clarification. 

As I understand `Kitura-NIO` is therefore an `async` variant of `Kitura` 

Feel free to add a   `PR` :heart:

PR #1320, opened 2019-05-15
[Swift] Add Kitura-NIO

shial4:

   + [x] `travis.yml`, for CI
+ [x] A `Dockerfile` exists ?
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~

### DIFF
--- a/swift/kitura-nio/main.swift
+++ b/swift/kitura-nio/main.swift
@@ -1,0 +1,22 @@
+import Kitura
+
+let router = Router()
+
+router.get("/") { _, res, next in
+    try res.send("")
+    next()
+}
+
+router.get("/user/:id") {  req, res, next in
+  let id = req.parameters["id"] ?? ""
+  try res.send(id).end()
+  next()
+}
+
+router.post("/user") { _, res, next in
+  try res.send("")
+  next()
+}
+
+Kitura.addHTTPServer(onPort: 3000, with: router)
+Kitura.run()
--- a/swift/kitura-nio/main.swift
+++ b/swift/kitura-nio/main.swift
@@ -7,10 +7,14 @@ router.get("/") { _, res, next in
     next()
 }
 
-router.get("/user/:id") {  req, res, next in
-  let id = req.parameters["id"] ?? ""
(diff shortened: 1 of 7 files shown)


swift/kitura-nio/main.swift
```diff
@@ -0,0 +1,22 @@
+import Kitura
+
+let router = Router()
+
+router.get("/") { _, res, next in
+    try res.send("")
+    next()
+}
+
+router.get("/user/:id") {  req, res, next in
```
waghanza: can this be constrained as digit ?

waghanza:
@shial4 can you update also `kitura` ?

shial4:
Updated @waghanza  :) 
Can't wait to compare the results!

swift/kitura-nio/main.swift
shial4: Not sure If I do understand, what exactly do you want to achieve?

shial4: We could use codable route. For instance:
```
//contains an identifier of type `Int`
func getUserByIdHandler(id: Int, completion: (Int, RequestError?) -> Void) {
    completion(id, nil)
}

router.get("/user/:id", handler: app.getUserByIdHandler)
```

Or we could use raw request and manually convert it to Int.

waghanza: Up to you :stuck_out_tongue: 

I honestly prefer using what is considered as a `best-practice` in `swift (clean / re-usable code)

shial4: Either are clean. Codable is model convertible language stuff. I do like raw requests style.

waghanza:
Close #1279

swift/kitura-nio/main.swift
waghanza: Could you homogenize also for `kitura` ?

I will update result **tomorrow**

shial4:
@waghanza  Specifies that only digits should be matched for user/:id route.
In both Kitura & Kitura-NIO example.

PR #1207, opened 2019-04-15
Add numercial constraint

waghanza:
Hi,

This `PR` force route **#2** (`/user/<id>`) to be constraint.

`<id>` must be an integer

Here is an update to date list of working implementations :

+ C
   + [ ] Agoo-C
   + [ ] Kore
+ C++
   + [ ] EvHTP
+ C#
   + [x] ASP.NET Core
+ Crystal
   + [ ] Amber
   + [X] Athena
   + [x] Kemal
   + [x] Lucky
   + [x] Onyx
   + [ ] Orion
   + [ ] Raze
   + [ ] Router.cr
   + [ ] Spider-Gazelle
+ Go
   + [x] Beego
   + [x] Chi
   + [ ] Echo
   + [ ] FastHTTPRouter
   + [ ] GF
   + [ ] Gin
   + [x] Gorilla/Mux
   + [x] Gorouter
   + [ ] Kami
   + [x] Violetear
+ Java
  + [ ] Act
  + [x] Rapidoid
  + [x] Spring boot
+ Kotlin
   + [x] Ktor
+ Nim
   + [ ] Jester
+ Javascript
   + [x] 0http
   + [x] express
   + [x] fastify
   + [x] foxify
   + [x] hapi
   + [x] koa
   + [x] muneem
   + [x] polka
   + [ ] rayo
   + [x] restana
   + [x] restify
   + [ ] turbo_polka
+ PHP
   + [x] Laravel
   + [x] Lumen
   + [x] Slim
   + [x] Symfony
   + [x] Zend-expressive
   + [ ] Zend-framework
+ Python
   + [x] AioHTTP
   + [x] Bocadillo
   + [x] Bottle
   + [x] Cyclone
   + [x] Django
   + [x] Falcon
   + [x] Fastapi
   + [x] Flask
   + [x] Hug
   + [ ] Japronto
   + [x] Masonite
   + [x] Molten
   + [x] Quart
   + [x] Responder
   + [x] Sanic
   + [x] Starlette
   + [x] Tornado
   + [x] Vibora
+ Ruby
   + [x] Agoo
   + [ ] Cuba
   + [x] Flame
   + [x] Hanami
   + [x] Rack-routing
   + [x] Rails
   + [x] Roda
   + [x] Sinatra
+ Rust
   + [ ] Actix-web
   + [ ] Gotham
   + [ ] Iron
   + [ ] Nickel
   + [ ] Rocket
+ Scala
   + [x] Akkahttp
   + [x] http4s
[...]

PR #1328, opened 2019-05-16
[Java] Add spring boot

david-m-s:

   + [x] `travis.yml`, for CI
+ [x] A `Dockerfile` exists ?
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~


java/spring-boot/src/main/java/benchmark/springboot/BenchmarkApplication.java
```diff
@@ -0,0 +1,34 @@
...
+
+  @RestController
+  public class BenchmarkController {
+
+    @GetMapping("/")
+    public void root() {
+    }
+
+    @GetMapping("/user/{id}")
+    public String userId(@PathVariable String id) {
```
waghanza: Could `{id}` be constrained ? (to be a numerical value)

java/spring-boot/src/test/java/benchmark/springboot/BenchmarkApplicationTests.java
```diff
@@ -0,0 +1,43 @@
+package benchmark.springboot;
```
waghanza: Not needed here :stuck_out_tongue:

java/spring-boot/pom.xml
```diff
@@ -0,0 +1,43 @@
...
+		<relativePath/> <!-- lookup parent from repository -->
+	</parent>
+	<groupId>benchmark.spring-boot</groupId>
+	<artifactId>benchmark</artifactId>
+	<version>0.0.1-SNAPSHOT</version>
+	<name>benchmark</name>
+	<description>Demo project for Spring Boot</description>
+
+	<properties>
+		<java.version>1.8</java.version>
```
waghanza: Why this has to be specified ?

java/spring-boot/src/test/java/benchmark/springboot/BenchmarkApplicationTests.java
david-m-s: It is a good practice to put the test in the same package as the class under test (but in the test source folder of course).

java/spring-boot/src/main/java/benchmark/springboot/BenchmarkApplication.java
david-m-s: Sure it can but the instructions of the API to provide do not specify this. Moreover, what should be the response if the constraint does not pass?

.dependabot/config.yml
```diff
@@ -437,7 +443,13 @@ update_configs:
     directory: "/java/act"
     default_labels:
       - docker
-    
+
+  - package_manager: "docker"
```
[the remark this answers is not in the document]
david-m-s: I don't get this one. There is a block for each framework tested, and all of them have a package_manager and a default label docker... why this one should be different?

java/spring-boot/pom.xml
david-m-s: It is an extra insurance the build will be in the intended version. In the case of travis ci already ensured by the Dockerfile but on local dev it is not. So, not absolutely mandatory but a good practice.

.dependabot/config.yml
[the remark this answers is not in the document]
waghanza: If fact, this block is to track docker image (having a `PR` when **docker image** is published)
The image for `java` is already tracked (in `act` folder), no need to track twice :stuck_out_tongue:

java/spring-boot/src/main/java/benchmark/springboot/BenchmarkApplication.java
waghanza: any http error 404, 400 => better is to follow java like guidelines

PS : This is to simplify work to do on https://github.com/the-benchmarker/web-frameworks/pull/1207

FRAMEWORKS.yml
```diff
@@ -314,6 +314,10 @@ java:
     website: actframework.org
     version: "1.8"
     language: "8"
+  spring-boot:
+    website: spring.io/projects/spring-boot
+    version: "2.1.5"
```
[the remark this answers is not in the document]
david-m-s: done

java/spring-boot/src/test/java/benchmark/springboot/BenchmarkApplicationTests.java
waghanza: sure, but in this case (a benchmarking one), we do not need them + some functional test are written in `crystal`

java/spring-boot/pom.xml
waghanza: ok

.dependabot/config.yml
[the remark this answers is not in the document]
david-m-s: Removed the block, unsure if it is what was needed.

java/spring-boot/pom.xml
waghanza: If it is a good practice, leave it :stuck_out_tongue:

PR #632, opened 2018-11-15
Cloudify

waghanza:
Hi,

Results are actually computed from a `local` **docker**. This decision has been taken to facilitate **frameworks** addition, but mess-up the results. The final goal of this tool is to present closest to (as we can) **production** results.

The idea behind this `PR` is to run on a cloud, for this [DigitalOcean](https://digitalocean.com)

Frameworks :

+ C
   + [ ] agoo-c
   + [ ] kore
+ C++
   + [ ] evhtp
+ C#
   + [ ] asp NET core
+ Crystal
   + [ ] amber
   + [ ] athena
   + [ ] kemal
   + [ ] lucky
   + [ ] onyx
   + [ ] orion
   + [ ] raze
   + [ ] router.cr
   + [ ] spider-gazelle
+ Elixir
   + [ ] phoenix
   + [ ] plug
+ Go
   + [ ] beego
   + [ ] chi
   + [ ] Eecho
   + [ ] fasthttprouter
   + [ ] gf
   + [ ] gin
   + [ ] gorilla-mux
   + [ ] gorouter
   + [ ] kami
   + [ ] violetear
+ Java
   + [ ] act
   + [ ] sprint-boot
+ Kotlin
   + [ ] ktor
+ Nim
   + [ ] jester
+ Node
   + [ ] express
   + [ ] fastify
   + [ ] foxify
   + [ ] hapi
   + [ ] koa
   + [ ] muneem
   + [ ] polka
   + [ ] rayo
   + [ ] restana
   + [ ] restify
   + [ ] turbo_polka
+ Objective C
   + [ ] Criollo
+ PHP
   + [ ] laravel
   + [ ] lumen
   + [ ] slim
   + [ ] symfony
   + [ ] zend-expressive
   + [ ] zend-framework
+ Python
   + [ ] aiohttp
   + [ ] bocadillo
   + [ ] bottle
   + [ ] cyclone
   + [ ] django
   + [ ] falcon
   + [ ] fastapi
   + [ ] flask
   + [ ] hug
   + [ ] japronto
   + [ ] masonite
   + [ ] molten
   + [ ] quart
   + [ ] responder
   + [ ] sanic
   + [ ] starlette
   + [ ] tornado
   + [ ] vibora
+ Ruby
   + [ ] agoo
   + [ ] cuba
   + [ ] flame
   + [ ] hanami
   + [ ] rack-routing
   + [ ] rails
   + [ ] roda
   + [ ] sinatra
+ Rust
   + [ ] actix-web
[...]

PR #1335, opened 2019-05-18
[PHP] Add Swoole

stdex:

   + [x] `travis.yml`, for CI
+ [x] A `Dockerfile` exists ?
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/neph ${FRAMEWORK}
crystal spec
~~~

```
| php (7.3) | [swoole](http://swoole.co.uk) (master) | 7.25 ms | 6.67 ms | 15.13 ms | 25.31 ms | 72.50 ms | 5861.33 | 
| php (7.3) | [swoole](http://swoole.co.uk) (master) | 147949.67 | 375.20 MB |
```


waghanza:
Hi @stdex,

Thanks for your contribution. 

As I understand, this `PR` is more to add **fast-route** than **swoole**.

**Swoole** will be an `async` replacement for **nginx/php-fpm**. 

I think the best option for us is to replace all php implementations here, with **swoole**

stdex:
@waghanza 

1. As you can see:
>ENTRYPOINT php /usr/src/app/public/index.php start
This mean, that we start swoole server (through php-cli, without php-fpm and nginx). This is the main difference with other php benchmarks (where uses php-fpm+nginx).
2. Swoole does not include a router, so chosen as the most lightweight router (common practice, e.g.: https://github.com/easy-swoole/http/blob/master/composer.json#L20)
3. This is not framework, this is custom solution without additional overhead.

If you want I can change benchmark name to swoole_fast-route

waghanza:
@stdex I understand what you're saying,  but my concern is about :
+ `swoole` is not a **framework**
+ `swoole` could not be compared to, for example, `symfony`

BTW, I think naming this **swoole_fast-route** is acceptable

php/swoole_fast-route/Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM twosee/swoole-coroutine:latest
```
waghanza: I think `zaherg/php-swoole`  is more accurate

FRAMEWORKS.yml
```diff
@@ -172,6 +172,10 @@ php:
     website: framework.zend.com
     version: "3.1"
     language: 7.3
+  swoole_fast-route:
+    website: swoole.co.uk
+    version: "master"
```
waghanza: please fix the version

php/swoole_fast-route/Dockerfile
waghanza: at least it use `pecl` -> allowing us to not use `swoole`  **master** but the last stable (4.3 as for now)

php/swoole_fast-route/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "nikic/fast-route": "^1.3"
```
waghanza: better using `~` instead of `^`

php/swoole_fast-route/public/index.php
```diff
@@ -0,0 +1,60 @@
...
+    $r->addRoute('POST', '/user', 'post_user');
+}, [
+    'cacheFile' => __DIR__ . '/route.cache',
+]);
+
+$host = '0.0.0.0';
+$port = 3000;
+
+$server = new Server($host, $port, SWOOLE_BASE);
+$server->set(['worker_num' => swoole_cpu_num() * 2]);
```
waghanza: why 2 workers per core ?

waghanza:
May I ask for your permission for editing you `PR` ?

I think I've found a way to use `swoole` alongside `php/fpm`

stdex:
@waghanza
yes, you can edit, if it improve solution performance

php/swoole_fast-route/Dockerfile
stdex: It's a bad idea. 
In my test this leads to performance reduce. (It's maybe alpine vs debian or pecl vs compile or etc problems).
On my hw:
1st:
```
Requests / s | Throughput |
138286.67 | 350.80 MB |
```
2nd:
```
Requests / s | Throughput |
81526.33 | 206.29 MB |
```

php/swoole_fast-route/public/index.php
stdex: value from official benchmark: https://github.com/swoole/swoole-src/blob/master/benchmark/benchmark.php#L5

php/swoole_fast-route/Dockerfile
waghanza: sure, I understand the performance gap, but using **master** leads to results fluctuating, I mean we can show any version

waghanza:
@stdex This project is about having representative results for any **web-frameworks**, including `router` only **web-frameworks**

btw, after https://github.com/the-benchmarker/web-frameworks/pull/1336, I think it is great feature to represent `php` **web-frameworks** with several variants :
+ `swoole`
+ `php-pm`
+ `php-fpm` (since php built-in)

As for example, we have a performance gap (but also stability gap) between native (`php-fpm`) and async (`swoole`)

| | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
|----------------------------|----------------:|----------------:|----------------:|----------------:|----------------:|----------------:|
| swoole (4.3) | 18.32 ms | 14.47 ms | 40.87 ms | 71.97 ms | 145.94 ms | 16361.00 |
| php-fpm (7.3) | 183.03 ms | 0.82 ms | 389.20 ms | 3437.41 ms | 7278.98 ms | 617079.67 |
| swoole (master) | 9.81 ms | 7.36 ms | 23.41 ms | 38.41 ms | 78.83 ms | 9476.67 |

PS : In here there is still **docker** noise -> results are not very accurate until https://github.com/the-benchmarker/web-frameworks/pull/632

php/swoole_fast-route/public/index.php
waghanza: ok

waghanza:
@stdex I will close your `PR` since I have open https://github.com/the-benchmarker/web-frameworks/issues/1481

As this project code base is in complete rewrite, I'll do this later

Thanks for your contribution

PR #1346, opened 2019-05-21
[Swift] Update Kitura Dockerfiles to Ubuntu 18.04

djones6:
The Kitura Dockerfiles are based on ibmcom/swift-ubuntu images, which are (currently) built on Ubuntu 14.04.  There are performance penalties in this old version compared to recent releases - in particular a performance issue with libicu, which hurts Swift's string comparison performance.  In my local testing, Kitura performs ~35% faster on 18.04 compared to 14.04.

This PR replaces those with the official swift images (`swift:5.0.1` which is built on Ubuntu 18.04).  This makes for a fairer comparison with Vapor (which runs the same image), and Perfect (although this is running on `swift:4.2` which is 16.04-based).

I also updated the `kitura` version to 2.7, which seems like it was missed in #1320.

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ bin/benchmarker [tools]
 ## Results
 
 <!-- Result from here -->
-Last update: 2019-05-21
+Last update: 2019-05-22
 ```
 OS: Linux (version: 5.0.16-300.fc30.x86_64, arch: x86_64)
 CPU Cores: 8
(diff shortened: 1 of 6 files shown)


waghanza:
@djones6 Thx for catch up. Can you update all `docker` images in `swift` ?

* kitura
* kitura-nio
* perfect
* vapor

PR #1352, opened 2019-05-22
[Crystal] Parallel with processes for most in Crystal

zhangkaizhao:
* kemal
* orion
* raze
* router.cr

(amber, onyx and spider-gazelle already have this, and it seems there is no easy method to set `reuse_port` argument for `HTTP::Server` in athena at the moment)

PS:

The [TechEmpower Framework Benchmarks (TFB)](https://github.com/TechEmpower/FrameworkBenchmarks) project uses a Bash script to fork processes for most web frameworks in Crystal:
* [amber](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/amber/run.sh)
* [crystal](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/run.sh)
* [crystal-radix](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/crystal/run-radix.sh)
* [h2o.cr](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/h2o.cr/run.sh)
* [kemal](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/kemal/run.sh)
* [raze](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Crystal/raze/run.sh)

except these two
* [onyx](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Crystal/onyx)
* [spider-gazelle](https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Crystal/spider-gazelle)

which do the work in Crystal the same as here.

Other references:

* http://blog.seraum.com/crystal-cluster-class-plus-so-reuseport-to-serve-millions-of-http-requests-per-second
* https://github.com/amberframework/amber/blob/v0.28.0/src/amber/server/cluster.cr

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ bin/benchmarker [tools]
 ## Results
 
 <!-- Result from here -->
-Last update: 2019-05-22
+Last update: 2019-05-23
 ```
 OS: Linux (version: 5.0.16-300.fc30.x86_64, arch: x86_64)
 CPU Cores: 8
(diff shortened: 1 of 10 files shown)


waghanza:
@Blacksmoke16 is there a way to use `reuse_port` in `athena` ?

Blacksmoke16:
Not at this moment, I can look into adding it this weekend.  Shouldn't be too hard.

Blacksmoke16:
@zhangkaizhao should be good to implement this now.

```crystal
System.cpu_count.times do
  Process.fork do
    Athena::Routing.run(reuse_port: true)
  end
end

sleep
```

zhangkaizhao:
@Blacksmoke16 @waghanza Updated to athena 0.6.2 now.

Blacksmoke16:
@zhangkaizhao Will also need to include the config file in the docker imagine it seems.

Just add a 
```
COPY athena.yml ./
```
into the `Dockerfile`.

waghanza:
@zhangkaizhao @Blacksmoke16 results updated :tada:

PR #1428, opened 2019-06-15
[Documentation] Ensure all frameworks links are reacheable

RAnders00:
I believe the PR checklist doesn't apply for a change this simple :)


waghanza:
Good catch :tada: 

However, the link are built dynamically (fix is not here).

I'll use this `PR` to ensure this mistake will not reproduce

PR #1434, opened 2019-06-18
[Java] Add rapidoid

david-m-s:

~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~

### DIFF
--- a/java/rapidoid/Dockerfile
+++ b/java/rapidoid/Dockerfile
@@ -7,4 +7,7 @@ RUN mvn compile assembly:single -q
 FROM openjdk:11-jre-slim
 WORKDIR /rapidoid
 COPY --from=maven /rapidoid/target/benchmark-0.0.1-SNAPSHOT-jar-with-dependencies.jar app.jar
+
+EXPOSE 3000
+
 CMD ["java", "-server", "-XX:+UseNUMA", "-XX:+UseParallelGC", "-XX:+AggressiveOpts", "-cp", "app.jar", "benchmark.rapidoid.Main", "profiles=production"]
--- a/java/rapidoid/Dockerfile
+++ b/java/rapidoid/Dockerfile
@@ -1,0 +1,10 @@
+FROM maven:3.6-jdk-8 as maven
+WORKDIR /rapidoid
+COPY pom.xml pom.xml
+COPY src src
+RUN mvn compile assembly:single -q
+
+FROM openjdk:11-jre-slim
+WORKDIR /rapidoid
+COPY --from=maven /rapidoid/target/benchmark-0.0.1-SNAPSHOT-jar-with-dependencies.jar app.jar
+CMD ["java", "-server", "-XX:+UseNUMA", "-XX:+UseParallelGC", "-XX:+AggressiveOpts", "-cp", "app.jar", "benchmark.rapidoid.Main", "profiles=production"]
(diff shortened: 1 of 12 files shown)


david-m-s:
https://www.rapidoid.org/

java/rapidoid/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM maven:3.6-jdk-8 as maven
```
waghanza: As I understand, `maven` build this on `java` **8**, but this run  on `jre` **11**

java/rapidoid/Dockerfile
```diff
@@ -0,0 +1,10 @@
...
+WORKDIR /rapidoid
+COPY pom.xml pom.xml
+COPY src src
+RUN mvn compile assembly:single -q
+
+FROM openjdk:11-jre-slim
+WORKDIR /rapidoid
+COPY --from=maven /rapidoid/target/benchmark-0.0.1-SNAPSHOT-jar-with-dependencies.jar app.jar
+CMD ["java", "-server", "-XX:+UseNUMA", "-XX:+UseParallelGC", "-XX:+AggressiveOpts", "-cp", "app.jar", "benchmark.rapidoid.Main", "profiles=production"]
```
waghanza: Could you explain a little bit the purpose of this parameters ?

PS : I think it could be cool to have the same optimization options on all `java` **frameworks** here

java/rapidoid/Dockerfile
david-m-s: Yes, that is true (also on the pom.xml the source is specified as java 8), but there was no specific intent.
This means we will not be using any language feature of java 11 but the runtime environment will be more current. What do you think we should do?

java/rapidoid/Dockerfile
david-m-s: To be honest I just copied this from somewhere. After a bit of searching I found out the default GC for java 11 is G1, so the setting overriding that to use parallel GC. The agressiveOpts is deprecated in java 11, and useNUMA is taking advantage of machines capable of discriminating differend speeds of access to memory. Works with parallel GC. See https://docs.oracle.com/javase/7/docs/technotes/guides/vm/performance-enhancements-7.html#numa
This kind of fine tuning requires benchmarking to find the sweet spot. What do you think?

waghanza: Honestly, I am not in `java` world, so I'll prefer letting you decide :stuck_out_tongue: 
However, what's important is to make sure
+ spring-boot
+ rapidoid
+ act
use the same tweaks

could you homogenize docker files ?

/cc @greenlaw110

java/rapidoid/Dockerfile
waghanza: It's a little bit confusing :stuck_out_tongue: 
I suggest to use the same runtime either for build and for run

java/rapidoid/Dockerfile
david-m-s: Done for rapidoid and spring boot. For Act it is not possible since it requires java 8 runtime (it makes use of sun.misc.Unsafe, which was removed at some point before java 11). Act also has its own opinionated launcher, so I cannot force XX options. Updated the Dockerfile so that it is similar to the others, though.

waghanza: :+1:

PR #1469, opened 2019-06-26
[Crystal] Disable logging for athena benchmarking

Blacksmoke16:
Sets the `ATHENA_ENV` to `test` to disable logging when running the benchmark.  This should regain some performance as logs would not be written to `STDOUT` or the `development.log` file.

### DIFF
--- a/crystal/athena/Dockerfile
+++ b/crystal/athena/Dockerfile
@@ -8,4 +8,6 @@ COPY athena.yml ./
 
 RUN shards build --release --no-debug
 
+ENV ATHENA_ENV test
+
 CMD bin/server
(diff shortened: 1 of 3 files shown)


crystal/athena/Dockerfile
```diff
@@ -6,6 +6,8 @@ COPY src src
 COPY shard.yml ./
 COPY athena.yml ./
 
-RUN shards build --release --no-debug
+RUN shards build --production --no-debug
```
waghanza: @Blacksmoke16 what is the difference between this version and the previous one ?

Blacksmoke16: `Run in release mode. No development dependencies and strict sync between shard.yml and shard.lock.` From the shards repo.

PR #1542, opened 2019-07-15
[C++] Add the Drogon framework

an-tao:
This PR adds the [Drogon](https://github.com/an-tao/drogon) framework.


waghanza:
Hi @an-tao,

I have made a comparison between [drogon](https://github.com/an-tao/drogon) and [libevhtp](https://github.com/criticalstack/libevhtp) (the only `C++` frameworks we have here)

And results are :tada: 

| Language (Runtime) | Framework (Middleware) | Average | 50th percentile | 90th percentile | 99th percentile | 99.9th percentile | Standard deviation |
| `cpp` (`14`) | [drogon](https://github.com/an-tao/drogon) (**1.0**) | 5.88 ms | **5.10** ms | 10.32 ms | 17.48 ms | 48.45 ms | **3657.00** | 
| `cpp` (`11`) | [evhtp](https://criticalstack/libevhtp) (**1.2**) | 7.32 ms | **6.94** ms | 11.55 ms | 19.01 ms | 60.14 ms | **3659.67** | 

| Language (Runtime) | Framework (Middleware) | Requests / s | Throughput |
|---------------------------|---------------------------|----------------:|---------:|
| `cpp` (`14`) | [drogon](https://github.com/an-tao/drogon) (**1.0**) | 149510.00 | **296.09** MB |
| `cpp` (`11`) | [evhtp](https://criticalstack/libevhtp) (**1.2**) | 121979.33 | **118.40** MB |

A question however,  which version of  `C++` `drogon` is in : **14** or **17**

PS : Benchmark made on
```
OS: Linux (version: 5.1.17-300.fc30.x86_64, arch: x86_64)
CPU Cores: 4
threads: 5, connections: 1000
```

waghanza:
Also @an-tao, can we disable log ?

an-tao:
@waghanza , C++14 and C++17 are both supported and automatically detected by cmake.

an-tao:
@waghanza, Does the test output a lot of logs?

waghanza:
@an-tao I don't know. The reason why I ask that is because logs are not very useful is a `benchmarking` use case (and disabled in others **frameworks**)

waghanza:
@an-tao I display the version on results, then is this right if I say that language is `C++ 14`  ?

an-tao:
@waghanza C++ 14 is fine, I prefer C++ 14/17.

an-tao:
@waghanza I'll disable log.

an-tao:
@waghanza ,currently, only one line of logs ("start running...") is output, which is generated by the framework. I found that you set the drogon version to 1.0.0.beta2 instead of the HEAD of the master branch, with the version the log cannot be ignored. I think it doesn't matter for benchmarking.

waghanza:
Hi @an-tao,

In deed, I prefer to stick version for each frameworks. 

It adds a useful information for readers.

As for now, having 1 line of log is not very a problem, but we need to avoid it in the near future 😁

an-tao:
@waghanza no problem! Thank you for your excellent project!

cpp/drogon/CMakeLists.txt
```diff
@@ -0,0 +1,110 @@
+cmake_minimum_required (VERSION 3.2)
+Project(drogon_benchmark)
+
+link_directories(/usr/local/lib)
+link_libraries(drogon trantor pthread dl)
+
+IF (CMAKE_SYSTEM_NAME MATCHES "Linux")
```
OvermindDL1: It's very bad form to put toolchain related code in the cmakelists.txt file, this kind of stuff belongs in the toolchain.  The cmakelists.txt file should not contain anything like this, the most it should contain is the feature set request, and if you want to mandate the C++ version then you should explicitly state so, not pass compiler flags, so something [like this](https://cmake.org/cmake/help/v3.15/prop_tgt/CXX_STANDARD.html#prop_tgt:CXX_STANDARD):
```cmakelist
set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_STANDARD_REQUIRED ON)
```
`CMAKE_CXX_STANDARD` will attempt to request that version (and there are detailed feature tests if you need those as well, but you always don't), and `CMAKE_CXX_STANDARD_REQUIRED` means that the standard version specified is hard required and that the code will not fall back on older versions.

an-tao: @OvermindDL1 , thank you for your explanation, but I still have a question, my framework needs to use the `any`, I need to choose std::any or std::experimental::any according to the standards supported by the compiler, so I need to detect the version of the compiler. How can I do the above things while avoiding the wrong practices you pointed out? I don't know much about this, thank you for your correction.

OvermindDL1: @an-tao Well first of all the Standards namespace should always be used, `std::any`, any compiler that doesn't support that isn't going to support much of anything at this point (even MSVC has full C++17 support, GCC and clang both long before that), so just setting the standards version and using `std::any` is always preferred.  Second, the `std::experimental` namespace should never, absolutely Never be used in a production product.  And third, if you really must choose between them then you 'could' use a compiler (like gcc/clang) specific way to test (a macro), but that's not portable, or the proper way with CMake would be to setup a Feature test, which can either be a built-in test (don't think `any` is standalone as it is covered by C++17, so you really really should just use that), or built your own (where you just make a minimal cpp file that fails to compile if the feature doesn't exist, I.E. just try including the proper header, then cmake can set a DEFINE on if that succeeds or not, which your code can conditionally compile differently).

Again though, I would strongly *strongly* recommend never Ever to use anything from `std::experimental` in a distributed or production set of code.  Just mandate C++17 minimum, every modern compiler supports it and has for years now.

If you really want an `any` type but want it to work on pre-C++17 versions, then I'd recommend using `boost::any` (which you can include via pure cmake with a [hunter.sh](https://hunter.sh) cmake file if you want to be particularly simple, or just strip it out of boost as boost has specific tools to do just that), but still don't use `std::experimental`.

an-tao: @OvermindDL1 Thank you very much. Your answer is very clear and reasonable. I will remove std::experimental as soon as possible. In fact, some old projects (like [this](https://github.com/an-tao/drogon/issues/157)) makes me have to consider supporting compilation with C++14.
I think boost::any is a good alternative. Would you like to review the code for me after I made the modification? Thank you very much!

OvermindDL1: I'd be up for that, just ping me on whatever you need.  I've been missing C++ (been mostly in OCaml/Rust/Elixir lately because of my day job but C++ has been my life-long love).  :-)

an-tao: @OvermindDL1 thank you very much! Today I tried to strip out any and string_view from boost by it's tool `bcp`, but more than 700 files were copied to my target folder, the files number is too large for me. Maybe the method of using bcp is wrong, I will study it tomorrow.
I have never used hunter.sh, I have to take some time to study it.
As for trantor, I have done some benchmarks against libevent, However, the tests I did were not comprehensive, and from the results, the performance of trantor is slightly better. I believe that the API with modern C++ is certainly more friendly to the user than C, so I think trantor is easier to use than libevent. However, as far as the API is concerned, the asio is obviously better because it is compatible with the C++ networking TS  and is likely to be received by the new standard.

OvermindDL1: > Today I tried to strip out any and string_view from boost by it's tool bcp, but more than 700 files were copied to my target folder, the files number is too large for me.

Knowing boost that might not actually be wrong...  ^.^;

However, I bet it can still be slimmed down a lot.  `bcp` has a number of options for that.

> I have never used hunter.sh, I have to take some time to study it.

It's basically a simplified dependency management system completely built in cmake (with shared cache, etc... etc... the usual things you'd expect, made by one of the big known CMake people).

In essence copy the 'huntergate' module into your project, traditionally into the `cmake` folder (the usual place to put cmake modules, but that's only convention and not a hard requirement), a simple command if you are in your root project folder is just `mkdir -p cmake; wget https://raw.githubusercontent.com/hunter-packages/gate/master/cmake/HunterGate.cmake -O cmake/HunterGate.cmake`, then to use it just add this to the base-most cmake file (the `CMakeLists.txt` file in the root project directory) to load and initialize the module:
```cmake
include("cmake/HunterGate.cmake")
```
Then just have it load a metadata repository version (which stays cached locally), like this is the latest (you don't need the latest, just whichever that contains the metadata of what you want, you *don't* need to keep it updated unless you want something newly added, this is immutable and static for a reason, build reproducibilty):
```cmake
HunterGate(
    URL "https://github.com/ruslo/hunter/archive/v0.23.204.tar.gz"
    SHA1 "32cfed254da901f6f184027d530d8da47e035b85"
)
```
And of course the user of the project can specify arguments to their `cmake` command to disable it, have it use the system instead, etc... etc..., but by default it will acquire and build whatever you request, which for boost::any (which is included in the default boost header set with no compile-time components needed):
```cmake
hunter_add_package(Boost)
```
You can also say to use a specific version of boost, otherwise the above will use the latest version of boost in the given metadata repository file, I generally prefer to put a version, so like this instead:
```cmake
hunter_config(
    Boost
    VERSION ${HUNTER_Boost_VERSION}
)
```
Where earlier you can set the `${HUNTER_Boost_VERSION}` variable to whatever boost version as a default value (not a hard set value, I like to use variables for it as it lets the user of your library override it with ease, although with boost::any being such a basic unchanging thing just leaving it versionless may be fine, but still, I opt for versions).

To use it you use it like any other library (hunter manages adding the proper paths for the dependency it acquired and everything, so all of this "just works" without you needing to supply any Find*.cmake files or anything):
```cmake
# Find it and require it to exist, using `CONFIG` here to get automatic source
# mapping and all, this is normal CMake stuff, nothing special.
find_package(Boost CONFIG REQUIRED)

# Then link the headers into your project as normal, just add `Boost::boost` to
# your target_link_libraries line.
target_link_libraries(... Boost::boost)
```

And of course whatever other libraries hunter.sh's default metadata repository has then you can depend on them that way as well, or you can go through the work (bit of trouble as it has a lot of steps to confirm everything works right with tests and all) to put your libraries in hunter.sh if you want as well (all documented in the docs, definitely not easy like rust's cargo/crate manager, but very reliable).

Different patterns, different choices, and of course there is a lot more as well.  ^.^

an-tao: Thank you so much, your introduction is very clear.

According to my understanding, hunter.sh will download the corresponding library to the build directory instead of installing it into the system, so if `trantor` uses hunter.sh, then the application that uses `trantor` must also use hunter.sh for the libraries needed by trantor, otherwise errors may occur because libraries were not found or different version of libraries in system were used. Users of trantor must know how to use hurter.sh correctly, which is not friendly to the user.
So is hunter.sh more suitable for building applications than libraries? This makes me a bit confused. If I understand it incorrectly, please correct me. Thank you

OvermindDL1: > hunter.sh will download the corresponding library to the build directory instead of installing it into the system

Actually by default it downloads it to something like `$HOME/.hunter` or so to be used as a global cache directory to save repeated compilation time (although there is an parameter to configure that as I recall).

> so if trantor uses hunter.sh, then the application that uses trantor must also use hunter.sh for the libraries needed by trantor, 

Nope, hunter is not 'infective', you can use as little or as much as you want, further dependencies do not need to use it, rather only 'your' library in their project will use it, and they can of course disable it to manually manage the library installation as well.

> Users of trantor must know how to use hurter.sh correctly, which is not friendly to the user.

Not at all, in 99% of cases users of your library won't know that hunter even exists.  It is designed to be as transparent as possible for users.

> So is hunter.sh more suitable for building applications than libraries? This makes me a bit confused. If I understand it incorrectly, please correct me. Thank you

It is suitable for any stage, and if used in libraries it is essentially invisible to them unless they actively pass the flag to disable it in your library.  In essence, using it means you don't need to put a Boost dependency (or whatever else you want hunter to acquire) in your readme, though it might be useful to say that you auto-acquire a fully embedded boost and if you want the end-user to manage installing it manually instead then they can pass in `-DHUNTER_ENABLED=NO` to their cmake invocation and everything about hunter is just gone, meaning they have to acquire boost, they have to setup the paths for it again, with a version that is compatible with your library, etc...  There is a list of user flags at https://docs.hunter.sh/en/latest/reference/user-variables.html if you are curious of the usual set.

Hunter is not like conan or vcpkg, it does not require installing secondary software or running pre-compilation steps, nor does it require the user to even know it exists for the 99% of usages when it is used in a library, it's designed to be as transparent, yet overrideable as possible.

an-tao: @OvermindDL1 Thanks a lot!

Currently, the usage of trantor is as follows:
* git clone ....trantor
* cmake ..
* make && make install(copies the header files and library files to the system)
* and then the user creates his own project, includes trantor header files, compiles and links the library files to generate their applications. 
But trantor has the following interfaces:
https://github.com/an-tao/trantor/blob/b2162c924ad4d2678fdfa813e021620bc1044d94/trantor/net/TcpConnection.h#L62-L66
As you can see, these interfaces contain references to std::any or boost::any, if the boost is download by hunter,  It must also be installed into the trantor installation directory (maybe there is another version of Boost in system, so trantor's boost should be installed into trantor's folder), otherwise the user must pay attention to the boost dependency problem. Do I understand this right?

Thank you very much for your introduction, I have learned a lot.

OvermindDL1: > Currently, the usage of trantor is as follows:
>
> * git clone ....trantor
> * cmake ..
> * make && make install(copies the header files and library files to the system)
> * and then the user creates his own project, includes trantor header files, compiles and links the library files to generate their applications.

This wouldn't change.

> As you can see, these interfaces contain references to std::any or boost::any, if the boost is download by hunter, It must also be installed into the trantor installation directory (maybe there is another version of Boost in system, so trantor's boost should be installed into trantor's folder), otherwise the user must pay attention to the boost dependency problem. Do I understand this right?

If you expose boost then you either need to have the user depend on boost (whether system or via hunter as well, you can make a simple cmake module they can include to auto-include whatever necessary if you want to make it easy), or you need to vendor in the boost parts (bcp again).

For public interfaces I would opt for bcp to include it internally.  But then again I highly question using `any` here at all, instead the TCPConnection class should be templated on the context type instead.  That's a decent bit of processing overhead for no reason.

an-tao: Yes, the template type sounds better, how could I miss this! 

Thank you very much. ^_^

an-tao: @OvermindDL1, I have made a [PR](https://github.com/an-tao/trantor/pull/40) on trantor to remove `any`, but I don't know how to ask you to review it, when I am going to do this in the right sidebar, there is no button to invite you.

屏幕快照 2019-07-20 下午12 29 58

Regarding this PR, I sent an email to your gmail email. I would be grateful if I could get your suggestion, thank you.

waghanza: @an-tao To add @OvermindDL1 as reviewer, you need to add him as collaborator first ^^
https://help.github.com/en/articles/inviting-collaborators-to-a-personal-repository :heart: 

BTW, very participative and clear discussion here :+1:

an-tao: @waghanza Thanks a lot.

OvermindDL1: > @OvermindDL1, I have made a PR on trantor to remove any, but I don't know how to ask you to review it, when I am going to do this in the right sidebar, there is no button to invite you.

Or just ping me in the comments and I can look at it, I'm pretty easy.  ^.^

an-tao:
@OvermindDL1 @waghanza I've modified CMakeLists.txt. Please check it, thanks!

OvermindDL1:
Looks good.  :-)

OvermindDL1:
Tried compiling and it's needing both a jsoncpp dependency and drogon itself, not automatically acquired, looks like the docker image grabs these manually by apt'ing in jsoncpp and git clone'ing drogon.  Why aren't this both acquired by the build system itself?  Look at `cpp/evhtp`, it acquires all of its dependencies itself, it can be compiled straight with just cmake and a C++ compiler, nothing else needed.

an-tao:
@OvermindDL1 , I'll make them submodules of the project.

OvermindDL1:
@an-tao Eh, docker will already handle that, I just tend to compile everything manually of any given language directly as a test, I'm not a fan of acquiring dependencies manually or installing things as root.  ^.^;

I compiled drogon inline and installed it to a local directory instead, adjusting the commands as necessary, still compiling everything.

an-tao:
@OvermindDL1 I could use [this drogon docker image](https://cloud.docker.com/u/drogonframework/repository/docker/drogonframework/drogon)

OvermindDL1:
@an-tao I can't use a docker image on the server I test on.  But it seems that this project isn't depending on drogon anyway, it seems to assume it is installed globally, which is making it really difficult to state where it is actually installed.  The CMakeLists.txt file doesn't reference a drogon library at all, so it doesn't even know how to link to it.  Only thing referencing a drogon binary at all are the lines:
```cmake
LINK_DIRECTORIES(/usr/local/lib)
LINK_LIBRARIES(drogon trantor pthread dl)
```
Which absolutely should never ever be done.  Never link a system-specific global directory.  Always use `find_package` and the other primitives instead.  So the cmakelists.txt file here needs to be updated to actually use a drogon cmake library, not a system installation that can't be performed on my server here (nor many others).

an-tao:
@OvermindDL1 I understand, I still have to make some stuff like FindDrogon.cmake, right?

OvermindDL1:
> @OvermindDL1 I understand, I still have to make some stuff like `FindDrogon.cmake`, right?

`Find*.cmake` is only for finding non-cmake libraries, it's slower, inefficient, uses heuristics (guesses), etc...  You want to use a `*Config.cmake` file intead, so a `DrogonConfig.cmake` and a `TrantorConfig.cmake`.  These should be generated by the drogon and trantor libraries and they give full information about all build flags, requirements, file locations, everything, and you just point cmake to those files (either by installing in a well-known global or local location or by just passing it in on the command line) and it will pick it up perfectly every time.

Though in an end-user case, it would probably be a *lot* easier to just git submodule it in, in this case, and just `add_subdirectory` it.  :-)

For final distribution, drogon and trantor absolutely should generate cmake config files though, but its not necessary in this specific case.

Here are some useful links:
Official Docs:  https://cmake.org/cmake/help/v3.0/manual/cmake-packages.7.html
Official Tutorial on <Project>Config.cmake files:  https://gitlab.kitware.com/cmake/community/wikis/doc/tutorials/How-to-create-a-ProjectConfig.cmake-file
Proper cmake dependency handling: 
 https://cgold.readthedocs.io/en/latest/tutorials/install/managing-dependencies.html
This might not be needed for these projects but useful info anyway:  https://cmake.org/cmake/help/latest/module/GenerateExportHeader.html

an-tao:
@OvermindDL1 Great, but I have to spend some time to learn :-) Thank you!

BTW: Do you have a technical blog?

OvermindDL1:
Got it manually linked together, initial tests look very nice!

It is spitting out a *lot* of errors on its output though:
```
20190730 15:08:04.914847 UTC 19040 DEBUG [handleError] [127.0.0.1:3000--127.0.0.1:35952] - SO_ERROR = 104 Connection reset by peer - TcpConnectionImpl.cc:259
```
Over and over endlessly.  Is it supposed to be doing this?
Tests are passing without error codes though:
```
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     5.03ms    4.45ms 109.28ms   82.19%
    Req/Sec    38.49k     5.93k   60.57k    72.97%
  1906108 requests in 10.08s, 238.13MB read
Requests/sec: 189067.42
Transfer/sec:     23.62MB
```
And for comparison, here's the stock example evhtp server:
```
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     5.96ms    3.95ms 114.35ms   78.85%
    Req/Sec    33.39k     4.55k   48.67k    70.30%
  1658968 requests in 10.08s, 101.26MB read
Requests/sec: 164517.84
Transfer/sec:     10.04MB
```

One thing sticks out, it looks like it's sending excess headers that it doesn't need to, it's sending:
```
HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/html; charset=utf-8
Server: drogon
Date: Tue, 30 Jul 2019 15:15:16 GMT
```
This is not necessary for these tests, and in fact are generally not wanted as you are trying to maximize connection handling.  The `Server` and `Date` lines are needless and if you have an option to not send those, then you should set it.  Not including the date should gain you a little bit in speed alone as it saves the current time lookup call from the kernel, that way you can get even faster.  :-)

This is the minimum headers necessary for the `/` path as an example (and is all the headers that a lot of the things here send):
```
HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/plain
```

OvermindDL1:
> BTW: Do you have a technical blog?

Ehh, not really.  I keep creating and destroying them over the years.  The current one is https://blog.overminddl1.com/ but I haven't touched it in a while again and might end up wiping it, again...  ^.^;

an-tao:
@OvermindDL1 ,Date and server headers are automatically added by drogon, and currently there are no options to disable them. In fact, drogon generates a Date header every second, so it's not a big load. But removing these headers will reduce a lot of memory copies, which will definitely improve throughput:-) 
I will check those extra logs and then remove them.
I have bookmarked your blog ^.^

OvermindDL1:
> Date and server headers are automatically added by drogon, and currently there are no options to disable them. In fact, drogon generates a Date header every second, so it's not a big load. But removing these headers will reduce a lot of memory copies, which will definitely improve throughput:-)

Ah, you definitely want options for those.  Having the wrong `Server` header can screw up some API's.  And the Date is optional in the standard as well and can be quite costly at times, in addition, some API's might not want it.  The user of the library should always have the ability to override any and *all* headers sent, including removal, as it might be mandated that they have to do so at times to work with some broken API client (personal experience...).

And technically, removing those headers will actually lower 'throughput' but raise 'simultaneous connection handling'.  Removing them will raise throughput if the packets start getting large enough to get fragmented at certain boundaries, but with such simple responses that won't happen here, just increase number of connections handled, which is always good.

waghanza:
>  I could use this drogon docker image

I prefer to use a standard image, at least a image that can be used on other `C++` **frameworks**

BTW, good job :tada: with `drogon`

waghanza:
@an-tao is this  `PR` ready for you ?

an-tao:
@waghanza I will remove the Server header and Date header from responses and add drogon as a submodule to the benchmark in one or two days.

waghanza:
@an-tao ok, just let me know when this `PR` is ready :heart:

an-tao:
@waghanza I've added drogon as a submodule of the benchmark suite. Please check it.

BTW: I didn't use the `git submodule add` command because it would make drogon as a submodule of the whole project.

waghanza:
@an-tao it's ok for me, I'm only removing any unnecessary stuff

is this `PR` ready ?

an-tao:
@waghanza ，Some removed options are useful for performance. Let me modify the config.json.

waghanza:
@an-tao ok, I let you do. But options should reflect the **real-word** usage, I mean `drogon` **SHOULD NOT** be tuned here for performance

OvermindDL1:
Although extraneous headers seem fine to be removed, that is common for real world applications, especially API servers.

The current PR builds and runs very cleanly with just a `git clone` of drogon and a `rm -r build; cmake -H. -Bbuild -DBUILD_ORM=OFF -DCMAKE_BUILD_TYPE=Release && cmake --build build --config Release` now, so that's very nice!  :-)

It still logs a lot of error messages, but current results as of ten minutes ago seem worse interestingly?!:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://localhost:3000/
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency    34.85ms  138.27ms   2.00s    97.00%
    Req/Sec    13.34k     1.81k   20.06k    78.00%
  663966 requests in 10.09s, 84.85MB read
  Socket errors: connect 0, read 0, write 0, timeout 3
Requests/sec:  65772.83
Transfer/sec:      8.41MB
```
And this holds steady with multiple runs.  `evhtp` is currently benching at:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://localhost:3000/
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     6.02ms    4.23ms  69.90ms   77.07%
    Req/Sec    32.99k     5.87k   76.49k    76.11%
  1637147 requests in 10.10s, 99.92MB read
Requests/sec: 162120.09
Transfer/sec:      9.90MB
```

OvermindDL1:
There was a force push a few seconds prior, let me test with that one now, results:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://localhost:3000/
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     5.02ms    6.51ms 213.46ms   92.71%
    Req/Sec    40.49k     5.17k   63.66k    74.39%
  2009166 requests in 10.10s, 256.76MB read
Requests/sec: 198942.67
Transfer/sec:     25.42MB
```
That's a lot better now!  :-)

an-tao:
@OvermindDL1 Try the new config.json, the 'thread_num' option is remove by @waghanza , and the default value of it is 1, setting it to 0 means the number of threads equals the number of CPU cores.

an-tao:
@OvermindDL1 setting the "log_level" to "WARN" disables all DEBUG logs.

BTW: if you try to use wrk with HTTP pipelining, the performance of drogon would be improved several times

an-tao:
@waghanza I think the PR is OK, thanks!
@OvermindDL1 Thank your for your help!

OvermindDL1:
> Try the new config.json, the 'thread_num' option is remove by @waghanza , and the default value of it is 1, setting it to 0 means the number of threads equals the number of CPU cores.

Just did, that was the latest results.

> BTW: if you try to use wrk with HTTP pipelining, the performance of drogon would be improved several times

Pipelining isn't generally useful on API endpoints, which is what these are emulating, but just for testing sake:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://localhost:3000/ -s pipelining.lua
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     6.10ms    4.47ms  58.28ms   78.26%
    Req/Sec    95.04k    12.29k  136.84k    73.47%
  4694952 requests in 10.06s, 599.98MB read
Requests/sec: 466888.05
Transfer/sec:     59.66MB
```
Very nice indeed!

an-tao:
@OvermindDL1 The optimization for pipeling has no adverse effect on general API requests, after all, it adds adaptability ^_^

And I am confused about the `Transfer` data:
```
evhtp:
Requests/sec: 162120.09
Transfer/sec:      9.90MB
drogon:
Requests/sec: 198942.67
Transfer/sec:     25.42MB
```
They seem out of proportion.

OvermindDL1:
> They seem out of proportion.

Because:

evhtp:
```
╰─➤  curl -D - -G http://localhost:3000/                        
HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/plain

```
drogon:
```
╰─➤  curl -D - -G http://localhost:3000/
HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/plain
Server: drogon/1.0.0.beta4.1006
Date: Thu, 01 Aug 2019 14:43:18 GMT

```
By default evhtp doesn't add any headers than what the spec says is required, anything beyond that is the users job, and right now drogon is adding a lot of needless headers.

an-tao:
```
        "enable_server_header": false,
        "enable_date_header": false
```
These options disable the headers, please clone the v1.0.0-bate5 tag of drogon.

```
RUN git clone --branch v1.0.0-beta5 https://github.com/an-tao/drogon
WORKDIR $DROGON_ROOT
RUN git submodule update --init

WORKDIR $IROOT/build
RUN cmake -DCMAKE_BUILD_TYPE=release ..
RUN make
CMD ./drogon_benchmark ../config.json
```

OvermindDL1:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://localhost:3000/
Running 10s test @ http://localhost:3000/
  5 threads and 1000 connections
    Latency     5.12ms    6.45ms 237.08ms   91.91%
    Req/Sec    40.38k     5.77k   63.52k    74.04%
  2001877 requests in 10.10s, 122.18MB read
Requests/sec: 198228.16
Transfer/sec:     12.10MB

╰─➤  curl -D - -G http://localhost:3000/                                    
HTTP/1.1 200 OK
Content-Length: 0
Content-Type: text/plain

```
Very nice.  :-)

OvermindDL1:
Caught my other server not being busy, so testing across a gigabit network:

drogon:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://192.168.1.89:3000/
Running 10s test @ http://daggoth:3000/
  5 threads and 1000 connections
    Latency    54.97ms   69.23ms   1.82s    83.11%
    Req/Sec     6.80k   812.55     8.97k    67.20%
  338429 requests in 10.08s, 20.66MB read
Requests/sec:  33559.17
Transfer/sec:      2.05MB
```

evhtp:
```
╰─➤  /home/overminddl1/tmp/wrk/wrk -t 5 -c 1000 -d 10 http://192.168.1.89:3000/
Running 10s test @ http://daggoth:3000/
  5 threads and 1000 connections
    Latency    54.74ms   68.10ms 931.37ms   82.74%
    Req/Sec     6.85k   831.09    10.84k    71.40%
  341330 requests in 10.07s, 20.83MB read
Requests/sec:  33884.29
Transfer/sec:      2.07MB
```

It looks like drogon's throughput is mostly in better socket handling, but when that is forced over a non-local link (no fast socket setup in the kernel) it might have a touch more slowdown in other places.

an-tao:
@OvermindDL1 , The results of drogon and evhtp are so consistent. They seem to have common bottlenecks, such as the network.

OvermindDL1:
> @OvermindDL1 , The results of drogon and evhtp are so consistent. They seem to have common bottlenecks, such as the network.

Yep, that's what I was saying, that on the local system drogon is so much faster than evhtp but on the network they are the same.  ^.^

It's not because of the network itself, it's because the kernel is having to perform a full socket setup, I'm guessing drogon handles quick setups for local connections a little better than evhtp.  This can actually be useful on reverse proxied setups.

waghanza:
@an-tao Results has been updated https://github.com/the-benchmarker/web-frameworks#results

Issue #1616, opened 2019-07-30
Use more flexible constraints


waghanza:
Hi,

Maintenance stuff take times (especially update), thanks to @dependabot to reduce it.

However, this time could be reduced.

Actually, some dependencies version are specified using `X`.`Y`.`Z` format.

It could be helpful to use a less strict visioning, `X`.`Y`.

Regards,

PR #1620, opened 2019-07-31
Flexible constraint

waghanza:
Hi,

This `PR` closes https://github.com/the-benchmarker/web-frameworks/issues/1616

It enhance flexibility of constrain to reduce maintenance cost.

+ [x] c  :information_source: Not possible :information_source:
+ [x] cpp :information_source: Not possible :information_source:
+ [x] crystal
+ [x] csharp
+ [x] go :information_source: Not possible :information_source:
+ [x] java
+ [x] kotlin
+ [x] nim
+ [x] javascript
+ [x] objc
+ [x] php
+ [x] python
+ [x] ruby
+ [x] rust
+ [x] scala
+ [x] swift
+ [x] docker :information_source: Not possible for languages < 1.0 :information_source: 

Regards,


waghanza:
@fracaron how can I specify that we want any release in `1.2`, thus not only `1.2.0`

Is it in `gradle.properties` ?

waghanza:
@kirillDanshin as I understand it is not possible to use `go modules` and to define a range of version to use ?

I mean specify that we want any version between `1.2.0` and (excluded) `1.3.0`

waghanza:
@johngcn @david-m-s I have modified version constraint here, I mean I want to define a range of version allow on build (using version between X and Y).

Could you check if `pom.xml` are modified for this ? I have no skills in `java`

kirillDanshin:
> @kirillDanshin as I understand it is not possible to use `go modules` and to define a range of version to use ?
> 
> I mean specify that we want any version between `1.2.0` and (excluded) `1.3.0`

not that I'm aware of

waghanza:
@an-tao @jorisvink @ohler55 @NathanFrench @LiamRandall I want to define a `range` to ensure that the implementations `here` are using version X (define a range between X and Y).

Is there a way to do such a thing ?

waghanza:
@dom96 I want to use all version in `0.4` (`0.4.0`, `0.4.1` ...), is https://github.com/waghanza/http-benchmark/blob/flexible_constraint/nim/jester/server_nim_jester.nimble right fr that ?

waghanza:
@dom96 found it using `jester >= 0.4 & < 0.5`

PR #1685, opened 2019-08-15
[Go] Add Air

### LINKED ISSUE #1683: Add air
Hi @aofei,

Would you mind to add (or consent that we add) [air](https://github.com/aofei/air) in `here` ?

Regards,

aofei:
This closes #1683.

~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~

### DIFF
--- a/go/air/Dockerfile
+++ b/go/air/Dockerfile
@@ -1,0 +1,10 @@
+FROM golang:1.12.8
+
+WORKDIR /go/src/app
+COPY main.go go.mod ./
+
+ENV GO111MODULE=on
+RUN go build .
+
+EXPOSE 3000
+CMD [ "./main" ]
--- a/go/air/Dockerfile
+++ b/go/air/Dockerfile
@@ -1,10 +1,14 @@
-FROM golang:1.12.8
-
-WORKDIR /go/src/app
-COPY main.go go.mod ./
+FROM golang:1.12.8 AS buildenv
 
 ENV GO111MODULE=on
-RUN go build .
+ENV CGO_ENABLED=0
 
-EXPOSE 3000
-CMD [ "./main" ]
+WORKDIR /go/src/app
+ADD . ./
+
+RUN go get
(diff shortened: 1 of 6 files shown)


go/air/config.yaml
```diff
@@ -0,0 +1,4 @@
+framework:
+  github: aofei/air
+  version: 0.12.4
```
waghanza: should be on X.Y format

go/air/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.12.8
+
+WORKDIR /go/src/app
+COPY main.go go.mod ./
+
+ENV GO111MODULE=on
```
waghanza: no need :stuck_out_tongue: 

could you copy the `Dockerfile` in `go/grouter` ?
I'll try to make uniform `Dockerfile`s for all frameworks

go/air/Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM golang:1.12.8 AS buildenv
+
+ENV GO111MODULE=on
```
waghanza: @aofei this seems to be required. why is that on `go` **1.12** ?

I have the same issue with https://github.com/gogf/gf

aofei: Currently, the default value for `GO111MODULE` is "auto" (same for the upcoming Go 1.13). In this option, modules will only be enabled if the working directory isn't in `GOPATH` and contains a `go.mod` file (note that the `GOPATH` condition will be ignored in Go 1.13). Therefore, we usually explicitly set `GO111MODULE` to "on" to avoid situations where modules aren't enabled.

One day (I guess it's Go 1.14), `GO111MODULE` will be defaulted to "on".

Issue #1720, opened 2019-08-22
Update Tornado handlers to return only 200 with empty response


adrianrv:
Tornado tests are not the same as Django one. In Django only status code is set and empty response is sent. I propose to do the same in Tornado by setting only the status code to 200 and returning empty response.

PR will follow asap.

waghanza:
Hi @adrianrv,

You're right. The `dango` implementation force the status code to be 200. 

I suggest, however, to delete it instead of turning the `tornado` one. It appears that the status code is forced for `django`, it'll be better to leave the default one.

adrianrv:
What I mean is that tornado handler will be faster by removing ```self.write('')``` and instead simply returning a 200 like the Django handler is doing.

In other words, in Tornado:
```self.write('')```
should be equivalent to this in Django:
```
response = HttpResponse()
response.write('')
```
Alternative, in Tornado we can do 
```
def get(self):
    pass
````
in which case will deliver a 200 as well.

waghanza:
Not a `django` not a `tornado` expert.

The purpose here is to be as-like-as a **production-ready** code

@bdarnell What do you recommend for this case (output and empty string and return a successful status code) ?

bdarnell:
I think if the goal is to have production-like code it would have been better to specify the benchmark as requiring a non-empty body, even just a constant string like "hello world", so that every implementation has to go through its equivalent of `self.write`.

Given that the benchmark uses empty responses, though, either `self.write(b'')` or `pass` will work. `pass` will be fastest, but calling `self.write` is more realistic since almost any real app will do that, and I doubt it makes a noticeable performance difference. I'm OK with the write being there. (note that using a byte literal instead of a character string literal in the write call will be very slightly faster and still realistic)

waghanza:
Ok. Thanks for advice @bdarnell, I'll then accept you're `PR` @adrianrv

PR #1721, opened 2019-08-22
Updating tornado handler to return 200 only with empty response

adrianrv:
This solves issue #1720


waghanza:
Hi @adrianrv,

Could you make all `python` implementations uniform here, I mean letting the default behaviour for status code ?

Thanks,

PS : I didn't check all of them

PR #1570, opened 2019-07-22
Introduce variants

jkyberneees:
Using `low` HTTP server (friendly Node.js wrapper around uWebSocket.js) 
Big performance gain!

PR #1741, opened 2019-08-31
[GO] gorouter-fasthttp

vardius:

~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~


waghanza:
Hi @vardius,

Could you split your `PR` in two parts :+1: 
* one to update `gorouter`
* one to add fasthttp, but I'm not sure it's accurate

Regards,

vardius:
Sure, if its not a problem i can doing in 14 days as i took my macbook to apple this morning for battery replacement.

Fasthttp - its still same router but different adapter (not go native http)

waghanza:
@vardius Yep, I understood that. I think my comment before could be misunderstood. 

In fact, the idea to use `fasthttp` as a **replacement** for `net/http` is good. However, adding such a tool could lead to a mess (it is the same idea of using multiple **runtime** in `python` for example).

Using `fasthttp` is more accurate in https://github.com/the-benchmarker/web-frameworks/pull/1570, since **variants** are not native/built-in.
+ [turbo](https://github.com/mafintosh/turbo-http) [uWebocket](https://github.com/uNetworking/uWebSockets.js) in `javascript`
+ [hypercorne](https://pgjones.gitlab.io/hypercorn) or [daphne](https://github.com/django/daphne) in `python`
+ [falcon](https://github.com/socketry/falcon) or [iodine](https://github.com/boazsegev/iodine) in `ruby`

vardius:
many go frameworks in this benchmark already use fasthttp, i simply added support for it to my router and i’d like both to be benchmarked separately if it make sense

vardius:
two top examples from benchmark results

https://github.com/savsgio/atreugo
https://godoc.org/github.com/buaazp/fasthttprouter

waghanza:
@vardius Sure ... I like your idea, I can merge now **but** I will seperate those implementations on #1570
=> the default implementation will be with the default adapter
=> an other will be with `fashttp`, but will be **mentionned**

this separation will be made because the targeted audience (any tech decision maker) could be confused with the second implementation (as it is non default).
however, it will be shown on github and in **expert** mode (when I implement a custom blog for the results) ?

are you okay with that ?

vardius:
sure, thanks

PR #1745, opened 2019-09-02
[Python] Remove vibora

### LINKED ISSUE #1744: Remove vibora
Hi @frnkvieira,

I will remove `vibora` from this benchmark since it is completely rewritten.

Feel free to ping when the rewritte will be complete.

Regards,

waghanza:
Hi,

This `PR` closes #1744 

It siply removes https://github.com/vibora-io/vibora

This framework will be re-introduced after its rewrite

Regards,


waghanza:
@OvermindDL1 @ohler55 I've update results computed on an other workstation https://github.com/waghanza/http-benchmark/blob/19dfc11cd1761d92f5351b249050ad03b858dec6/README.md#results

Event this has been made on an https://ark.intel.com/content/www/us/en/ark/products/134899/intel-core-i7-8850h-processor-9m-cache-up-to-4-30-ghz.html results are significantly different than https://github.com/the-benchmarker/web-frameworks/blob/cb76256782185c616bbf0a3341a8c1200409155f/README.md#results

Any advice according to a sane environments (resources) to run this benchmark ?

ohler55:
Other than an overall change in performance the relative times for both latency and throughput seem similar. Is the new machine supposed to be faster than the old one?

waghanza:
The new machine is really **faster**. The question was more about -> what kind of hardware should be used to reflect real usage ?

:stuck_out_tongue:

ohler55:
That really is the question, what kind of hardware will the servers be deployed on. I don't have a great answer for that. I can say that I have seen everything from a single old machine to a cluster of hot boxes behind a firewall and of course VMs.

I've mentioned the way latency is calculated on these benchmarks does not represent real use so thats something else to consider. I wonder if having benchmarks on fast and also slow machines would be useful or if that just makes things too complicated. Right now it is easy to understand.

waghanza:
I plan to run test on either `kubernetes` (wathever hardware) and _classical vms_.

As for now, I'll run on this machine. Feel free to ping me if any idea come to your mind :heart:

Issue #1728, opened 2019-08-25
More realistic implementation


waghanza:
Hi,

The implementations here are very simple, this is on purpose. 

The routes to implement here are :

+ `GET` `/`, **SHOULD** return a successful status code and an empty string
+ `GET` `/user/:id`, **SHOULD** return a successful status code and the id
+ `POST` `/user`, **SHOULD** return a successful status code and an empty string

I propose to update code base to have a test on all (most used) HTTP methods and `application/json` instead of `plain/text`:

+ `GET` `/user/1`, **SHOULD** return a `json` representation of a user
+ `POST`, `/user`, **SHOULD** return a **successful** status code when sending a `json` representation of a user
+ `PUT`, `/user/1`, **SHOULD** return a **successful** status code when sending a **complete** `json` representation of a user
+ `PATCH`, `/user/1`, **SHOULD** return a **successful** status code when sending a **partial** `json` representation of a user
+ `DELETE`, `/user/1`, **SHOULD** return a **successful** status code

The idea is to be closest to a **real-word** reprensentation

Regards,

waghanza:
Yes. With such a system we have to add a dependency (ideally the same lib/built-in for each frameworks per language).

The purpose here is more to have a standard implementation, I mean something like in real-word.

I'm not convinced that `plain/text` API is really use in **production** nowadays, and `json` (even not the fastest format to parse/emit) is a kind of standard.

We also can use any standard format, such as `jsonapi` or else.

I'm open to any suggestions :heart:

PS : The infamous **hello world** seems to have the same problem. Which production API has this functionality

waghanza:
In addition what I've said, I think @ohler55 is working on a `graphql` implementation of this kind of benchmark.

The idea is to have this repo, as the benchmarking tool, and one repo per implementation (simple api, graphql, ...)

waghanza:
Sure, I can contribute with some `python`, `ruby` and `crystal`

You can move it here @ohler55 if you want to combien results 😀

waghanza:
Yep @ohler55, but I do not want to mix results. What I have in my mind is to have one repo per (what so called in **tfb**) scenario

waghanza:
Tfb is https://github.com/TechEmpower/FrameworkBenchmarks

Yes, I'm ready if you want to transfer ownership

waghanza:
I've granted you more privileges

waghanza:
[...]

PR #1787, opened 2019-09-17
[Ruby] Introduce grape

### LINKED ISSUE #1168: Introduce grape
+ Language : ruby
+ Repo : https://github.com/ruby-grape/grape
+ Author : @mbleigh

waghanza:
Hi,

This `PR` adds http://www.ruby-grape.org/ here.

@dblock Could you review this `PR` ?

The purpose is to have a as-close-as **real-world** implementation.

Regards,

---------

Close #1168

### DIFF
--- a/ruby/grape/config.ru
+++ b/ruby/grape/config.ru
@@ -1,0 +1,25 @@
+Bundler.require :default
+
+module Bench
+  class BaseAPI < Grape::API
+    get '' do
+      ''
+    end
+  end
+
+  class UserAPI < Grape::API
+    get '/user/:id' do
+      params[:id]
+    end
+    post '/user' do
+      ''
+    end
+  end
+
+  class API < Grape::API
+    mount ::Bench::BaseAPI
+    mount ::Bench::UserAPI
+  end
+end
+
+run Bench::API
--- a/ruby/grape/config.ru
+++ b/ruby/grape/config.ru
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 Bundler.require :default
 
 module Bench
--- a/ruby/grape/config.ru
+++ b/ruby/grape/config.ru
@@ -2,8 +2,8 @@ Bundler.require :default
 
 module Bench
   class BaseAPI < Grape::API
-    get '' do
-      ''
+    get do
+      body false
     end
   end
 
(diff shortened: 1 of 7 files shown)


ruby/grape/config.ru
```diff
@@ -0,0 +1,25 @@
+Bundler.require :default
+
+module Bench
+  class BaseAPI < Grape::API
+    get '' do
```
dblock: `get do` without an argument

ruby/grape/config.ru
```diff
@@ -0,0 +1,25 @@
...
+      ''
```
dblock: Is the intent to return no content or a stirng? If it's the former, you need to return no content and a 204. Otherwise this returns a JSON empty string by default, invoking a JSON serializer.

ruby/grape/config.ru
```diff
@@ -0,0 +1,25 @@
...
+    end
+  end
+
+  class UserAPI < Grape::API
+    get '/user/:id' do
+      params[:id]
+    end
+    post '/user' do
+      ''
```
dblock: A big part of performance benchmarking is serializing responses. I suggest returning JSON here.

waghanza: it will be done later, see https://github.com/the-benchmarker/web-frameworks/issues/1728 but not now (this requires much much work)

ruby/grape/config.ru
waghanza: The purpose is to return an empty string. 

I do not understand what you are saying. CI is passing, it test that this endpoint return an empty string (but does not check the header though)

waghanza:
Thanks @dblock. 

More that checking if code is working, the main idea is to more cleanest / standard folder structure.

Regards,

ruby/grape/config.ru
dblock: Empty string in which content type? If you're returning `application/json`, it should return `""`, if it's `text/plain`, you can't return an empty string at all. If you want not to return content it should be a 204 without a body like you changed it to. What I am saying is that all these have radically different performance characteristics because serializing data, even an empty string, is not free.

dblock:
Related, for benchmarking APIs I think it would be reasonable to compare JSON output. That's the most common scenario.

PR #1806, opened 2019-09-22
[Ruby] Add Plezi

boazsegev:
Adding the [Plezi](http://www.plezi.io) framework.

The Plezi framework was designed for small, micro services that could offer both REST and WebSocket high performance APIs. The framework can also be used as Middleware in bigger frameworks (such as Rails).

The Plezi framework isn't designed to grow more features. If anything, future releases will have features removed. It's designed to be a convenient wrapper around HTTP, SSE, WebSocket and Pub/Sub features offered by iodine, making these features more accessible.

- [x] A `Dockerfile` exists, or can be built ?
- [x] I've *manually* updated `CI` config file (and **commited** `.travis.yml`)
- [x] I've *manually* updated `.dependabot` config file  (and **commited** `.dependabot/config.yml`)
- [x] All tests passes ?

Please note:  the updates were performed manually.

### DIFF
--- a/ruby/plezi/Dockerfile
+++ b/ruby/plezi/Dockerfile
@@ -6,4 +6,4 @@ COPY Gemfile config.ru ./
 
 RUN bundle install
 
-CMD iodine -w -2 -t 2
+CMD bundle exec iodine -w $(nproc)  -p 3000 -t 16
--- a/ruby/plezi/Dockerfile
+++ b/ruby/plezi/Dockerfile
@@ -1,0 +1,9 @@
+FROM ruby:2.6
+
+WORKDIR /usr/src/app
+
+COPY Gemfile config.ru ./
+
+RUN bundle install
+
+CMD iodine -w -2 -t 2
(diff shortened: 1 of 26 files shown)


ruby/plezi/Dockerfile
```diff
@@ -0,0 +1,9 @@
+FROM ruby:2.6
+
+WORKDIR /usr/src/app
+
+COPY Gemfile config.ru ./
+
+RUN bundle install
+
+CMD iodine -w -2 -t 2
```
boazsegev: The Ruby `iodine` server settings used here might be suboptimal for the environment.

Currently, `iodine` will run the same amount of workers as half the detected CPU cores (`-2`), using 2 threads per worker:


A single thread might work better when there's no database involved. Performance is always a matter of balancing resources, which I did in a very offhanded manner.

waghanza: I understand you're concern @boazsegev. 
However
+ the implementation will be reviewed (changed) @see https://github.com/the-benchmarker/web-frameworks/issues/1728
+ I prefer to have the same optimisation in **iodine** that in **puma**

PR #1829, opened 2019-10-01
[Ruby] Use a single thread for the server

boazsegev:
When using `nproc` X 16 threads, there are 16 times the number of concurrent threads then there are processor cores.

This slows down the machine due to excessive lock contention and context switches. It only makes sense if most of the time the threads are "sleeping" or waiting on external IO (database requests).

For this simple plaintext test, a single thread would be better.

In fact, if I were to leave `nproc` behind, I would probably leave one CPU core available for OS tasks (`-w -1` would default to that state)... but maybe that could be tested across the board for everyone (by editing `nproc`).

### DIFF
--- a/ruby/plezi/Dockerfile
+++ b/ruby/plezi/Dockerfile
@@ -6,4 +6,4 @@ COPY Gemfile config.ru ./
 
 RUN bundle install
 
-CMD bundle exec iodine -w $(nproc)  -p 3000 -t 16
+CMD bundle exec iodine -w $(nproc) -p 3000 -t 1


waghanza:
Thanks @boazsegev. Can you make the same customization for puma ?

Regards,

boazsegev:
Changing this value for [Puma will result in a quiet DoS for some clients](https://bowild.wordpress.com/2019/09/24/puma-server-a-client-at-a-time-security-issue/).

`wrk` won't report the DoS, which is something of a mystery to me.

Once the security issue with Puma is fixed, we could update Puma's values. Otherwise, Puma's behavior doesn't really allow for less threads.

In fact, unless `nproc`X16 >= 1000, Puma might actually require more threads to prevent DoS.

waghanza:
Thanks @boazsegev, nice info.

Can you create an issue `here` with the link of this `puma` issue ?

This could help to change settings when the issue wille be fixed

Regards,

waghanza:
Does version `4.2` of **puma** still affected ?

boazsegev:
> version 4.2 of puma still affected?

I tested with 4.2.0 and it was still effected.

> Can you create an issue here with the link of this puma issue ?

I emailed Evan Phoenix (@evanphx) privately and I was manually observing for a patch, since I didn't want to open a security issue in the issue tracker... but maybe it's time that I did. Maybe Evan has a thought about it.

PR #1885, opened 2019-10-17
[Python] Add CherryPy

murilobr:

Just added a possible implementation.

ref #1802 

-----------------------

edit : Closes https://github.com/the-benchmarker/web-frameworks/issues/1802

### DIFF
--- a/python/cherrypy/config.yaml
+++ b/python/cherrypy/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: cherrypy/cherrypy
-  version: 18.3.0
+  version: 18.3
 
 command: >
   gunicorn  \
--- a/python/cherrypy/config.yaml
+++ b/python/cherrypy/config.yaml
@@ -1,0 +1,11 @@
+framework:
+  github: cherrypy/cherrypy
+  version: 18.3.0
+
+command: >
+  gunicorn  \
+    --log-level warning  \
+    --bind 0.0.0.0:3000  \
+    --reuse-port  \
+    --workers $(nproc)  \
+      server:app
--- a/python/cherrypy/config.yaml
+++ b/python/cherrypy/config.yaml
@@ -3,9 +3,9 @@ framework:
   version: 18.3
 
 command: >
-  gunicorn  \
-    --log-level warning  \
-    --bind 0.0.0.0:3000  \
-    --reuse-port  \
-    --workers $(nproc)  \
-      server:app
+  uwsgi \
+    --socket 0.0.0.0:3000 \
+    --async 10 \
+    --protocol=http \
+    --wsgi-file server.py \
+    --callable app
--- a/python/cherrypy/requirements.txt
+++ b/python/cherrypy/requirements.txt
@@ -1,0 +1,2 @@
+cherrypy>=18.3.0
+gunicorn
--- a/python/cherrypy/requirements.txt
+++ b/python/cherrypy/requirements.txt
@@ -1,2 +1,2 @@
-cherrypy>=18.3.0
+cherrypy>=18.3.0,<19.4.0
 gunicorn
--- a/python/cherrypy/requirements.txt
+++ b/python/cherrypy/requirements.txt
@@ -1,2 +1,2 @@
 cherrypy>=18.3.0,<19.4.0
-gunicorn
+uwsgi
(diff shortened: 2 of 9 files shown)


python/cherrypy/server.py
```diff
@@ -0,0 +1,19 @@
+import cherrypy
+
+class Root(object):
```
webknjaz: Python 3 classes are new-style by default.
```suggestion
class Root:
```

webknjaz: What is the spec for the server? Maybe it's better/cleaner to use a method dispatcher instead of traversal.

python/cherrypy/server.py
```diff
@@ -0,0 +1,19 @@
...
+    @cherrypy.expose
+    def index(self):
+        return ""
+
+    @cherrypy.expose
+    def user(self, *args):
+        if cherrypy.request.method == "POST" and len(args) == 0:
```
webknjaz: This looks weird. What is it that you're trying to achieve?

waghanza:
@murilobr @webknjaz the implementation **SHOULD** reply to 3 endpoints (more will come) :

python/cherrypy/server.py
waghanza: This seem to handle an `POST` request on `/user`

murilobr: @webknjaz I didn't found a way to use rest with method specific requests.
I tried to use popargs (https://docs.cherrypy.org/en/latest/advanced.html#the-popargs-decorator) with MethodDispatcher (https://docs.cherrypy.org/en/latest/tutorials.html#tutorial-7-give-us-a-rest) but didn't work. I had to do this shit if case.

python/cherrypy/server.py
murilobr: Ok. I will change. You need to update cherrypy documentation. Just used as it is in there.

python/cherrypy/requirements.txt
```diff
@@ -0,0 +1,2 @@
+cherrypy>=18.3.0
```
waghanza: you must set also `<19.4.0` in order to ensure to be on **18.3**
~~~python
cherrypy>=18.3.0,<19.4.0
~~~

python/cherrypy/config.yaml
```diff
@@ -0,0 +1,11 @@
+framework:
+  github: cherrypy/cherrypy
+  version: 18.3.0
```
waghanza: please use `18.3`, the format `X.Y` is more accurate to display on result page

python/cherrypy/config.yaml
```diff
@@ -0,0 +1,11 @@
...
+
+command: >
```
waghanza: you could you a non-blocking worker ? `uvloop`, `gevent`, `meinheld`

python/cherrypy/server.py
webknjaz: I believe it'd be enough to use just a method dispatcher I guess

python/cherrypy/config.yaml
webknjaz: what is the purpose of this?

webknjaz:
@murilobr change the letter case in title to `CherryPy`

murilobr:
> @murilobr change the letter case in title to `CherryPy`

done

python/cherrypy/config.yaml
waghanza: has a as-fair-as we can python benchmark, **wsgi** use `meinheld` and `asgi` use `uvicorn` (or else)

webknjaz: well, CherryPy can do WSGI on both server and app sides

webknjaz:
@murilobr I made a PR against your branch https://github.com/murilobr/web-frameworks/pull/1

murilobr:
> @murilobr I made a PR against your branch [murilobr#1](https://github.com/murilobr/web-frameworks/pull/1)

Approved

python/cherrypy/config.yaml
waghanza: fyi: I can measure an average of **169.65 ms** with `meinheld`, **233.19 ms** without any specific `gunicorn` **worker** and **166.45 ms** for `uwsgi`

webknjaz: @waghanza the doc only shows a possible use of `uwsgi`, as I said earlier, there's no "recommended" way. And it's only listed in docs because somebody contributed it there.

PR #1905, opened 2019-10-18
[Python] Add nameko

### LINKED ISSUE #1269: Introduce nameko
+ Language : python
+ Repo : https://github.com/nameko/nameko
+ Author : @mattbennett

waghanza:
Hi,

This `PR` add https://github.com/nameko/nameko to this list

Closes #1269 

Regards,

/cc @mattbennett


python/nameko/nameko.yaml
```diff
@@ -0,0 +1 @@
+WEB_SERVER_ADDRESS: '0.0.0.0:3000'
```
mattbennett: Nameko has an unhelpful default number of maximum workers. Please add the following to the config file:

```
max_workers=10000  
```

waghanza: ok, thanks

may I ask why `1000` ?

mattbennett: This is arbitrary, but it should be higher than the maximum number of concurrent requests that you reasonably expect the benchmark to generate.

waghanza: thanks for clarification, it is then required to add a comment :stuck_out_tongue:

waghanza: however, I wonder how this could be achieve on **production** ?
I mean, if I understand, this is to optimize `nameko` to a concurrency level, but in a **production** environment we **CAN NOT** know the concurrency level before deploy

mattbennett: It's not exactly an optimisation. This config setting controls the size of the global worker pool, which is really a legacy feature that is no longer required (where it matters we now limit concurrency per entrypoint, rather than globally).

[Future versions of Nameko will remove it completely](https://www.notion.so/Replace-global-worker-pool-c2607ee0c46b4881a4951f11ec08d644), but for now in order to get very high throughput you have to set it to a large number.

waghanza: ok, so this option will no longer be active in the future.

is there a github issue related ?

mattbennett: No, no github issue. Just the roadmap item I linked above.

PR #1909, opened 2019-10-21
[PHP] BasicPHP updated to v0.9.1

ray-ang:


php/basicphp/Dockerfile
```diff
@@ -0,0 +1,53 @@
+FROM php:7.3-fpm
```
waghanza: @ray-ang You can remove this file :stuck_out_tongue: 

I have made `Dockerfile` creation dynamic using https://github.com/the-benchmarker/web-frameworks/blob/master/php/Dockerfile and https://github.com/the-benchmarker/web-frameworks/blob/master/php/basicphp/config.yaml

php/basicphp/routes.php
```diff
@@ -2,33 +2,72 @@
...
+
 }
 
+/*
+|--------------------------------------------------------------------------
+| JSON-RPC v2.0 Compatibility Layer with 'method' member as 'class.method'
+|--------------------------------------------------------------------------
+*/
+
+route_rpc();
```
waghanza: is is really necessary ?

php/basicphp/routes.php
```diff
@@ -37,7 +76,7 @@
 |--------------------------------------------------------------------------
 */
 
-// route_auto();
+route_auto();
```
waghanza: is is really necessary ?

php/basicphp/controllers/AppController.php
```diff
@@ -1,9 +1,26 @@
...
+{
+
+	public function index()
+	{
+		echo '';
+	}
+
+	public function viewUser()
+	{
+		echo url_value(2);
```
waghanza: does not return user id => https://travis-ci.com/the-benchmarker/web-frameworks/jobs/247934237#L508

php/basicphp/Dockerfile
ray-ang: Removed Dockerfile as advised.

php/basicphp/routes.php
ray-ang: I wanted to know how the framework performs if the RPC routing is enabled since this would be used in actual application.

php/basicphp/controllers/AppController.php
ray-ang: Apparently, $_SERVER['REDIRECT_URL'] is only applicable in Apache server... had to revert to request_uri.

waghanza: I have a `404` for **POST** on `/user`

waghanza:
@ray-ang I have a `404` for **POST** on `/user`

ray-ang:
@waghanza I will check this out since it works in my machine...

ray-ang:
The use of REQUEST_URI throughout the framework would hopefully eliminate incompatibilities with Nginx...

waghanza:
@ray-ang I've messed your **master**, could you please **force** push.

I have found the solution using `try_files` on `nginx` **config**

waghanza:
@ray-ang :warning: do not `git pull` before ... only `git push -f` without anything before

ray-ang:
@waghanza I updated my master branch using the updated files - AppController.php, index.php, config.php, functions.php and routes.php. Good thing only a few files in all for the framework. :-) I use git using GitHub via web and not CLI, so am not familiar with force push via CLI. So what I did is to add the commits through this pull request.

PR #1912, opened 2019-10-22
[JAVA] Add micronaut framework

gnomeria:

~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
export FRAMEWORK=micronaut
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~


waghanza:
why closed ? any contribution is valuable

gnomeria:
I still need to update the ci build etc I guess. I'll reopen it later when I'm done @waghanza.

waghanza:
@gnomeria you could re-open, i'll help (make a `PR` to merge in this) if needed :heart:

gnomeria:
Also my project was currently in gradle, does the dependabot support it?

waghanza:
@gnomeria I am surely not able to determine which is better for your needs, `maven` or `gradle`
what I can say is that `here` we use `gradle` for `kotlin` and `maven` for `java`, since it is the most common use case

however, if you prefer to use `gradle` in can turn this project using `bot`. I just need a functional implementation :stuck_out_tongue:

java/micronaut/config.yaml
```diff
@@ -0,0 +1,16 @@
+framework:
+  website: micronaut.io
+  version: 1.2.5
+
+image: gradle
```
waghanza: should be 
```yaml
deps:
  - gradle
```

I have made a slight modification to code base so as to run `java` frameworks on `alpine`

java/micronaut/config.yaml
```diff
@@ -0,0 +1,16 @@
+framework:
+  website: micronaut.io
+  version: 1.2.5
```
waghanza: it should be 1.2, using X.Y.Z is too heavy, in term of maintenance

java/micronaut/gradle.properties
```diff
@@ -0,0 +1 @@
+micronautVersion=1.2.5
```
waghanza: could it be a range ?
using `1.2.5` is too strict, I do not know `gradle`, but I think using
```ini
micronautVersion=1.2.+
```
should work as expected

gnomeria: sure i'll update it

java/micronaut/config.yaml
gnomeria: so i just need to put as :
```
image: gradle
deps:
  - gradle
```

waghanza:
@gnomeria could you run
~~~sh
shards install
shards build
bin/make ci_config
bin_make deps_config
~~~
so as to resolve conflicts (`.travis.yml` and `.dependabot/config.yml` are generated) ?

waghanza:
@gnomeria Still failing, since `gradle` version is **4.10.3** on `alpine`

I'll fix that tomorrow

gnomeria:
> @gnomeria Still failing, since `gradle` version is **4.10.3** on `alpine`
> 
> I'll fix that tomorrow

I forgot about the gradle version. Previously I thought I could just define my own `Dockerfile`, so I just install the gradle version on the alpine by:
```bash
ENV GRADLE_VERSION 5.5
# get gradle and supporting libs
RUN apk -U add --no-cache curl; \
  curl https://downloads.gradle.org/distributions/gradle-${GRADLE_VERSION}-bin.zip > gradle.zip; \
  unzip gradle.zip; \
  rm gradle.zip; \
  apk del curl; \
  apk update && apk add --no-cache libstdc++ && rm -rf /var/cache/apk/*
ENV PATH=${PATH}:/gradle-${GRADLE_VERSION}/bin
```

waghanza:
> Previously I thought I could just define my own Dockerfile

Could be that, but I try to have only `Dockerfile` per language, in order to keep run _fair_ (I mean running on the same version of **language** ...)

PR #1944, opened 2019-10-24
[PHP] add Ubiquity

jcheron:
It is a Micro-application version, which corresponds exactly to the specifications requested at the moment (see https://github.com/the-benchmarker/web-frameworks/issues/1925).

Before submitting y our PR, please review the following checklist :

~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~

--------------
edit : Closes #1944


waghanza:
> Strange since the previous command seems to install sqlite3 

You must install the `devel` package

jcheron:
what does that mean in practice?
> You must install the `devel` package

waghanza:
sorry @jcheron, `devel` packages contains headers used for compilation / linking / external use

+ on `debian` based, by convention, the suffix is **-dev**
+ on `rhel` based, by convention, the suffix is **-devel**

so I think, you might install **libsqlite3-dev** depending on your `distribution`

jcheron:
Okay, it's moving forward,
I still have a little error at the end of the build:

>Successfully built 5fa028f7d811
Successfully tagged ubiquity:latest
[Error]  -- STDERR(.neph/ubiquity/log/log.err) -- 
Unhandled exception: no such table: languages (SQLite3::Exception)
  from ???
  from /usr/share/crystal/src/hash.cr:883:11 in 'build_statement'
  from /home/jc/web-frameworks/lib/db/src/db/statement.cr:57:5 in 'build'
  from /home/jc/web-frameworks/lib/db/src/db/query_methods.cr:271:7 in 'run'
  from /home/jc/web-frameworks/tools/src/client.cr:84:1 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:47:14 in 'main'
  from __libc_start_main
  from _start
  from ???

waghanza:
You have not initialized the database
```sh
bin/db init
~~~

:information_source: This will create a sqlite database n `data.db`

**Comment by jcheron** [comment]
I'm trying to start **nginX** with a specific configuration (start script in the **deploy** folder), provided as a command (see https://github.com/the-benchmarker/web-frameworks/pull/1944/commits/5e7474ee0452f043284d8b4e76d22ec54702a439).
The execution:
```bash
bin/neph ${FRAMEWORK}
```
Ends with an error:
>Successfully tagged ubiquity:latest
[Error]  -- STDERR(.neph/ubiquity/log/log.err) -- 
Unhandled exception: Index out of bounds (IndexError)
  from ???
  from /usr/share/crystal/src/channel.cr:176:14 in 'run'
  from /home/jc/web-frameworks/tools/src/client.cr:84:1 in '__crystal_main'
  from /usr/share/crystal/src/crystal/main.cr:47:14 in 'main'
  from __libc_start_main
  from _start
  from ???

The log file (.neph/ubiquity/log/log.err) contains no errors.

waghanza:
This kind of error is due to an unreachable host, the framework is not listen on the right port (or any other network issue)

I'll see this tomorrow

waghanza:
Hi @jcheron,

I've seen the configuration files used on **tfb**, to understand how  `ubiquity` is working. I've noticed some differences `here`:

+ `realpath_cache_ttl` in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L347, the directive does not exists in **official** docker image
+ `expose_php` in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L374 is `Off`, it's `On` in **official** docker image
+ `cgi.fix_path_info` in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L776 is `0`  , it is `1` in **official** docker image
+ `date.timezone` is **UTC** in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L939, the directive is disable in **official** docker image
+ `session.gc_probability` is **0** in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L1413, it is **1** on the **official** docker image
+ `opcache.enable_file_override` is **1** in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L1808, we do not have this directive `here`
+ `opcache.huge_code_pages` is **1** in https://github.com/TechEmpower/FrameworkBenchmarks/blob/aa42742681c84434aeaad2db55fa18cab5b426d1/frameworks/PHP/ubiquity/deploy/conf/php.ini#L1881, we do not have this directive `here`

Why are those configuration made ?

jcheron:
Hi @waghanza 
It's just optimization, unrelated to the problem.

if I do the dockerfile manually with the same parameters as the ones I put in my bootstrap.sh file, it works fine.
The problem is related to launching nginX with the shell script....

waghanza:
@jcheron Please a working version of your `Dockerfile` (seems not to work with this version)

php/ubiquity/Dockerfile-u
```diff
@@ -0,0 +1,57 @@
...
+COPY . ./
+
+RUN curl -sSL https://getcomposer.org/installer | php -- --install-dir=/usr/local/bin --filename=composer
+RUN composer install --no-dev --prefer-dist --classmap-authoritative
+RUN composer dumpautoload -o
+
+
+  RUN composer update --dev
+  RUN chmod +x vendor/bin/Ubiquity
+  RUN ./vendor/bin/Ubiquity bootstrap dev
```
waghanza: why use `dev` ?

php/ubiquity/.htaccess
```diff
@@ -0,0 +1,10 @@
+AddDefaultCharset UTF-8
```
waghanza: can we remove this file ?

I mean this is for `apache` compatibility, right ?

waghanza:
@jcheron Using `2.3.1`, I have somes **[Composer\Downloader\TransportException]**. any thought ?

php/ubiquity/Dockerfile-u
jcheron: I saw that the `composer update` was in `dev` by default
so the `--dev` is useless.
But other than that, the docker makes a composer install --no-dev
However, I have to install `Ubiquity-devtools` to generate the cache.

php/ubiquity/.htaccess
jcheron: yes, we can!

jcheron:
> @jcheron Using `2.3.1`, I have some **[Composer\Downloader\TransportException]**. any thought ?

It seems there are problems right now, on some packagist repos:
see https://github.com/composer/packagist/issues/940

php/ubiquity/Dockerfile-u
waghanza: Perhaps, `Ubiquity-devtools` in **production** setup. Indeed, cache generation is accurate in **production**

php/ubiquity/.ubiquity/config.php
```diff
@@ -0,0 +1,2 @@
+<?php
+return array("host"=>"127.0.0.1","port"=>8090,"sessionName"=>null);
```
waghanza: what is this configuration for ?

php/ubiquity/README.md
```diff
@@ -0,0 +1,43 @@
+# Ubiquity Benchmarking Test
```
waghanza: could we remove this file ?
Using a `README` sounds like a `best-practice` but in this case it **NOT** relevant (all implementations **SHOULD** do the same thing)

php/ubiquity/app/cache/controllers/routes.default.cache.php
```diff
@@ -0,0 +1,2 @@
+<?php
```
waghanza: can we **NOT** create `cache` on container ?

php/ubiquity/composer.json
```diff
@@ -0,0 +1,27 @@
+{
+	"require" : {
+		"php" : "^7.1",
+		"phpmv/ubiquity" : "dev-master"
```
waghanza: please stick to the last version, ideally
~~~json
"phpmv/ubiquity" : "~2.3.1"
~~~
to stick to `2.3`

php/ubiquity/composer.json
```diff
@@ -0,0 +1,27 @@
...
+	},
+	"require-dev": {
+		"mindplay/annotations": "^1.3",
+		"phpmv/ubiquity-dev": "dev-master",
+		"phpmv/ubiquity-devtools": "dev-master"
```
waghanza: should it not be used for **production** also ? (despite the name)

php/ubiquity/composer.json
```diff
@@ -0,0 +1,27 @@
...
+	},
+	"autoload" : {
+		"psr-4" : {
+			"" : "app/"
+		}
+	},
+	"minimum-stability": "dev",
```
waghanza: not good, the purpose is to have **production-ready** code

php/ubiquity/.ubiquity/config.php
jcheron: Hi @waghanza 
this is for the embedded php server:
is used to execute:
`Ubiquity serve`
But irrelevant here

jcheron:
> Could you clean your code base ? Using **PSR2**

No, I am fighting against the `PSR-2` dictate. (see https://github.com/phpMv/ubiquity/issues/47)

php/ubiquity/composer.json
jcheron: no, this is only for dev

php/ubiquity/app/cache/controllers/routes.default.cache.php
jcheron: I don't understand the question.
Do you want me to remove this file?
It is automatically generated when the container is created

php/ubiquity/composer.json
jcheron: I think we should leave the freedom of choice of version to the creators. If I use a dev version, it's to eventually optimize without having to do a PR.
In a permanent run context like on tfb, this is important.

php/ubiquity/Dockerfile-u
jcheron: Once in production, `devtools` are no longer useful and unnecessarily overload the composer autoloading.

waghanza:
> No, I am fighting against the PSR-2 dictate.

I understand, I have a different point of view (especially for **tabs**) ... but the most important in **NOT** to choose about a coding format, but more to use a common one (and why not have exceptions)

It appears that in `php` **world** this is a `best-practice`, even we can argue against. For example, in `ruby` **world** there is `rubocop` (not the same but the most used), even if it is not ideal, it is a `best-practice` to use (and more if you are collaborating)

Disclaimer : With `rubocop` we can customize rules, I do no know if we can do the same in `php`

jcheron:
> I understand, I have a different point of view (especially for **tabs**) ... but the most important in **NOT** to choose about a coding format, but more to use a common one (and why not have exceptions)

This is one of the problems with PSRs:
Most people think that if you don't comply with them, then no standards are met, and that it has something to do with the quality of the code!

Some choices can be made at the team level or in a project, , whether it is a simple choice of presentation, or more complex design choices.
**Ubiquity** respects some PSRs, with adjustments, and violates others, by design choice: see https://micro-framework.readthedocs.io/en/latest/contributing/coding.html#coding-standards

waghanza:
I understand that you don't want to do use PSRs for **ubiquity**, but I don't want you to do that for this framework, but only for the implementation here

php/ubiquity/composer.json
waghanza: I mean **devtools**, there `here` to do some stuff like warming up caches no ?
Some kind of things we **MUST** run on **production**, as an example, I had to add trhe console manager on `symfony` to warmup the cache on **production**
https://github.com/the-benchmarker/web-frameworks/blob/master/php/symfony/config.yaml#L15

jcheron: **devtools** allow the generation of the cache for production, indeed. But you don't leave them in place in production, for performance reasons.
You just have one single command to do that:
`Ubiquity bootstrap prod`
This should also be the case for the Symfony console.

waghanza: I'm not a  `php` dev, but for example, in `rails` we generally leave the file `bin/rails` in **production**.
having those **binstubs** (`bin/rails` is one of them) are used in **production** to migrate the database (as an example) so having them in **production** is required.
It seems that `ubiquity/devtools` is the same, no ?

jcheron: No, because **devtools** can be installed globally (with `composer global require phpmv/ubiquity-devtools`).
This avoids **dev** parts in a project in production.

php/ubiquity/composer.json
```diff
@@ -1,12 +1,12 @@
 {
 	"require" : {
 		"php" : "^7.1",
-		"phpmv/ubiquity" : "~2.3.2"
+		"phpmv/ubiquity" : "~2.3.2",
+		"phpmv/ubiquity-devtools": "~1.2.10"
```
jcheron: As I have already explained to you (see https://github.com/the-benchmarker/web-frameworks/pull/1944#discussion_r341587296), I disapprove this practice, since the **devtools** can be installed globally, without overloading the **composer** autoloading for the project.

PR #1945, opened 2019-10-25
Add plain fasthttp

### LINKED ISSUE #1915: Add fasthttp
Hi @valyala,

Would you mind to add (or consent that we add) [fasthttp](https://github.com/valyala/fasthttp) in `here` ?

Regards,

erikdubbelboer:
Improve fasthttprouter, fmt is slow.

Fixes #1915


go/fasthttp/go.mod
```diff
@@ -0,0 +1,5 @@
+module main
+
+require github.com/valyala/fasthttp v1.5.0
+
+go 1.13
```
waghanza: this line is not required

go/fasthttp/go.sum
```diff
@@ -0,0 +1,13 @@
+github.com/klauspost/compress v1.8.2 h1:Bx0qjetmNjdFXASH02NSAREKpiaDwkO1DRZ3dV2KCcs=
```
waghanza: keeping the `.sum` is useful, but not for this use-case

go/fasthttprouter/go.mod
```diff
@@ -1,4 +1,8 @@
 module main
 
+require (
+	github.com/buaazp/fasthttprouter v0.1.1
+	github.com/valyala/fasthttp v1.5.0
+)
 
-require github.com/buaazp/fasthttprouter v0.1.1
\ No newline at end of file
+go 1.13
```
waghanza: not required

go/fasthttprouter/go.sum
```diff
@@ -0,0 +1,15 @@
+github.com/buaazp/fasthttprouter v0.1.1 h1:4oAnN0C3xZjylvZJdP35cxfclyn4TYkW6Y+DSvS+h8Q=
```
waghanza: keeping the `.sum` is useful, but not for this use-case

go/fasthttprouter/main.go
```diff
@@ -1,22 +1,20 @@
...
-
 	"github.com/buaazp/fasthttprouter"
 	"github.com/valyala/fasthttp"
 )
 
 func main() {
 	router := fasthttprouter.New()
 	router.GET("/", func(ctx *fasthttp.RequestCtx) {
-		fmt.Fprint(ctx, "")
+		ctx.SetBodyString("")
```
waghanza: thanks :tada:

go/fasthttp/main.go
```diff
@@ -0,0 +1,40 @@
...
+	"bytes"
+
+	"github.com/valyala/fasthttp"
+)
+
+var (
+	strGET            = []byte("GET")
+	strPOST           = []byte("POST")
+	strSlash          = []byte("/")
+	strSlashUserSlash = []byte("/user/")
```
waghanza: is there any way to have a dynamic URL ?

I mean `strSlashUserSlash` could lead either to `/user/0` than `/user/$hacking_value*` ?

go/fasthttp/go.mod
erikdubbelboer: It's automatically added by Go 1.13 each time you run any of the `go mod` commands.

go/fasthttprouter/go.sum
erikdubbelboer: The whole idea behind the `go.sum` file is to commit it so you have reproducible builds that always use the same dependencies.

go/fasthttp/main.go
erikdubbelboer: That's why I'm using `bytes.HasPrefix` to check. There is no default router build into fasthttp.

go/fasthttprouter/go.sum
waghanza: sure, I know, like a `lock` file in `ruby` ..., but this also means that be versioned ... as this is not an end application, the benefit is not accurate `here`

could it be removed ?

PS : We have no `.sum` files here, I think it's **NOT** needed here :stuck_out_tongue:

go/fasthttp/main.go
waghanza: ok, it seems different than other frameworks I know :heart:

erikdubbelboer:
I have removed the things you requested.

PR #1955, opened 2019-10-29
[C++] Update drogon to 1.0b9

an-tao:


waghanza:
@an-tao I've updated results with the latest `drogon`'s beta

PR #1958, opened 2019-10-30
[spider-gazelle] enable clustering and update shards

stakach:


crystal/spider-gazelle/shard.yml
```diff
@@ -6,7 +6,7 @@ licence: MIT
 dependencies:
   action-controller:
     github: spider-gazelle/action-controller
-    version: ~> 1.6.0
+    version: ~> 2.0
```
waghanza: Is this saying that we are using **spider-gazelle** `2.0` ?

stakach: Yeah, I am tracking it as version.feature_release.bug_fix
So version 2 as we re-wrote the logging system which required configuration changes for all existing applications. Feature releases will always be non-breaking changes.

waghanza: ok, so could you also update file `crystal/spider-gazelle/config.yaml`

is what you are saying signify that `action-controller` version is the **same** than `spider-gazelle` ?

PR #1968, opened 2019-11-02
Use alpine to build all containers

waghanza:
Hi,

This `PR` is about to use  `alpine` to run **frameworks** `here` :

- [x] C
- [x] Clojure
- [x] C++
- [x] Crystal
- [x] C#
- [x] F#
- [x] Go
- [x] Java :information_source: I've patch the `Dockerfile`, needs to be fixed
- [x] Kotlin
- [x] Nim
- [x] Javascript
- [x] Julia :warning: No **official** alpine image found
- [x] Perl :warning: No **official** alpine image found
- [x] Php
- [x] Python
- [x] Ruby
- [x] Rust :information_source: Can **not** build, but can run on **alpine**
- [x] Scala :warning: There is no **official** image, but it runs on **alpine** and **openjdk**
- [x] Swift :warning: `Swift` can not run on **alpine**

The **main** reason behind that decision is https://brianchristner.io/docker-is-moving-to-alpine-linux/

Regards,


waghanza:
@OvermindDL1 any idea to compile https://github.com/criticalstack/libevhtp on **alpine** ?

OvermindDL1:
> @OvermindDL1 any idea to compile criticalstack/libevhtp on alpine ?

I'd be surprised if it didn't "Just work" once the deps are installed (like gcc, cmake, etc...).  Did you get an error?

waghanza:
@OvermindDL1 maybe I missing something (to setup a build dependency, or else)

`cmake -H. -B_builds -DCMAKE_BUILD_TYPE=Release` leads me to this

+ log file => https://gist.github.com/5f9cc77aa3caf909266497aa730e00f4 
+ error file => https://gist.github.com/7c8cfbf9e0bc4f6ff6db93eda773a147

OvermindDL1:
Perl is a dependency of openssl, and perl should be default included in any posix standard system, so it's confusing that the alpine image doesn't already have it (does it not include all the standard posix stuff?), just install perl5 and it should be good.

waghanza:
> Perl is a dependency of openssl, and perl should be default included in any posix standard system, so it's confusing that the alpine image doesn't already have it (does it not include all the standard posix stuff?), just install perl5 and it should be good (at least fixing 'that' error, who knows what else alpine is messed up on, lol).

Thanks @OvermindDL1, with `perl` and `linux-headers` installation, it works :+1: 

However, I do not have to install `perl` for **drogon**

OvermindDL1:
> However, I do not have to install perl for drogon

It doesn't build openssl for SSL connections, instead it links the current linkable runtime by default (if configured to use SSL at all, which I'm not sure it is for these tests, so it probably doesn't even touch it).

waghanza:
@dalisoft How can I run `nanoexpress`/`nanoexpress-pro` on **alpine** ?

waghanza:
I've found a fix in https://nanoexpress.js.org/docker.html

dalisoft:
Yes, you can! Thanks a lot for fix

PR #1982, opened 2019-11-07
[Javascript] Add nanoexpress

dalisoft:

+ [ ] I've updated `CI` config file (and **commited** `.travis.yml`)
~~~sh
bin/make ci_config
~~~
~~~sh
bin/make deps_config
~~~
+ [x] All tests passes ?
~~~
shards install
shards build --release
bin/make config
bin/neph ${FRAMEWORK}
crystal spec
~~~

### DIFF
--- a/javascript/nanoexpress/config.yaml
+++ b/javascript/nanoexpress/config.yaml
@@ -1,5 +1,4 @@
 framework:
-  github: dalisoft/nanoexpress
   website: nanoexpress.js.org
-  version: 1.1.4
+  version: 1.1
   
--- a/javascript/nanoexpress/config.yaml
+++ b/javascript/nanoexpress/config.yaml
@@ -1,0 +1,5 @@
+framework:
+  github: dalisoft/nanoexpress
+  website: nanoexpress.js.org
+  version: 1.1.4
+  
--- a/javascript/nanoexpress/package.json
+++ b/javascript/nanoexpress/package.json
@@ -1,5 +1,5 @@
 {
   "dependencies": {
-    "nanoexpress": "^1.1.4"
+    "nanoexpress": "^1.1"
   }
 }
--- a/javascript/nanoexpress/package.json
+++ b/javascript/nanoexpress/package.json
@@ -1,0 +1,5 @@
+{
+  "dependencies": {
+    "nanoexpress": "^1.1.4"
+  }
+}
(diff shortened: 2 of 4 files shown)


javascript/nanoexpress/config.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+  github: dalisoft/nanoexpress
+  website: nanoexpress.js.org
```
waghanza: could you use only one markup, rather `website`

javascript/nanoexpress/config.yaml
```diff
@@ -0,0 +1,5 @@
...
+  version: 1.1.4
```
waghanza: could you use X.Y format ?

javascript/nanoexpress/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "nanoexpress": "^1.1.4"
```
waghanza: could you use `~1.1.4` to keep in `1.1` ?

dalisoft:
@waghanza Pro is licensed under GPL-3.0 and about code-changes

- Faster
- API more similar to Express than default version
- Less configurations than default version

All other things are same as default and code-base as well as plugins are backwards-compatible with default version.

javascript/nanoexpress/package.json
dalisoft: I fixed correctly? I want use always use latest ver in benchmarks

javascript/nanoexpress/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "nanoexpress": "^1.1"
```
waghanza: rather use tilde @dalisoft 

to stick into the minor

use carret could allow usage of `1.2` or `1.9`

dalisoft: I think @dependabot does the job for update, right?

I changed :)

waghanza: Yes

waghanza:
could you rebase and resolve conflict ?

either manually, or with `bin/make deps_config` :stuck_out_tongue:

dalisoft:
just 5 min, i will do :)

dalisoft:
done

Issue #446, opened 2018-10-21
Add Spring(+Java)


mcauto:
+ Language : Java
+ Repo : https://github.com/spring-projects/spring-framework
+ Author : spring-projects
Spring Framework is the populest web framework in Korea.

waghanza:
@greenlaw110 interested in contributing (comparing with `act`) ?

david-m-s:
https://github.com/the-benchmarker/web-frameworks/pull/1418

waghanza:
@david-m-s spring and spring-boot are the same ?

david-m-s:
I just searched for someone that had put it in a concise way. From https://www.baeldung.com/spring-vs-spring-boot:

"Spring Boot is basically an extension of the Spring framework which eliminated the boilerplate configurations required for setting up a Spring application.

It takes an opinionated view of the Spring platform which paved the way for a faster and more efficient development eco-system."

I'd say for the sake of benchmarking they are the same.

waghanza:
ok understood, `spring-boot` is then `spring` but with a more efficient eco-system

I think, then, it could be intersting to have both here :
+ To see the performance gap of the `-boot` part
+ To see the differences between each codebase

But, it will be good to add this response into a documentation

waghanza:
@david-m-s do you want to add an implementation of `spring` (classical) `here` ? I think it could be good to display result for the same implementation for `spring-boot` and `spring`

PR #2009, opened 2019-11-21
Add java spring-framework

david-m-s:
Add implementation with bare spring framework (no spring boot). Use jetty runner as servlet container.

#446

### DIFF
--- a/java/spring-framework/pom.xml
+++ b/java/spring-framework/pom.xml
@@ -1,0 +1,40 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>benchmark.spring-framework</groupId>
+    <artifactId>benchmark</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <name>benchmark</name>
+    <description>Demo project for Spring Framework</description>
+    <packaging>war</packaging>
+
+    <parent>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-framework-bom</artifactId>
+        <version>5.2.0.RELEASE</version>
+    </parent>
+    <properties>
+        <java.version>1.8</java.version>
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <failOnMissingWebXml>false</failOnMissingWebXml>
+    </properties>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webmvc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.10.1</version>
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <version>4.0.1</version>
+            <scope>provided</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/java/spring-framework/pom.xml
+++ b/java/spring-framework/pom.xml
@@ -28,7 +28,7 @@
         <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-databind</artifactId>
-            <version>2.10.1</version>
+            <version>[2,3)</version>
         </dependency>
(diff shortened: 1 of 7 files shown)


java/spring-framework/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+    <artifactId>benchmark</artifactId>
+    <version>0.0.1-SNAPSHOT</version>
+    <name>benchmark</name>
+    <description>Demo project for Spring Framework</description>
+    <packaging>war</packaging>
+
+    <parent>
+        <groupId>org.springframework</groupId>
+        <artifactId>spring-framework-bom</artifactId>
+        <version>5.2.0.RELEASE</version>
```
waghanza: could you use a less restict constraint
```xml
<version>[5.2,5.3)</version>
```
to ensure that we stick on **5.2** :+1: 

PS : I've also notice that it is not the case for `spring-boot`

java/spring-framework/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+
+    <dependencies>
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-webmvc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+            <version>2.10.1</version>
```
waghanza: could you remove this ? 
it's add more maintenance (e.g : we do not care about which version of each dependency, but we care about version of **frameworks** in use)

java/spring-framework/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+        </dependency>
+        <dependency>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+            <version>4.0.1</version>
```
waghanza: could you remove this ? 
it's add more maintenance (e.g : we do not care about which version of each dependency, but we care about version of **frameworks** in use)

java/spring-framework/pom.xml
david-m-s: It is necessary to specify this dependency because it is required for the /user/{id} endpoint, but spring marks it as optional.

java/spring-framework/pom.xml
david-m-s: This is necessary as we are writing a servlet we must specify which version of the spec it is in.

java/spring-framework/pom.xml
david-m-s: Done

java/spring-framework/pom.xml
waghanza: in fact I mean specify the version of dependency, not dependency itself

java/spring-framework/pom.xml
waghanza: in fact I mean specify the version of dependency, not dependency itself

java/spring-framework/pom.xml
david-m-s: Sorry let me make sure I understand you: you mean the version should be extracted like a property? e.g. <version>${jackson-databind.version}</version>

waghanza: nop
I mean, this version tag is not very much required, in my opinion.
Because :
+ it adds more maintenance / updates
+ it could lead to mistaken result in case of this is outdated (in the future)

java/spring-framework/pom.xml
david-m-s: In this case we should not update the dependency: The current spring version targets servlet 4.0 spec, which is the last one. Since spring 5.2 will not target any greater servlet version we can be sure this one does not need maintenance.

java/spring-framework/pom.xml
david-m-s: I'm confused. The dependency to some version of jackson-databind is mandatory. Then to specify it in maven, there is no other way than adding the dependency and the version because spring framework lists it as optional.
See https://github.com/spring-projects/spring-framework/blob/21d751d22202c6df2d0144b155979c37b45fb167/build.gradle#L46

However we can stick to 2.10.x or even 2.x by using the [2.10,2.11) notation.

java/spring-framework/pom.xml
waghanza: fair enough

however, is there any way to say `use the latest version available` ?

java/spring-framework/pom.xml
waghanza: Yeah.

This is not ideal, but helps a lot :heart: 

Could you check for the same thing on all `java` **frameworks** `here` ?

java/spring-framework/pom.xml
waghanza: @david-m-s could you use `[4,5)` for  `javax.servlet` ?

waghanza:
CI is OK, since bugs are on other **frameworks** (solution found)

PR #2014, opened 2019-11-26
Add Aero

akyoto:
As requested in #2007.
Please let me know if there are any troubles.

### DIFF
--- a/go/aero/config.yaml
+++ b/go/aero/config.yaml
@@ -1,3 +1,3 @@
 framework:
-  website: github.com/aerogo/aero
-  version: 1.3.48
+  github: aerogo/aero
+  version: 1.3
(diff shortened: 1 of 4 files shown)


go/aero/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: github.com/aerogo/aero
```
waghanza: rather use the shortcut `github: aerogo/aero`

go/aero/config.yaml
```diff
@@ -0,0 +1,3 @@
...
+  version: 1.3.48
```
waghanza: could you use X.Y format, more readable for user, and less maintenance to do

akyoto:
:+1:

PR #2025, opened 2019-12-01
Update results

waghanza:
Hi,

This `PR` update results since last update.

Regards,


dalisoft:
Hi @waghanza and thanks for benchmark, but this update not updated `nanoexpress-pro` update. After how much time will be result updating? Thanks

PR #2062, opened 2019-12-20
Use SWOOLE_BASE mode

Yurunsoft:


waghanza:
Hi @Yurunsoft,

How does it affect `imi` ?

Regards,

Yurunsoft:
`imi` is based on `Swoole`

`Swoole` has two modes:

`SWOOLE_PROCESS` multiprocess mode

`SWOOLE_BASE` basic mode. Using the `SWOOLE_BASE` mode, the `IPC` overhead can be reduced twice and the performance can be improved

waghanza:
@Yurunsoft Thanks. Very clear. Could you make sure this mode is used on all `swoole` based frameworks, here ?

I have no `php` skills, and comparing various frameworks  is better if the same mode is used ;-)

Yurunsoft:
Swoole is coroutine asynchronous programming framework for PHP.

Enable PHP developers to write high-performance, scalable, concurrent TCP, UDP, Unix socket, HTTP, Websocket services in PHP programming language without too much knowledge about non-blocking I/O programming and low-level Linux kernel.

Not all PHP frameworks are based on swoole.

The two modes have their own advantages and disadvantages. It is up to the developer to decide which mode to use.

Most of the swoole frameworks here choose the SWOOLE_BASE mode

waghanza:
You're right. This could be strange to compare two frameworks based on 2 modes, but I understand.

PR #2067, opened 2019-12-24
[PHP] Update one to 1.9

waghanza:
Hi @lizhichao,

This `PR` is a superset of yours (I can not right on yours, so I create a new one with your work).

I have seen that composer has a unexpected behavior when using `~1.9`.

Contrary to **semver**, it allow to install `1.10`

@see https://jubianchi.github.io/semver-check/#/constraint/~1.9

I've thus checked all `composer.json` here, and make sure that the right constraint is used ;-)

-----------------------------------
Closes #2066


lizhichao:
What can I do?

waghanza:
Nothing, I've just pinged you so you can see that I take you're work :heart: 

PS : It will be better for me if you create your futur `PR` from non-master branch

lizhichao:
ok
If I submit PR next, I will create a new branch submit

PR #2068, opened 2019-12-25
[Python] Introduce pyramid

### LINKED ISSUE #1934: Add  pyramid
Hi @mmerickel @stevepiercy @caseman @mcdonc,

Would you mind to add (or consent that we add) [pyramid](https://github.com/Pylons/pyramid) in `here` ?

Regards,

waghanza:
Hi,

This `PR` adds `pyramid` here.

@mmerickel @stevepiercy @caseman @mcdonc, Could you please take a look at this `PR`  and let me know if there is some `pyramid` best-practices I'd rather use ?

Regards,

----------------------------
Closes #1934


python/pyramid/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: trypyramid.com
```
stevepiercy:
```suggestion
  website: https://trypyramid.com/
```

python/pyramid/server.py
```diff
@@ -0,0 +1,22 @@
+from wsgiref.simple_server import make_server
+from pyramid.config import Configurator
+from pyramid.response import Response
+from pyramid.response import Response
```
stevepiercy: Lines 3 and 4 are duplicates of one another.

```suggestion
```

python/pyramid/server.py
```diff
@@ -0,0 +1,22 @@
...
+
+
+with Configurator() as config:
+    config.add_route("index", "/")
+    config.add_route("create_user", "/user", request_method="POST")
+    config.add_route("get_user", "/user/{id}")
+    config.add_view(empty, route_name="index")
+    config.add_view(empty, route_name="create_user")
+    config.add_view(show, route_name="get_user")
+    app = config.make_wsgi_app()
```
stevepiercy: I'm not sure why you omitted a few lines that typically exist in the simplest Pyramid app, maybe because you use gunicorn for benchmarking? If yours does not run, then try the following.

```python
if __name__ == '__main__':
    with Configurator() as config:
        config.add_route("index", "/")
        config.add_route("create_user", "/user", request_method="POST")
        config.add_route("get_user", "/user/{id}")
        config.add_view(empty, route_name="index")
        config.add_view(empty, route_name="create_user")
        config.add_view(show, route_name="get_user")
        app = config.make_wsgi_app()
    server = make_server('0.0.0.0', 6543, app)
    server.serve_forever()
```

python/pyramid/config.yaml
waghanza: This URL will be displayed in results (I dynamically adds `https://`) so we do not need this `here`

python/pyramid/server.py
waghanza: Yes, I need `gunicorn` to run this, of course I'm open to suggestions :heart: 

I've not seed any **recommended** way in the documentation

stevepiercy: Pyramid has no opinion about the web server you use in front of your app. You will not find a recommended way.

The [Pyramid Community Cookbook offers a suggestion for deploying with gunicorn](https://docs.pylonsproject.org/projects/pyramid_cookbook/en/latest/deployment/gunicorn.html). This is not officially supported, but something the community contributed. I think it is outdated because it suggests using `paste` to start serving a Pyramid app. `pserve` is more common usage.

You will see that your minimalist app uses Python's [`wsgiref.simple_server`](https://docs.python.org/3/library/wsgiref.html#module-wsgiref.simple_server) to serve the app. If you want to use gunicorn, you will need to make some modifications.

The easiest and officially supported way to get a working Pyramid app is to start with the [pyramid-cookiecutter-starter](https://github.com/Pylons/pyramid-cookiecutter-starter) to generate the app, then follow the instructions of the generated `README.txt` to install your app and its dependencies, and to run it. You could compare what is generated from that, and modify your one-file app into a distributable package accordingly.

Our Quick Tutorial includes one step [03: Application Configuration with `.ini` Files](https://docs.pylonsproject.org/projects/pyramid/en/latest/quick_tutorial/ini.html) that demonstrates and explains how this all works, with further references to the documentation.

PR #2108, opened 2020-01-04
[Nim] Update containers to 1.04

waghanza:
Hi,

This `PR` update `nim` to `1.0.4`.

@dom96 @moigagoo Actually **nimlang/nim:1.0** in `docker`

I think it could be good, to have a **1.0** tag that link to the last patched version (`1.0.2`, `1.0.4` ...) in this **major**

Regards,

### DIFF
--- a/nim/Dockerfile
+++ b/nim/Dockerfile
@@ -1,4 +1,4 @@
-FROM nimlang/nim:1.0.2-alpine
+FROM nimlang/nim:1.0.4-alpine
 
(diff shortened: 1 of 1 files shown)


moigagoo:
@waghanza Hi! Why not use `latest` for now? Until 2.0 is released, 1.0 and latest will point to the same version. And 2.0 isn't coming out for a couple of years I think at least.

PR #2125, opened 2020-01-06
[Javascript] Update nanoexpress-pro to 1.9

waghanza:
Hi,

This `PR` update `nanoexpress-pro` from `1.8` to `1.9`

Regards,

@dalisoft : Slight decrease, maybe insignificant

+ `1.8` : `204 040` `requests/s`
+ `1.9` : `201 072` `requests/s`

--------------
Supersed #2120


dalisoft:
@waghanza Thanks for notify. I am did not change anything core-related, just added new feature to plugin which does not used by in this benchmark. I think this because of always different result for RPS (speed)

waghanza:
@dalisoft In deed, when results will be computed on a **rock solid** `infrastructure, their will not be any fluctuation, but know we are only on a **local** `docker` so ... results are not perfect, but could give some ideas

the above change are not very significant

:heart:

PR #2172, opened 2020-01-16
[Python] Aligned Emmett to other benchmarks (using gunicorn)

gi0baro:


waghanza:
Most of `python` **frameworks** `here` are ran with `gunicorn`. I'm questioning if it is not a better idea to use their buit-in appserver (sanic, cyclone, tornado, ... have their own ways).

any tought @gi0baro

gi0baro:
@waghanza Is not such an easy question.

AFAIK, today the vast majority of async python frameworks rely on *uvicorn*, *hypercorn* or *daphne* for the ASGI server implementation. Even frameworks started with their own server implementation at the end switched on one of those options.
Gunicorn only helps to manage processes and workers as an interface to those servers.

If the aim of the benchmark suite you put together is to measure the different frameworks *overhead*, comparing one to each other, then using the same server implementation should – statistically speaking – give better quantifiable results.

On the other side, if the aim of the suite is to benchmark frameworks also considering the implementation for the server part, then it might have sense to use all the peculiar solutions everyone of them provides. In this case, however, it would be harder to match results from the one with their own servers and the ones missing it.

I've added gunicorn to Emmett (which internally use uvicorn) just to take advantage of the number of the cpus involved in the test like the other ones.

waghanza:
@gi0baro in fact, majority of framework use `gunicorn`, **but** with specific workers -> `uvicorn` for **asgi** for example

PR #2176, opened 2020-01-18
turn off sanic debugging and access logging

syfluqs:
this will turn off sanics access and debug logs, as per their documentation, which is supposed to be done in a production environment


waghanza:
@syfluqs Could you provide the link of this documentation ?
The above code is not working
```python
    app = Sanic(debug=False, access_log=False)
TypeError: __init__() got an unexpected keyword argument 'debug'
```

PR #2198, opened 2020-01-26
Update shard.yml

grkek:
Pin the framework since the newer version implements breaking changes.

### DIFF
--- a/crystal/grip/shard.yml
+++ b/crystal/grip/shard.yml
@@ -11,6 +11,6 @@ targets:
 dependencies:
   grip:
     github: grkek/grip
-    version: ~> 0.26.3
+    version: 0.26.5
 
 license: MIT


crystal/grip/shard.yml
```diff
@@ -11,6 +11,6 @@ targets:
 dependencies:
   grip:
     github: grkek/grip
```
waghanza: Could you please keep the `~>` ?

grkek: With the `~>` other PR's will break as well because the new version is way different than the old Grip.

waghanza: ah, you mean that **breaking changes** will occurs on `0.26.6`
understood

but if I act like that, please make update on each `grip` release

PR #2235, opened 2020-02-04
[Clojure] Use OpenJDK 14

waghanza:
Hi,

This `PR` ensure that OpenJDK 14 is use to
+ build
+ run

@swlkr I could not find any `docker` **image** (for _clojure_) based on **alpine**. I then use the default image (on **debian**) to build and run on **alpine**.

Is this ok to do so ?

Regards,

----------
Closes https://github.com/the-benchmarker/web-frameworks/issues/2160


swlkr:
Yeah it's probably find, the jdk is what matters really

PR #2244, opened 2020-02-07
Update Athena version

Blacksmoke16:
* Update Athena version
* Upgrade Crystal to `0.32.1` via the official alpine docker image
* Support specifying ENV vars in container
* Add db command to clear data of previous runs
* Add `docker-compose.yml` to make running the postgres server easier


waghanza:
Awesome, `crystal` is not shipped on **alpine** and `docker` :+1: 

Good idea to add a `docker-compose`

PR #2252, opened 2020-02-09
Use multiple concurrency

waghanza:
Hi,

As https://www.techempower.com/benchmarks/ does (in _Data Table_ tab) it could be accurate to compute with several concurrency level.

The idea behind that is to identify frameworks that perform differently on various concurrency level.

This is raw result, and we are far away from a ranking.

Regards,


waghanza:
@ray-ang `basicphp` was not working `here`. I've made 2 things two it so as it works again :
+ [x] Rename `url_value` to `url_path`
+ [x] Comment `error404` function

Could you please look and correct me ? The second point is clearly not a **fix**

ray-ang:
Yes, you are right. url_value should be url_path in the controller.


-------- Original message --------

waghanza:
@ray-ang and for `error404` ?

ray-ang:
error404 function is just for error handling. No problem if it gets executed.


-------- Original message --------

waghanza:
@ray-ang no. having it throw 404 in any route

PR #2297, opened 2020-02-25
Build Crystal shards in release mode

Sija:
That should make a difference in benchmarks ;)


Sija:
@waghanza It would be good to update the Crystal-related benchmarks.

PR #2314, opened 2020-02-28
[Go] Upgrade atreugo to v10.4.1

savsgio:


savsgio:
Thanks for your notice @waghanza

PR #2342, opened 2020-03-09
Upgrade workerman to 4.x

walkor:
Upgrade workerman to 4.x


walkor:
image

Here is my command. How can i fix the ci error?

waghanza:
Ah, I see. In order to rebase, you SHOULD not have any local modification. There is two things to do to solve that. Stash save before (and stash pop after) or checkout all files (but this erase your modifications)

PR #2355, opened 2020-03-16
update One to ~2.0

lizhichao:
update One to ~2.0


php/one-fpm/App/index.php
```diff
@@ -4,6 +4,11 @@
 
 define('_APP_PATH_VIEW_', __DIR__ . '/View');
 
+function uuid()
```
waghanza: what is that function for ?

php/one/App/swoole.php
```diff
@@ -7,6 +7,11 @@
 
 define('_APP_PATH_VIEW_', __DIR__ . '/View');
 
+function uuid()
```
waghanza: what is this function for ?

php/one-fpm/App/index.php
lizhichao: By default, the framework sets a unique id for logging. There is no need for logging here. I rewritten it.

php/one/App/swoole.php
lizhichao: By default, the framework sets a unique id for logging. There is no need for logging here. I rewritten it.

waghanza:
@lizhichao could you update files

+ `php/one-fpm/config.yaml`
+ `php/one/config.yaml`

to reflect  the updated `one` version ?

PR #2358, opened 2020-03-16
Remove duplicated instructions

j8r:
I think it was introduced during your rebase in https://github.com/the-benchmarker/web-frameworks/pull/2347.


waghanza:
My bad, I've not seen it 

Thanks for catch up 🎉

Issue #2217, opened 2020-01-31
Can you try using rust's Actix framework


amoghrandhar:
**Actix Framework**: https://actix.rs/

As per [this benchmark](https://www.techempower.com/benchmarks/#section=data-r18) it's titled as the fastest web-server there.

inlet3355:
Previously, the original owner requested to remove from the benchmark:
https://github.com/the-benchmarker/web-frameworks/issues/2093

Actix and Actix-web are now manage by new project leader, @JohnTitor

waghanza:
Thanks for reaching us. If the `actix` community wants to re-activate it `here`, it'll be a pleasure to help :heart:

waghanza:
I close this issue, please feel free to re-open if @JohnTitor or any `actix` community member want to add it `here`

JohnTitor:
Hi all, sorry for the delay!
I received an email and I'm fine to re-add the actix project. If you want my help, just let me know!

waghanza:
Sorry, I've received any notification for your comments ;-)

It will be a pleasure to re-add `actix` here, feel free to `PR` @JohnTitor 

The idea is to have a very basic impl for now :

+ GET /, SHOULD return a successful status code and an empty string
+ GET /user/:id, SHOULD return a successful status code and the id
+ POST /user, SHOULD return a successful status code and an empty string

DestyNova:
Hi! I added a small PR for a basic Actix implementation. I don't know what steps are required for the Dockerfile -- can you advise if I need to do something special with that?

PR #2404, opened 2020-03-26
Add minimal Rust Actix implementation

DestyNova:
Add an implementation for Rust's [Actix framework](https://actix.rs).

I don't know much Rust and just started looking at Actix, but it seems pretty straightforward. The release build takes quite a while (at least, on my old i3 laptop), so I didn't test any of the Dockerfile stuff.

I added the target directories for Rust implementations to `.gitignore`, and am thinking _perhaps_ `Cargo.lock` should go there as well, but people generally keep package lock files in repos nowadays.

Fixes issue #2217.


waghanza:
@DestyNova would you mind if I write in this `PR` ?

rust/actix/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "server"
+version = "0.1.0"
+authors = ["Oisín Mac Fhearaí <denpashogai@gmail.com>"]
+edition = "2018"
+
+[dependencies]
+actix-web = "2.0.0"
```
waghanza: Please use `~`, because it's more flexible

rust/actix/Cargo.toml
```diff
@@ -0,0 +1,15 @@
...
+actix-rt = "1.0.0"
```
waghanza: Please use `*`. Specifying the version of `actix-rt` is the job of `actix-web` in https://github.com/actix/actix-web/blob/206733188478f4d18df376a72fac3a4938b99bd6/Cargo.toml#L66

rust/actix/config.yaml
```diff
@@ -0,0 +1,4 @@
+framework:
+  website: actix.rs
+  version: 2.0.0
```
waghanza: Please use `X.Y` format

rust/actix/src/main.rs
```diff
@@ -0,0 +1,18 @@
...
+
+#[actix_rt::main]
+async fn main() -> std::io::Result<()> {
+    HttpServer::new(|| {
+        App::new()
+            .route("/", web::get().to(|| HttpResponse::Ok()))
+            .route("/user", web::post().to(|| HttpResponse::Ok()))
+            .route("/user/{id}", web::get().to(get_user))
+    })
+    .bind("127.0.0.1:3000")?
```
waghanza: Please use `0.0.0.0`. This app will be built on a container, thus need to be called outside of it

rust/actix/Cargo.toml
JohnTitor: Hm, this specification is common usage and I think it's fine.

waghanza: @JohnTitor but if `actix-rt` version is specified `here`, it requires to update it on each `actix-web` update. Is it not more flexible to use `*` as `actix-web` specify the version of `actix-rt` ?

DestyNova:
Please feel free to add any commits to my PR. I'm doing work stuff now but
will address comments in a couple of hours.

DestyNova:
I've made the changes and will push in a few mins after this clean build completes, just to be sure. I see the CI build failed, maybe due to a timeout. Is there anything I can do to improve that?

DestyNova:
Looking at the CI output, Actix is on track to be the second slowest build at around 9 minutes :sweat_smile: Guess that's just a thing with Rust -- hopefully build times (and size of the dependency directories) will reduce in future. Still, when I tested locally with `wrk` it was pretty fast when finally built.

waghanza:
For the history, `actix` build  time was the reason why I introduce  `travis_wait` on `CI`.

The build time was more than **10** minutes, so ...

Btw, the results are computing right, but seems to be more close to `go` frameworks (~150K `req/s`).

The results are still **NOT** *production-ready`, but I am on it

DestyNova:
> For the history, `actix` build time was the reason why I introduce `travis_wait` on `CI`.
> 
> The build time was more than **10** minutes, so ...

Yeah it's pretty big. I had a quick look at the `.travis.yml` file and didn't see a [cache configuration](https://docs.travis-ci.com/user/caching/#arbitrary-directories). Do you think we could cache the target directories for all the frameworks, or would it not suit the way we're using Docker?

Maybe we could do something like:

1. Specify cache location
```
cache:
  directories:
  - $HOME/build_cache
```
2. Mount the cache directory into the Docker container when starting (e.g. in /tmp/build_cache)
3. Create a directory for the current framework (/tmp/build_cache/rust/actix in this case)
4. Symlink that directory to wherever it's required (i.e. target for the Rust frameworks, but maybe `$HOME/.nimble` for Nim -- different for each language :confused:)
5. Run the rest of the build as normal.

waghanza:
Sadly, this does not work ;-) ...

PR #2415, opened 2020-03-30
Add simps framework

sy-records:


php/simps/sbin/simps.php
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env php
+<?php
+
+ini_set('display_errors', 'on');
+ini_set('display_startup_errors', 'on');
+
+error_reporting(E_ALL);
+date_default_timezone_set('Asia/Shanghai');
```
waghanza: Is this required ?

php/simps/swoole.ini
```diff
@@ -0,0 +1,3 @@
+[swoole]
```
waghanza: Could you remove this file ?
It's not needed

php/simps/sbin/simps.php
sy-records: Not necessary, I will delete him later

php/simps/swoole.ini
sy-records: I need to turn off enable_library because the composer of `swoole/library` conflicts with the underlying swoole.
Or I modified the command to `php -d swoole.enable_library=off sbin/simps.php http: start`?

sy-records: Oh, i see `before_command`, Let me delete it

sy-records:
Thanks.

PR #2429, opened 2020-04-02
Update composer.json

sy-records:
* `swoole/ibrary` is not required

### DIFF
--- a/php/simps/composer.json
+++ b/php/simps/composer.json
@@ -6,8 +6,5 @@
     "psr-4": {
       "App\\": "app/"
     }
-  },
-  "config": {
-    "optimize-autoloader": true
   }
 }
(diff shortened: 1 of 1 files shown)


php/simps/composer.json
```diff
@@ -1,11 +1,13 @@
...
-    "swoole/library": "dev-master"
+    "simple-swoole/simps": "dev-master"
   },
   "autoload": {
     "psr-4": {
       "App\\": "app/"
     }
+  },
+  "config": {
+    "optimize-autoloader": true
```
waghanza, line 11: what does it mean ?

is it related to `composer dumpautoload -o`

sy-records: Yes, it is. Is there anything wrong?

sy-records: I saw it in the dockerfile, let me delete it

PR #2439, opened 2020-04-06
[PHP] Modify configuration for simps

sy-records:
@waghanza Hello, can you review it? If there is no problem, can you update the [results list](https://github.com/the-benchmarker/web-frameworks#results)?


sy-records:
> I'm not sure, this can be merged, afaik. The purpose of this project is to measure some indicators of frameworks, but using the closest **entreprise-grade** implementations.
> 
> I am not sure this implementation (not having a separated `routes.php`) could be use in such an environment

I don’t think there is any problem. He can be used in businesses with simple requirements. For example, I have a file system, I don’t need to login, there is no database, I just parse the corresponding address and read the makedown file.

sy-records:
The definition of this framework is simple and lightweight, I think this is better, although it also supports the use of `routes.php`

waghanza:
Ah, I see ... `simps` could be used as a `proxy` or something else ...

Ok.

I have to categorize **frameworks** in `here`, I'll contact you when I'll start this sub project :heart:

sy-records:
Well, okay, can I resubmit a pr? For the master code now, modify some configuration?

waghanza:
I do not understand ...

I'll check this later

sy-records:
I mean, if this pr is not merged for the time being, can I submit another pr to modify the configuration? Continue to use `route.php`.
Thank you.

waghanza:
I think I mis-lead you.

In fact I don't mind how this frameworks intend to be used (having an external `route` or else), the idea is to have a **production-grade** implementation.

So feel free to open any `PR` you want. The only thing to know, is that I prefer one implementation per framework (or some what I called variants, for example symfony/nginx vs symfony/swoole)

sy-records:
Haha. Actually I think the framework should be more free, let users decide how to use it, and choose the appropriate production-grade implementation according to their needs.

sy-records:
I leave some things to the users to handle, let the users make their own decisions, rather than the framework for various encapsulation.

So I didn’t understand why you said this

> I have to categorize frameworks in here, I'll contact you when I'll start this sub project

waghanza:
@sy-records By categorizing I mean label the framework as full-stack, micro, routeur only ...

sy-records:
Oh, I see.

PR #2485, opened 2020-04-15
[PHP] Upgrade Simps to 1.0.0

sy-records:


sy-records:
Thanks for reminding.

php/simps/composer.json
```diff
@@ -1,6 +1,6 @@
 {
   "require": {
-    "simple-swoole/simps": "dev-master"
+    "simple-swoole/simps": "^1.0"
```
waghanza, line 3: I prefer using `~`. Using carret will break logic, it could install **1.1**, but generate a `README` with a wrong information :stuck_out_tongue:

waghanza: only the case if using `~1.0.0`

php/simps/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   website: simps.io
-  version: preview
+  version: 1.0.0
```
waghanza, line 3: Could you please specify `X.Y` format ?

PR #2486, opened 2020-04-15
[PHP] Upgrade hyperf to 1.1.*

limingxinleo:


php/hyperf/.gitignore
```diff
@@ -0,0 +1,3 @@
+vendor
```
waghanza, line 1: could you please add this to `php/.gitignore` ? I prefer to have one `.gitignore` per **language**

php/hyperf/.php_cs
```diff
@@ -0,0 +1,89 @@
+<?php
```
waghanza, line 1: could you please remove this file ?

I prefer to have the same conventions on all php frameworks, or at least if you fine it useful to create an issue to adopt community style for all php frameworks

PR #2508, opened 2020-04-18
[PHP] Add ice 1.5.1

### LINKED ISSUE #2504: Add ice
Hi @mruz ,

Would you mind to add (or consent that we add) [ice](https://www.iceframework.org/
) in `here` ?

Regards,

mruz:
Closes #2504


php/ice/config.yaml
```diff
@@ -0,0 +1,23 @@
+framework:
+  website: iceframework.org
+  version: 1.5.1
```
waghanza, line 3: Could you use `X.Y` format here ?

php/ice/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+build_deps:
+  - bash
+
+before_build:
+  - wget -O /tmp/framework-1.5.1.tar.gz https://github.com/ice/framework/archive/1.5.1.tar.gz
```
waghanza, line 9: could ice be installed with `pecl` or some other tool ?

php/ice/config.yaml
```diff
@@ -0,0 +1,23 @@
...
+  - cd /tmp/ && tar -xzf framework-1.5.1.tar.gz && cd framework-1.5.1 && bash install
+  - cp /tmp/ framework-1.5.1/tests/ci/ice.ini /usr/local/etc/php/conf.d/docker-php-ext-ice.ini
```
waghanza, line 11: is this config file tuned for CI ? could it be tuned for a real-word app ?

php/ice/config.yaml
mruz: There is a packege, but only for the openSUSE linux https://software.opensuse.org/package/php7-ice, so it's needed to compile.

php/ice/config.yaml
mruz: It's just to turn the extension on:
```ini
extension=ice.so
```

waghanza: Rather use an `echo extension=ice.so` then. 

I find it more explicit

php/ice/config.yaml
waghanza: In fact, I thought about a `pecl`, so as to have a tool doing compilation stuff fo us.
However, it is ok for this project. Do you have considered to distribute`ice` with `pecl` (it coul be easier when this project will run on cloud)

mruz: Agree. I will try to add to the pecl repository.

php/ice/public/index.php
```diff
@@ -0,0 +1,19 @@
...
+
+$di->loader
+    ->addNamespace('App', __DIR__ . '/../App')
+    ->register();
+
+$di->dispatcher
+    ->setNamespace('App');
+
+$di->router
+    ->setRoutes([
```
waghanza, line 13: @mruz a route is missing : `['GET', '/']`

mruz: https://github.com/the-benchmarker/web-frameworks/blob/684ac4a6088f640d11820dda79b5d1bfa1cc3c21/php/ice/public/index.php#L14
It's defined in the optional action `[/{action}]`, so not needed to define it again. Default action `index` will be used.

waghanza: Strange then I've a server error on while requesting this endpoint

waghanza: Could you rebase and push to trigger`CI` ?

waghanza: Strange, travis do not start

mruz: Screenshot_20200419_183556
I'm not sure why, but it says "Abuse detected" for all my commits https://github.com/travis-ci/travis-ci/issues/8626

waghanza: I've seen that, I sent an email to travis' support. I'll probably create a `PR` in this repository, and include your commits :stuck_out_tongue:

waghanza: this abuse was detected on my side, I donot know more, yet

PR #2516, opened 2020-04-19
[Go] Add tango

### LINKED ISSUE #2506: Add tango
Hi @lunny ,

Would you mind to add (or consent that we add) [tango](https://gitea.com/lunny/tango
) in `here` ?

Regards,

waghanza:
Hi @lunny,

This `PR` add `tango`. 

Could you please check this code, and give me your advices :heart: 

Purpose is to have **production-grade** implementations `here` so we can compare (performances) on factual basis.

Regards,

------------------------------------

Closes #2506

### DIFF
--- a/go/tango/main.go
+++ b/go/tango/main.go
@@ -23,7 +23,12 @@ func (Empty) Post() string {
 }
 
 func main() {
-	t := tango.Classic()
+	t := tango.New(
+		tango.Logging(),
+		tango.Recovery(false),
+		tango.Return(),
+		tango.Param(),
+	)
 	t.Get("/", new(Empty))
 	t.Get("/user/:id", new(Params))
 	t.Post("/user", new(Empty))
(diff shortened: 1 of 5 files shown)


go/tango/main.go
```diff
@@ -0,0 +1,31 @@
...
+func (Empty) Get() string {
+	return ""
+}
+
+func (Empty) Post() string {
+	return ""
+}
+
+func main() {
+	t := tango.Classic()
```
lunny, line 26: I suggest to use 
```
t := tango.New(
		tango.Logging(),
		tango.Recovery(false),
		tango.Return(),
		tango.Param(),
	)
```

And the default port is 8000.

lunny:
And please mark as `language:golang` but not `language:php`.

waghanza:
> And please mark as language:golang but not language:php.

Oops

go/tango/main.go
waghanza: sure :tada: 

However, the port is set by an environment variable

PR #2559, opened 2020-04-22
Scala and akka-http bumped

mfirry:
Scala bumped to 2.13.1;
akka-http bumped to 10.1.11;
meanwhile:
sbt bumped to 1.3.10;
sbt-assembly bumped to 0.14.10;


scala/akkahttp/build.sbt
```diff
@@ -2,9 +2,9 @@ import sbtassembly.AssemblyPlugin.defaultShellScript
 
 name := "AkkaHttp"
 
-version := "0.0.1"
+version := "0.0.2"
 
-scalaVersion := "2.12.2"
+scalaVersion := "2.13.1"
```
waghanza, line 7: is their a way to set a more relax version of `scala` ? I mean using a range like `[2.13,2.14)` in order to be sure to stick to **2.13.x** but use the most up-to-date version in this minor

scala/akkahttp/build.sbt
```diff
@@ -14,6 +14,7 @@ assemblyOption in assembly := (assemblyOption in assembly).value
 assemblyJarName in assembly := "server_scala_akkahttp"
 
 libraryDependencies ++= Seq(
-  "com.typesafe.akka" %% "akka-http" % "[10.0,10.1)",
```
waghanza, line 17: could we use `[10.1,10.2)` in order to stick to **10.1** but have last up-to-date patch version ?

scala/akkahttp/build.sbt
```diff
@@ -14,6 +14,7 @@ assemblyOption in assembly := (assemblyOption in assembly).value
...
-  "com.typesafe.akka" %% "akka-http-core" % "[10.0,10.1)"
+  "com.typesafe.akka" %% "akka-http" % "10.1.11",
```
waghanza, line 17: could we use `[10.1,10.2)` in order to stick to **10.1** but have last up-to-date patch version ?

scala/akkahttp/build.sbt
```diff
@@ -14,6 +14,7 @@ assemblyOption in assembly := (assemblyOption in assembly).value
...
+  // "com.typesafe.akka" %% "akka-http-core" % "10.1.11"
+  "com.typesafe.akka" %% "akka-stream" % "2.6.4"
```
waghanza, line 19: is this safe to specify the version _any_ `here`
I mean, for each release of `akka-stream` we should bump if we specify a version `here`

scala/akkahttp/project/assembly.sbt
```diff
@@ -1 +1 @@
-addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.4")
+addSbtPlugin("com.eed3si9n" % "sbt-assembly" % "0.14.10")
```
waghanza, line 1: could we use the last version (if it is safe) ?

scala/akkahttp/build.sbt
mfirry: Not that I know of.

scala/akkahttp/build.sbt
mfirry: I don't understand what you mean here. Akka Http latest version is 10.1.11. Before that, 10.1.10.

scala/akkahttp/build.sbt
mfirry: You don't _need_ to upgrade for each new version of akka-stream. But you can. The choice (for akka-http) was made that you **need** to specify the version of akka-stream you want to use.

scala/akkahttp/project/assembly.sbt
mfirry: 0.14.10 is the latest as far as I can see

scala/akkahttp/build.sbt
waghanza: I mean that using **10.1.11** specify that we want to use a particular version.
Using `[10.1,10.2)` range will use that last version in **10.1** (as I understand)

mfirry: I'd rather have a fixed version so that with every build you're sure you're testing the same thing. But it's your call. you could put `[10.1,10.2)` if you want... or better `[10.1.11,10.2)`... it should work okay.

waghanza: the thing is :
+ I have to stick to the version displayed in results (for convenience reason I display `X.Y`)
+ I want the less maintenance works
in consequence of what, I prefer `[10.1.11,10.2)` (this is also the case for `java` and `kotlin`)

mfirry: Ok

mfirry:
I'm trying to run this locally but it's not working: here's the output https://gist.github.com/mfirry/726ca285358dfa1efda53a5d1653a580

any suggestion?

waghanza:
Which is your OS ? Seems that `xargs` is not the same as me ^^

mfirry:
macOS Mojave

waghanza:
ah, I'll try on a mac then :heart: I am on [fedora](https://fedoraproject.org) and it was convenient for me to use the internals (`bash` funtions)

mfirry:
Ok. Please let me know since I want to test some frameworks locally.

PR #2560, opened 2020-04-22
Update server.cr

grkek:


waghanza:
This mean that each method must return `response` ?

grkek:
Correct!

PR #2562, opened 2020-04-22
Rails jruby truffleruby

firodj:
1. add ruby for rails framework using jruby engine
2. add ruby for rails framework using truffleruby (graalvm) engine
3. Change ruby Dockerfile, move the order of COPY . to prevent redownload Gemfile if Gemfile is not changed.


firodj:
got it 👍

PR #2571, opened 2020-04-23
add nestjs with express/fastify for benchmarker

zieglar:
Closes [#954](https://github.com/the-benchmarker/web-frameworks/issues/954#issuecomment-520960790)


javascript/nestjs-fastify/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "dependencies": {
+    "@nestjs/common": "^6.10.14",
```
waghanza, line 3: could you use `~6.10.14` instead of `^6.10.14` ?

zieglar:
> I assume some files are build by `nest`. How are they generated ?
> 
> It is better to generate them directly onto the container

if want generate them directly onto the container, javascript Dockerfile must rebuild :(

waghanza:
Yep, this is the idea.

But I do now know which `nest` command to use

zieglar:
> Yep, this is the idea.
> 
> But I do now know which `nest` command to use

If you want, I can submit source code.
step list:
1. run `npm -g install @nestjs/cli`
2. run `nest build`
3. copy dist/[*.ts,*.js,*.map] to container 
4. copy package.json to container
5. run `npm install`
6. pm2 start app.js

waghanza:
> If you want, I can submit source code.

It will be awesome

waghanza:
I will create an other `PR` based on your code, to give you my corrections after that :stuck_out_tongue:

zieglar:
> I will create an other `PR` based on your code, to give you my corrections after that 😛

done

waghanza:
I have an error while running `nest build`
~~~javascript
node_modules/@nestjs/common/interfaces/external/mqtt-options.interface.d.ts:104:39 - error TS2580: Cannot find name 'Buffer'. Do you need to install type definitions for node? Try `npm i @types/node`.

104     ca?: string | string[] | Buffer | Buffer[];
                                          ~~~~~~
~~~
This, however, seems to be a warning, but fails docker creation

zieglar:
> I have an error while running `nest build`

maybe run `npm install` before run `nest build` ?

waghanza:
That's what I do ...

zieglar:
> That's what I do ...

I update `package.json` and now nest build is ok

waghanza:
@zieglar results has been updated

PR #2591, opened 2020-04-26
[Crystal] Update lucky to 0.21

waghanza:
/cc @paulcsmith


crystal/lucky/config/log.cr
```diff
@@ -0,0 +1,3 @@
+log = Log::IOBackend.new
+log.formatter = Dexter::JSONLogFormatter.proc
+Log.dexter.configure(:info, log)
```
paulcsmith, line 3: This should probably be :none so that it does not log anything

waghanza: I am a little bit nervous about shipping an app, that does not log anything. The idea is to have a `production-grade` application @paulcsmith

paulcsmith: I think it is common practice not to log in benchmarks. I believe all the other crystal frameworks also have removed logging. But if you feel strongly then this can be left in 👍

waghanza: I think a better idea is to enable it `here`, and enable (what is the common usage for **production**) logging on all others
wdyt ?

paulcsmith: I think it is pretty common to disable logging and I'd guess not all of the other frameworks log by default.

If every single framework has logging then I'd agree, but if not, then I think it is more reliable to disable them all. Kind of "all-or-nothing". Otherwise the results will be skewed

waghanza: Agree. I'll check and open a `PR` if needed

Issue #2491, opened 2020-04-16
Add Siler


RicardoSette:
Hi @waghanza,

add Siler in tests

+ Language : PHP
+ Repo : https://github.com/leocavalcante/siler
+ Author : @leocavalcante 
+ Fork : https://github.com/RicardoSette/web-frameworks/tree/master/php/siler

RicardoSette:
Hi @waghanza, 

Here #2598

Issue #2490, opened 2020-04-15
Add Yii2+Swoole


RicardoSette:
Hi @waghanza,

add Yii2+Swoole in tests

+ Language : PHP
+ Repo : https://github.com/yiisoft/yii2
+ Repo : https://github.com/swoole-foundation/yii2-swoole-extension
+ Author : @qiangxue , @samdark and @xialeistudio

+ Fork : https://github.com/RicardoSette/web-frameworks/tree/yii2-swoole/php/yii-swoole

RicardoSette:
Hi @waghanza, 

Here #2598

PR #2598, opened 2020-04-26
Add Siler, Siler-Swoole and Yii-Swoole

RicardoSette:
Hi @waghanza,

from #2491 , #2492 and #2490


php/siler-swoole/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+      "leocavalcante/siler": "dev-master"
```
waghanza, line 3: could you use `~1.7.2` instead of `dev-master` ?

php/siler/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+  "require": {
+    "leocavalcante/siler": "dev-master"
```
waghanza, line 3: could you use `~1.7.2` instead of `dev-master` ?

php/siler-swoole/config.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+  github: leocavalcante/siler
```
waghanza, line 2: this line could be removed (having `website` give the same kind of info)

php/siler/config.yaml
```diff
@@ -0,0 +1,4 @@
+framework:
+  github: leocavalcante/siler
```
waghanza, line 2: this line could be removed (having `website` give the same kind of info)

php/siler-swoole/composer.json
RicardoSette: @leocavalcante it is ok?

leocavalcante: @RicardoSette, no problem. The latest stable release supports features used in the test.

waghanza: @RicardoSette @leocavalcante version will be in-sync :heart: Thanks to `dependabot` and `renovate`

leocavalcante: @waghanza how cool! Just added to Siler itself, thanks for the hint.

PR #2627, opened 2020-05-03
Use dynamic concurrency level

waghanza:
Hi,

As I have 8 cores, sieging uses 64,256,512,1024,2048 concurrency level.

Those level are based on my number of cores (`8^2`, `8^2*4`, `8^2*8`,`8^2*16`,`8^2*32`).

This `PR` fix `db` tool to get for this database the accurate concurrency level (it was previously **hard-coded**).

Regards,

### DIFF
--- a/tools/src/db.cr
+++ b/tools/src/db.cr
@@ -23,6 +23,12 @@ end
 class App < Admiral::Command
   class ReadmeWriter < Admiral::Command
     def run
+      factor = System.cpu_count**2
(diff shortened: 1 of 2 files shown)


whiplash:
Not the best at reviewing this change, I guess. I've tried locally to make these changes and it worked for me, for what it's worth.

waghanza:
Glad this is working for you, I'll merge after CI is ok

PR #2640, opened 2020-05-04
Added Nette framework

lulco:


php/nette/composer.json
```diff
@@ -0,0 +1,10 @@
+{
+    "require": {
+        "nette/application": "^3.0",
```
waghanza, line 3: Could you use `~3.0.0`  ? We need to stick to **3.0**, the `^` notation could change the major

php/nette/composer.json
```diff
@@ -0,0 +1,10 @@
...
+        "nette/bootstrap": "^3.0",
```
waghanza, line 4: Same as above

php/nette/composer.json
```diff
@@ -0,0 +1,10 @@
...
+        "nette/di": "^3.0"
```
waghanza, line 5: Same as above

php/nette/composer.json
```diff
@@ -0,0 +1,10 @@
...
+    },
+    "autoload": {
```
waghanza, line 7: What is this for ?

waghanza:
@dg Are you Ok (at least not opposed) to include `nette` in this list ?

dg:
> @dg Are you Ok (at least not opposed) to include `nette` in this list ?

Yes, of course :)

php/nette/composer.json
lulco: sure, I will change

php/nette/composer.json
lulco: this is for loading all classes in `app` directory using composer, I am pretty sure that application will not work without it

waghanza: ah, it is not related to `composer autoload` (which is done in `Dockerfile`)

it was probably a noob question, I am more a `rubyist` that a `php` adopter

lulco:
Thanks for merge! How often results are updated?

PR #2643, opened 2020-05-05
Clojure.luminus

### LINKED ISSUE #1391: Introduce luminus
+ Language : clojure
+ Repo : https://github.com/luminus-framework/luminus-template
+ Author : @yogthos

whiplash:
Closes #1391 #2645


waghanza:
Could you take a quick look @yogthos, and give us your feedback ?

We want to have a **production-grade** `luminus` app :heart:

yogthos:
Hi,

Just looked over the commit, and the app looks reasonable to me. The default template is tuned to be production ready out of the box. When the app is built using `lein uberjar`, it will produce a standalone runnable jar that can be used for benchmarking.

whiplash:
Hi @yogthos and thanks. Just one question. I wanted to take out the layout part, since it's not really needed here. What do you say? Also `nrepl` isn't needed, is it?

yogthos:
Some files could be trimmed, but they shouldn't affect runtime performance. If the benchmark gets expanded later, these might be useful for additional functionality.

whiplash:
Thanks.

PR #2657, opened 2020-05-07
[Pony] Use static compilation

waghanza:
Hi,

It convenient for maintenance tasks that `pony` compiles in one single binary.

It avoid having to setup third-party dependencies.

Regards,

### DIFF
--- a/pony/Dockerfile
+++ b/pony/Dockerfile
@@ -1,7 +1,6 @@
 FROM ponylang/ponyc:0.35.1
 
-RUN apt-get update
-RUN apt-get install -y clang libssl-dev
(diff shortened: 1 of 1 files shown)


waghanza:
/cc @firodj

@Theodus is there any way to avoid `libatomic` dependency ?

Theodus:
Not on an OS that uses glibc. It's a requirement for the Pony runtime

waghanza:
ah ok, I thought `pony` create fully static binaries :stuck_out_tongue: 

however, I compile / run on alpine, but `apk add libatomic libgcc libexecinfo` are required

Issue #2622, opened 2020-05-01
Homogenize logging feature


waghanza:
Hi,

Some framework are made to disable logging. 

To ensure fairness (at least maximum we can do), we need to have a **production-level** logging level.

For example, `lucky` is configured to handle `info` logs.

Regards,

Blacksmoke16:
Reiterating my thoughts from #2915:

I can see the point of wanting to provide benchmarks that would match what a user would see when they go to use it.  However my concern is it would bias the results towards frameworks that have less logging in the first place.  The only sensible & fair solution is to disable logging across the board.

[TechEmpower](https://www.techempower.com/benchmarks/)'s guidelines take this stance:

> All test implementations must disable all disk logging. For many reasons, we expect all tests will run without writing logs to disk. Most importantly, the volume of requests is sufficiently high to fill up disks even with only a single line written to disk per request. Please disable all forms of disk logging. We recommend but do not require disabling console logging as well.

Given they say **We recommend but do not require disabling console logging as well**, I can't imagine any framework not doing that.

> For example, `lucky` is configured to handle `info` logs.

That might explain why it's the 4th slowest framework in the list 😉.

waghanza:
To summarize, usually on **production** app, we **SHOULD NOT** disable logging.

However, your main point @Blacksmoke16 is that logging **SHOULD** be disabled `here`, because logging consist in disk operations are in tank the results.

So,
+ in one hand, we have the idea to run as-production
+ in the other hand, we have the idea that we can not use **logging** as production (it mess-up results)

After reflexion, @Blacksmoke16, I think we **SHOULD** disable **logging**. However, we **SHOULD** adds this information clearly.

Disable logging should in all frameworks. @Blacksmoke16, could you open a `PR` to disable it on all `crystal` frameworks ? I'll take care of the other. The idea is to disable logging for all, in order not to be "unfair"

PR #2685, opened 2020-05-14
Change worker from non-optimized worker in Sanic

ahopkins:
I know I have brought this up before (see #1253) that using `gunicorn` is not an optimal strategy for testing Sanic. 

And, while it is true that Sanic does have a gunicorn worker, it is there for compatibility and not performance.

Seeing that there are a number of ASGI frameworks that are being tested using the `uvicorn.workers.UvicornWorker`, I think it is fitting that Sanic also is tested against this. On my machine, and with my `wrk` settings, I saw a 300% increase in req/s.


waghanza:
Hi @ahopkins,

Welcome, the idea is to have a **production-grade** implementation for each framework. I admit, in the past, I thought using `gunicorn` was a way to avoid apple Vs banana benchmark.

However, to be fair in comparison, it appears that we have more advantages to compare what technologies are made for, so not to try to have the same tool (appserver in this case).

ahopkins:
I think that makes a lot of sense. Sanic is both a framework and a server. The best example of its implementation would probably by to run it by itself. However, for comparison against other ASGI, perhaps the best way to run it is with `uvicorn` or `hypercorn`. 

Which leads to another question, should the benchmarks offer multiple deployment options to allow for proper review?

waghanza:
Nop, not yet @ahopkins.

BTW, `uvicorn` is not present (in `requirements.txt`). 

Actually, I prefer to use the simplest version (`uvicorn` is :ok: ), and I'll add variants (https://github.com/the-benchmarker/web-frameworks/pull/1570#issuecomment-547308791) but I don't know when (it depends of my personal schedule)

waghanza:
What is the recommended way to deploy `sanic` ? In other terms, how are the majority of **production** deployed ?

ahopkins:
It is difficult to say with certainty since we do not keep statistics on that. But, from anecdotal evidence (meaning by reviewing the questions people ask), I would say that by far the most common deployment strategy is to use the Sanic server itself.

```python
app = Sanic()
if __name__ == "__main__":
    app.run()
```

and 

```
$ python server.py
```

waghanza:
So, be it :tada:

I think, we should then implement `sanic` this way

ahopkins:
Nice. I will adjust the PR and resubmit tonight.

ahopkins:
Using AMD Ryzen 5 2600 Six-Core Processor

```
gunicorn  \
  --log-level warning  \
  --bind 0.0.0.0:3000  \
  --reuse-port  \
  --workers $(nproc)  \
  --worker-class sanic.worker.GunicornWorker \
    server:app

Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency    12.65ms   11.37ms 147.61ms   78.97%
    Req/Sec     3.05k     1.12k    7.30k    69.72%
  1092465 requests in 30.07s, 122.94MB read
Requests/sec:  36331.74
Transfer/sec:      4.09MB



uvicorn server:app --workers=12 --port=3000 --no-access-log 

Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency     7.55ms    7.86ms 132.64ms   87.92%
    Req/Sec     5.54k     1.72k   17.85k    75.11%
  1985957 requests in 30.10s, 251.90MB read
Requests/sec:  65980.10
Transfer/sec:      8.37MB



gunicorn  \
  --log-level warning  \
  --bind 0.0.0.0:3000  \
  --reuse-port  \
  --workers $(nproc)  \
  --worker-class uvicorn.workers.UvicornWorker \
    server:app

Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency     6.91ms    6.97ms 123.28ms   88.14%
    Req/Sec     5.94k     1.58k   22.36k    72.83%
  2127624 requests in 30.09s, 269.87MB read
Requests/sec:  70699.19
Transfer/sec:      8.97MB



python server.py

Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency     5.43ms    6.26ms 131.58ms   87.47%
    Req/Sec     8.37k     2.46k   25.15k    76.60%
  2995908 requests in 30.10s, 337.14MB read
Requests/sec:  99532.11
Transfer/sec:     11.20MB
```

python/sanic/server.py
```diff
@@ -17,3 +21,11 @@ async def user_info(request, id):
 @app.route("/user", methods=["POST"])
 async def user(request):
     return text("")
+
+if __name__ == "__main__":
+    try:
```
waghanza, line 26: why a try ? I thing the line #29 without the try could be :ok_hand: for anythin in **production**

python/sanic/server.py
```diff
@@ -17,3 +21,11 @@ async def user_info(request, id):
...
+        workers = int(sys.argv[1])
+    except IndexError:
+        workers = multiprocessing.cpu_count()
+    app.run(host='0.0.0.0', port=3000, workers=workers,
+            debug=False, access_log=False)
```
waghanza, line 31: why disabling access logs ?

python/sanic/server.py
ahopkins: Was there only for a catch if number of processors was not passed. Want me to remove it?

waghanza: In fact I'm not sure handling a failure case of a `core` **python** is our scope.
Yeah, I pref to remove it

python/sanic/server.py
ahopkins: According to my review, this is the way other frameworks were being benchmarked. In particular, I noted this across all of the ASGI frameworks (which all should benchmark fairly close to the same `70000` that Sanic does in ASGI mode), and also Flask (which is around 50k). I did not dig into the Django benchmark, that's a whole different beast.

And, I think that sort of makes sense. Writing to `stdout` in Python is terribly slow, and is not a good practice.

In my experience on Sanic projects, and in general, a **much** better solution would be to put `nginx` as a proxy in front of your server, and handle access logs there.

Indeed, adding `access_log=True` changes the results **considerably**.

```
Running 30s test @ http://127.0.0.1:3000
  12 threads and 400 connections
    Latency    13.17ms   12.34ms 157.35ms   83.24%
    Req/Sec     2.99k     1.00k   11.33k    73.22%
  1071363 requests in 30.08s, 120.56MB read
Requests/sec:  35620.79
Transfer/sec:      4.01MB
```

More importantly, however, I think there needs to be some guidelines and transparency about this because at least for Python, that makes a **huge** difference, and is rarely implemented in production. I am a big advocate for allowing tools like `nginx` and `apache` do what they do best (including logging).

waghanza: > this is the way other frameworks were being benchmarked

Yes, but **this** is wrong

Correct me if I'm wrong, but access log should **NOT** be disabled. I have use some `python` **framework** in **production**, but _never_ disabled logs

ahopkins: This is your benchmark, so it is your call. But, to me this project is useful to see how these frameworks compare with one another in an attempt at objectivity. Access logs have no use here for several reasons:

1. It is an absurd degradation [to just write to `stdout`](https://stackoverflow.com/questions/3857052/why-is-printing-to-stdout-so-slow-can-it-be-sped-up). This should **not** be done in production. It is terrible for performance, yes, but it is also drain on system resources because of the lack of buffering. I don't know many production systems logging like this. If they are, they shouldn't be.
1. It clouds the evaluations of the frameworks.
1. If you wanted a "production" ready implementation, I would highly suggest running some builds inside containers so that there would be multiple potential layers, load balancing, etc. To me this is much too complicated and brings the utility of this benchmarking down since it vastly increases the variables involved. 
1. If you were really looking for a production ready build, the apps need to be much more complicated. There are **VAST** differences that can be introduced [based upon the complexity of the routing](https://github.com/vltr/very-stupid-simple-routers-bench). Your tests do not account for this, or other network calls, which therefore does not approximate production usage. This is **NOT** a fault. I am not critiquing. I think the simplistic approach makes sense for what you are trying to achieve. With that in mind, using sub-optimal, non-production worthy logging sort of breaks that mold.

waghanza: why are you saying that log are just writing to _stdout_ ? I don't know the internal of `gunicorn` (for example), so I can't say.

However, when you deploy an application in **production**, it is a `non-sens` for me to run an application in **production** without logging (not speaking about logging level).

For me this benchmark is useful to see how technologies can react (performances, but other insights will come) in the real-word.

I agree with your 4 points, but :
1. I mean logging, no matter if it is `stdout` (generally it is not)
2. Yes, writing to `stdout` could mess result, but why writing to `stdout` ?
3. In reflection ... it's very complex, but totally feasible (I admin, I'm an optimistic guy :heart: )
4. In reflection ... this project will not be only a benchmark, but also a tool benchmarking a more complex implementation (not made by me, perhaps https://github.com/gothinkster/realworld)

Finally, I had to say that there is plenty of benchmark all over the internet. The very value of this project is to (mainly) reply to 2 cases :
+ I have to choose a stack for a new project
+ I want to improve my framework

ahopkins: Like I said, your project, your goals. I am not here to tell you how to do it. Just merely pointing out that if you want to have a fair shake at representation of various strategies, then you need **much** more complex setups for each project, which probably includes involving many more maintainers. Having looked through your setups (for the Python frameworks at least), they are still far too simplistic to be realistic.

If you are going for "realism", why not have more complex apps with more routes? I think this would be very helpful to see how frameworks handle complex routing, which in my experience has a major impact on performance.

Also, any change that you could set the project up to handle containers? The reason I ask is that I think the **best** (in an overly generalized way) and most practical deployment strategy for any of the Python frameworks would be to sit them behind `nginx`.

ahopkins: Also, if the goal is to help people decide what to choose, shouldn't you also have some factor for showing popularity, download rates, code update frequency, etc?

waghanza: > If you are going for "realism", why not have more complex apps with more routes?

Think of it ...

>  sit them behind nginx

I'll also produce some results, with nginx alongside bjoern ...

> showing popularity, download rates, code update frequency, etc?

Yes, in a second time. I have to keep focus on results right now

Could logging be enable for sanic ? It use stdout or logging python module ?

ahopkins: Standard python module. By default the config is set to output to st out. But I will push another update to use the null logger so it is doing the logging.

ahopkins: I spent some more time looking through this... I really think that access logs have no place here. Most of the other projects that you are testing do **not** even offer them, and from what I can see, you have everything set to `WARNING`. Which means that there would not even be any access logs.

I would suggest merging this PR with a similar logging config as to the other projects, and then coming back to this question afterwards and doing a single PR to add consistent logging across all tests. This seems to be the best way to get "fair" results. Right now Sanic is listed at the bottom of the Python frameworks, and it should be listed as one of the faster Python frameworks.

waghanza: > By default the config is set to output to st out.

I has not aware of that. How, then log are used in **production** (`rsyslog` or an other strategy) ?

> I would suggest merging this PR with a similar logging config as to the other projects, and then coming back to this question afterwards and doing a single PR to add consistent logging across all tests. 

Absolutely, the logging issue **SHOULD** not block this `PR` :heart:

python/sanic/server.py
```diff
@@ -17,3 +67,8 @@ async def user_info(request, id):
 @app.route("/user", methods=["POST"])
 async def user(request):
     return text("")
+
+if __name__ == "__main__":
+    workers = int(sys.argv[1])
```
waghanza, line 72: `sys.argv[1]` will never be set, I prefer to let `sanic` spread across all cores

python/sanic/server.py
ahopkins: > How, then log are used in production (rsyslog or an other strategy) ?

I can only speak anecdotally since I have never done a survey of the issue. However, I suspect the most normal setup is to use `nginx` in front and log to syslog or something like that. Perhaps I am mistaken.

But, the Sanic docs do sort of say that the access logging that is enabled by default is not performant. It is there to help users get up and running fast. The goal of the project is not only performance, but also fast development.

python/sanic/server.py
ahopkins: This is to allow passing in $(nproc) in the command line:

```
python server.py $(nproc)
```

I opted for this over `multiprocessing.cpu_count()` since this might not be reliable depending upon the hardware you are using.

ahopkins:
Oops... I hit resolve by accident. Which would you rather rely upon, nproc or pythons multiprocessing module?

waghanza:
> nproc or pythons multiprocessing

I prefer multiprocessing, since `nproc` has to be evaluated (by `shell` on deployment)

python/sanic/server.py
waghanza: Ok. That's relevant. We can then investigate / decide in https://github.com/the-benchmarker/web-frameworks/issues/2622

PR #2687, opened 2020-05-15
haskell.servant framework added.

whiplash:


waghanza:
Hi @phadej @jkarni,

Would you consent (or not opposed to) add https://github.com/haskell-servant/servant in `here` ?

Regards,

haskell/servant/servant-proj.cabal
```diff
@@ -0,0 +1,24 @@
...
+  build-depends:       base >= 4.7 && < 5
+                     , servant-server
+                     , wai
+                     , warp
+  default-language:    Haskell2010
+
+executable servant-proj-exe
+  hs-source-dirs:      app
+  main-is:             Main.hs
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N
```
jkarni, line 21: Either here or as a CLI option to `stack` in the Dockerfile, I'd add the `-O2` option (I *think* by default stack doesn't pass them). Probably also a few other flags (`-A64m` for example, or ` -funbox-strict-fields`) will help a lot. But usually figuring out the exact parameters requires testing in a machine similar to the one you'll ultimately be using for the benchmark.

haskell/servant/servant-proj.cabal
```diff
@@ -0,0 +1,24 @@
...
+  build-depends:       base >= 4.7 && < 5
+                     , servant-server
+                     , wai
+                     , warp
+  default-language:    Haskell2010
+
+executable servant-proj-exe
+  hs-source-dirs:      app
+  main-is:             Main.hs
+  ghc-options:         -threaded -rtsopts -with-rtsopts=-N -with-rtsopts=-A64M -O2 -funbox-strict-fields
```
waghanza, line 21: Could this options be directly passed to `stack build` @whiplash ?

whiplash: I think so. But why do you think it's better? `stack build --ghc-options="-O2"` should work

waghanza: I prefer to set those options on `Dockerfile`, in order to have the same compilation option, if other `haskell` **frameworks** comes :stuck_out_tongue_closed_eyes:

haskell/servant/servant-proj.cabal
```diff
@@ -0,0 +1,24 @@
...
+-- description:
+category:            Web
+build-type:          Simple
+cabal-version:       >=1.10
+
+library
+  hs-source-dirs:      src
+  exposed-modules:     Lib
+  build-depends:       base >= 4.7 && < 5
+                     , servant-server
```
waghanza, line 13: I think it could be better to use `servant-server>= 0.18 && < 0.17` to stick to `0.17` @whiplash

whiplash: honestly not sure how stack/cabal handle (semantic?) versioning.

waghanza: I'm sure @jkarni has the answer, but seems that `>= 0.17 && < 0.18` ensure we stick to **0.17**

Refering to https://www.haskell.org/cabal/release/cabal-1.14.0/doc/users-guide/developing-packages.html#package-properties

PR #2715, opened 2020-05-26
Upgrade webgo to v3.4.2

bnkamalesh:

PR #2719, opened 2020-05-27
fix: remove indirect tag from webgo

rarkins:


waghanza:
Thanks @rarkins, I've already done in https://github.com/the-benchmarker/web-frameworks/pull/2715

PR #2832, opened 2020-06-01
go: Add macaron test code

humaidq:
Hello,

As suggested (go-macaron/macaron#203), I tried to add test code for [macaron](https://go-macaron.com).
I didn't test the code (don't have time to setup the database/docker).

Thank you.


humaidq:
> However, is this implementation suitable for **production** usage, I mean it's important that we can compare **framework** on real usage.

The implementation is the same how you would use it in production, I implemented it according to the routes in your tests. Macaron is used in projects such as Gitea/Gogs, so the framework is suitable for production, if that's what you are asking.

waghanza:
I meant that some implementations `here` are tuned for benchmarks, and I prefer to have some implementations tuned for `production` ❤️

unknwon:
Thanks @humaidq !

PR #2857, opened 2020-06-04
[Nim] Add akane

waghanza:
Hi @Ethosa,

Following https://github.com/Ethosa/akane/issues/1, this `PR` adds `akane` in here.

Thanks for your help :tada:

However, is there any way to :

+ send `plain/text` instead of **json** or **html**
+ avoid having `<head><meta charset='utf-8'></head>` while `response.answer`

Also, I have `SIGSEGV: Illegal storage access. (Attempt to read from nil?)` for routes `/user/:id` and `/user`

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -54,6 +54,7 @@ env:
       - FRAMEWORK=c.kore
       - FRAMEWORK=c.agoo-c
       - FRAMEWORK=perl.dancer2
+      - FRAMEWORK=nim.akane
       - FRAMEWORK=nim.jester
       - FRAMEWORK=nim.whip
       - FRAMEWORK=nim.rosencrantz
--- a/nim/Dockerfile
+++ b/nim/Dockerfile
@@ -4,7 +4,11 @@ WORKDIR /usr/src/app
 
 COPY server.nim server.nimble ./
 
-RUN nimble c \
+{{#deps}}
+  RUN apk --no-cache add {{{.}}}
+{{/deps}}
+
+RUN nimble c {{#build_opts}} {{{.}}} {{/build_opts}} \
     -d:danger \
     -d:noSignalHandler \
     --gc:markAndSweep \
--- a/nim/akane/config.yaml
+++ b/nim/akane/config.yaml
@@ -1,7 +1,3 @@
 framework:
-  github: dom96/jester
-  version: 0.4
-
-files:
-  - /usr/src/app/server
-  
+  github: Ethosa/akane
+  version: 0.1
--- a/nim/akane/config.yaml
+++ b/nim/akane/config.yaml
@@ -1,0 +1,7 @@
+framework:
+  github: dom96/jester
+  version: 0.4
+
+files:
+  - /usr/src/app/server
+  
--- a/nim/akane/config.yaml
+++ b/nim/akane/config.yaml
@@ -1,3 +1,8 @@
 framework:
   github: Ethosa/akane
   version: 0.1
+
+deps:
+  - pcre-dev
+
+build_opts: '--define: usePcreHeader --passL: "/usr/lib/libpcre.a"'
--- a/nim/akane/server.nim
+++ b/nim/akane/server.nim
@@ -1,0 +1,28 @@
+import akane
+
+
+proc main {.gcsafe.} =
+  let server = newServer("0.0.0.0", 3000)
+
+  server.pages:
+    "/": 
+      if request.reqMethod == HttpGet:
+        await request.answer("")
+      else:
+        await request.error("not GET :(")
+
+    "/user":
+      if request.reqMethod == HttpGet:
+        await request.answer("")
+      else:
+        await request.error("not GET :(")
+
+    regex(re"\A/user/id(\d+)\Z"):
+      if request.reqMethod == HttpPost:
+        await request.sendJson(%*{"id": url[0]})
+      else:
+        await request.error("not POST :(")
+
+  server.start()
+
+main()
--- a/nim/akane/server.nim
+++ b/nim/akane/server.nim
@@ -7,21 +7,21 @@ proc main {.gcsafe.} =
   server.pages:
     "/": 
       if request.reqMethod == HttpGet:
-        await request.answer("")
+        await request.send("")
       else:
         await request.error("not GET :(")
 
     "/user":
-      if request.reqMethod == HttpGet:
-        await request.answer("")
-      else:
-        await request.error("not GET :(")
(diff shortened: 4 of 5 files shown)


Ethosa:
I added the macros `sendPlainText` and `send`.
https://github.com/Ethosa/akane/blob/master/tests/test8/main.nim
image
image
what about the `attempt to read from nil` error, which version of Nim was used?

waghanza:
I have `Error: attempting to call undeclared routine: 'sendPlaintext'` with a **nimble** file
```nim
# requires "akane >= 0.1.1 & < 0.2"
requires "https://github.com/Ethosa/akane.git"
```

Ethosa:
The bug was fixed, the current version is 0.1.3
https://github.com/Ethosa/akane/releases/tag/v0.1.3

waghanza:
It's OK for `sendPlainText`, but I still have `SIGSEGV: Illegal storage access.` for `await request.sendPlainText(url[0])`


I'm on **nim** `1.2.0`

waghanza:
and  I have 
```
could not load: libpcre.so(.3|.1|)
compile with -d:nimDebugDlOpen for more information
```
when try to launch on **alpine**

waghanza:
@Ethosa seems that :
+ any `POST` method leads to a `SIGSEGV: Illegal storage access. (Attempt to read from nil?)`
+ `sendPlainText` leads to the above message, but I replace it with `send`

Ethosa:
I didn’t understand where the `Attempt to read from nil` error came from. Everything works on Windows 7 x64.
The error `could not load: libpcre.so(.3|.1 |)` on **alpine** occurred due to the fact that there, apparently, there was no PCRE library, on the basis of which the standard module was implemented [`re`](https://nim-lang.org/docs/re.html).

waghanza:
Seems that I have this (`SIGSEGV: Illegal storage access.`) on all `POST` methods, either on `linux` or on `osx`

(Also folks have the same on nil discord channel)

I do not know why this not occurs on windows

PS : I'm using **1.2.0**

Willyboar:
@waghanza and @Ethosa this code works for me:

```nim 
import akane

proc main {.gcsafe.} =
  let server = newServer("0.0.0.0", 3000)

  server.pages:
    "/": 
      if request.reqMethod == HttpGet:
        await request.send("")
      else:
        await request.error("not GET :(")

    "/user":
      if request.reqMethod == HttpPost:
        await request.send("")
      else:
        await request.error("not POST :(")

    regex(re"\A/user/(\d+)\Z"):
      if request.reqMethod == HttpGet:
        await request.send(url[0])
      else:
        await request.error("method not allowed")

    notfound:
      await request.error("not found")

  server.start()

main()

```

waghanza:
I have always a `SIGSEGV: Illegal storage access. (Attempt to read from nil?)` on `POST`
compile from **master** without options and into `nimlang/nim:1.2.0`

waghanza:
@Ethosa @Willyboar specify a content-length -> `curl http://0.0.0.0:3000/user -X POST -H 'Content-Length: 0' -v` there is not issue

the issue seems to be in async dispatch, the full stack
```
   Success: Execution finished
Traceback (most recent call last)
/home/waghanza/workspace/benchmark/web/nim/akane/server.nim(30) server
/home/waghanza/workspace/benchmark/web/nim/akane/server.nim(28) main
/usr/share/nim/lib/pure/asyncdispatch.nim(1886) waitFor
/usr/share/nim/lib/pure/asyncdispatch.nim(1576) poll
/usr/share/nim/lib/pure/asyncdispatch.nim(1340) runOnce
/usr/share/nim/lib/pure/asyncdispatch.nim(210) processPendingCallbacks
/usr/share/nim/lib/pure/asyncmacro.nim(37) processRequestNimAsyncContinue
/usr/share/nim/lib/pure/asynchttpserver.nim(254) processRequestIter
/usr/share/nim/lib/pure/asynchttpserver.nim(103) respond
/usr/share/nim/lib/pure/httpcore.nim(180) hasKey
SIGSEGV: Illegal storage access. (Attempt to read from nil?)
```
leads me to thing this was in `asynchttpserver.nim` in `stdlib`

waghanza:
:fire: :warning: we have to have for `nim` next release to merge it

waghanza:
Merge is blocked by #2989

waghanza:
We are almost done

2 issues :

+ `alpine` required `libprce`  even if I Have install `pcre-dev`

```
/usr/src/app # ./server
could not load: libpcre.so(.3|.1|)
compile with -d:nimDebugDlOpen for more information
```

/cc @Willyboar @Ethosa 

+ returning a `411 Content-Length required.`

/cc @OvermindDL1 is the content-Length required ?

OvermindDL1:
> is the content-Length required ?

Depends on a few factors such as the call type and the http version, but for, say, GET requests it's (basically) always required unless you are sending the data chunked, or else the client wouldn't know when the body would stop.

waghanza:
Thanks @OvermindDL1 🎉 
In fact the question is more for `POST`, `akane` is the **only** framework** that return a `411` on `POST` (we are the default http version, I think `1.1`)

OvermindDL1:
POST is also a call type that returns a body, so it should have a content-length as well or be a chunked response.

waghanza:
Ah yes of course. It should have a `Content-Length` in the response :ok_hand: 

However, with `akane` a **request** ` curl -v http://0.0.0.0:3000/user -X POST` return a **411** (specifying the `Content-Length` header fix this)

OvermindDL1:
Personally sounds like either a bug in akane that's not filling it in automatically based on the body length, or it should be documented that it must be user supplied?

waghanza:
I also find this weird @OvermindDL1, let's wait for @Ethosa

waghanza:
solved with https://forum.nim-lang.org/t/2236#14054

I understand this could fix `could not load: libpcre.so(.3|.1|)` on **alpine**, but I don't understand how it could impact the `Content-Length` issue

Ethosa:
> Ah yes of course. It should have a `Content-Length` in the response 
> 
> However, with `akane` a **request** ` curl -v http://0.0.0.0:3000/user -X POST` return a **411** (specifying the `Content-Length` header fix this)



I updated `akane` to version 0.1.5 and added a release, maybe now everything will work. :eyes:

> Personally sounds like either a bug in akane that's not filling it in automatically based on the body length, or it should be documented that it must be user supplied?

Also documented the macros `sendJson`, `send`, `error`, `answer` and `sendPlainText`.

PR #2897, opened 2020-06-07
[F#] Add Saturn 🪐

### LINKED ISSUE #2886: [F#] Add Saturn
Hi @Krzysztof-Cieslak,

Would you mind to add (or consent that we add) [Saturn](https://github.com/SaturnFramework/Saturn) in `here` ?

Regards,

Krzysztof-Cieslak:
Fix #2886


fsharp/saturn/web.fsproj
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Program.fs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Saturn" Version="0.13.1" />
```
waghanza, line 12: better us `0.13.*` to stick to **0.13**

Krzysztof-Cieslak:
> Can I make you code owners of this implementation ?

Sure, yes. And yes, please feel free to make any changes that are necessary.

waghanza:
@panesofglass The version ranges were not correctly specified. According to  https://docs.microsoft.com/en-us/nuget/concepts/package-versioning#version-ranges we have to use `maven` range style

@Krzysztof-Cieslak could I add you to `CODEOWNER` for `saturn` ? so as you can be notified on each change according it ?

Krzysztof-Cieslak:
> @Krzysztof-Cieslak could I add you to CODEOWNER for saturn ? so as you can be notified on each change according it ?

Yes, sounds good 👍

Krzysztof-Cieslak:
@waghanza, can you wait a bit with merging this - I'll probably release new Saturn version today or tomorrow, so we'll need to rerun tests and update results

waghanza:
sure, I'll convert as draft then (so as I remember not to merge)

feel free to undraft when ready

baronfel:
@waghanza a quick note on nuget version ranges: the nuget client resolves to the lowest matching version when given a range like has been done he re for these web libraries, so as new versions are released your benchmarks will stay pinned to the lowest version. In actual .net applications you typically pin exact versions for this reason. Libraries are free to specify version ranges so that they are more broadly compatible with different versions of dependencies without forcing a consuming application to use specific versions.

Krzysztof-Cieslak:
Ok, I've updated Saturn version, we can rerun the tests. Because of the reason mentioned above by @baronfel I updated the range in Saturn project to use `0.13.2` minimal.

waghanza:
Thanks for update @baronfel. This was unknow behavior, but with helps `here` we will change this.

I've raised an issue on https://github.com/renovatebot/renovate/issues/6487 for `renovatebot`, since it is use to take care of up-to-date versions.

fsharp/saturn/web.fsproj
```diff
@@ -0,0 +1,14 @@
...
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Include="Program.fs" />
+  </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Saturn" Version="[0.13.2,0.14)" />
```
waghanza, line 12: @Krzysztof-Cieslak this will to `0.13.2` no matter what, we can use `0.13.*`

baronfel:
Some egg on my face here, my previous comment was wrong about the behavior of floating ranges.  Per https://docs.microsoft.com/en-us/nuget/concepts/dependency-resolution#floating-versions, for floating references specifically (in PackageReference elements) nuget will float to the latest available version that the star provides.  So then these ranges should not be the explicit bounds, but rather the *, eg.`0.13.*`

baronfel:
This is mostly confusing because of the conflict with legacy nuget behavior, apologies all around.

waghanza:
@Krzysztof-Cieslak ready for merging ?

Krzysztof-Cieslak:
@waghanza, yeah I think from my side everything is ready. One quick question - it seems that README update with new results was done before Saturn update to `0.13.2` - shouldn't it be updated again with new results?

waghanza:
Yes, I'll try to update in this `PR` :heart:

PR #2900, opened 2020-06-08
[V] Add valval

waghanza:
Hi @taojy123

Thanks for you work at https://github.com/the-benchmarker/web-frameworks/issues/2892 :tada: 

Is there a way I can handle `/user/:id`, I mean extract **0** in **/user/0**

Regards,

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -182,6 +182,7 @@ env:
       - FRAMEWORK=go.fasthttprouter
       - FRAMEWORK=go.rte
       - FRAMEWORK=go.atreugo
+      - FRAMEWORK=v.valval
 
 before_script:
   - psql -c 'CREATE DATABASE benchmark;' -U postgres;
--- a/v/Dockerfile
+++ b/v/Dockerfile
@@ -1,4 +1,4 @@
-FROM taojy123/vlang:v0.1.24
+FROM gladmo/vlang:0.1.27-alpine
 
 WORKDIR /usr/src/app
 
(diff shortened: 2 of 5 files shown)


taojy123:
Hi @waghanza 
Sorry，valval does not support `path params`  yet, so I use `query params` instead.

waghanza:
A damn ... We could not merge it `here` tho.

Any plan to support that @taojy123  ?

taojy123:
hi, @waghanza 
Need to add `COPY . ./` to Dockerfile?
And, you install valval by `v install`, so you should replace `valval` with `watchmen123456.valval` in mian.v
such as:
```v
module main

import watchmen123456.valval

fn index(req watchmen123456.valval.Request) watchmen123456.valval.Response {
	return watchmen123456.valval.response_ok('')
}

fn user(req watchmen123456.valval.Request) watchmen123456.valval.Response {
	if req.method == 'POST' {
		return watchmen123456.valval.response_ok('')
	}
	id := req.get('id', '-1')
	return watchmen123456.valval.response_ok(id)
}

fn main() {
	mut app := watchmen123456.valval.new_app(true)
	app.route('/', index)
	app.route('/user', user)
	watchmen123456.valval.runserver(app, 3000)
}
```
(This code style for v-language 0.1.24 is so ugly.)

waghanza:
Could you update the code base to `0.1.27` ?

taojy123:
0.1.27 has some problem. so I just can support for 0.1.24

waghanza:
> 0.1.27 has some problem. so I just can support for 0.1.24

which problems @taojy123 ?

taojy123:
such as v install xxx  can not install package successful. 
and so on

waghanza:
let's wait then. The idea is to maintain `here` the last version of each framework and languages :heart:

taojy123:
Ok, I will try to fix the compatibility issues with the latest version.

PR #2902, opened 2020-06-08
[V] Add vape

exastencil:


v/vape/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  github: exastencil/vape
+  version: 0.3.0
```
waghanza, line 3: could you use `X.Y. format ?

exastencil:
This failure is definitely due to an old version of V. I'll see if I can find or create a newer one, preferably on alpine. And I'll change the version format while I'm at it.

v/Dockerfile
```diff
@@ -0,0 +1,19 @@
+FROM taojy123/vlang:v0.1.24
```
waghanza, line 1: we can exceptionally use `:latest` @exastencil

waghanza: or use https://hub.docker.com/layers/gladmo/vlang/0.1.27-alpine/images/sha256-db6ca5e8ed213c1639d3cb42a0979b707f68c55fde926a0e869168cb6368a524?context=explore

exastencil:
Currently I get an error when installing the framework:

```
Step 3/9 : RUN v install exastencil.vape
 ---> Running in d8e76df3cd82
[Error]  -- STDERR(.neph/v.vape/log/log.err) -- 
OCI runtime create failed: container_linux.go:346: starting container process caused "process_linux.go:319: getting the final child's pid from pipe caused \"read init-p: connection reset by peer\"": unknown
```

v/vape/main.v
```diff
@@ -0,0 +1,24 @@
+import exastencil.vape
+
+mut server := vape.Server{port: 3000}
```
waghanza, line 3: what does `mut` mean ?

exastencil: It makes the server instance mutable. Variables are immutable by default in V. The server instance needs to be mutable since registering handlers adds to the router inside the struct.

exastencil: Normally you wouldn't have to write that code. Vape includes a helper `vape dev` that generates an executable for you. But to simplify the code here, I just generated the build source myself so the command doesn't need to be installed and executed.

waghanza:
which `OS` do you use @exastencil. I do not have issue on `linux`

| 1 | crystal (0.34)| [kemal](https://kemalcr.com) (0.26) | 150 147 | 154 411 | 154 224 |
| 2 | go (1.14)| [fiber](https://gofiber.io) (1.10) | 149 772 | 166 832 | 166 175 |
| 3 | v (0.1)| [vape](https://github.com/exastencil/vape) (0.3) | 2 948 | 2 003 | 1 803 |

exastencil:
I'm on macOS at the moment.

waghanza:
You must use `docker-machine` then, which version are you using ?

exastencil:
The latest one from Homebrew.

```
➜  web-frameworks git:(v/vape) docker-machine -v
docker-machine version 0.16.2, build bd45ab1
```

waghanza:
I have the same version, but I can built it. Honestly, I'm not so skilled with `macos` but
```
docker-machine rm default --force && docker-machine create default && eval $(docker-machine env default) && export FRAMEWORK=v.vape && shards build && bin/make config --keep --without-sieger --driver docker-machine && bin/neph ${FRAMEWORK} && crystal spec
```
is successful for testing purposes. 

We had to use `docker-machine` on non `linux` oses since docker client is only available in `linux` 

However, the command 
```
export DATABASE_URL=postgresql://postgres@localhost/benchmark && dropdb -U postgres benchmark && createdb -U postgres benchmark && psql -U postgres benchmark < .ci/dump.sql&& docker-machine rm default --force && docker-machine create default && eval $(docker-machine env default) && export FRAMEWORK=v.vape && shards build && bin/make config --driver docker-machine && bin/neph ${FRAMEWORK} --seq
```
is not working. I suspect `vape` is failing at some high concurrency level

exastencil:
I get the same error. However, I do have some values in my database.

```SQL
SELECT frameworks.label AS framework, concurrencies.level AS concurrency, values.value AS rate 
FROM metrics
JOIN values ON metrics.value_id = values.id
JOIN frameworks ON metrics.framework_id = frameworks.id
JOIN concurrencies ON metrics.concurrency_id = concurrencies.id
```

framework | concurrency | rate
-----------|--------------|----
vape | 16	| 572,71
vape | 64 | 169,47
vape | 128 | 2218,71
vape | 16 | 977,88

However, I should point out that vape is still proof of concept and doesn't yet deal with concurrency, so it doesn't surprise me that increasing concurrency decreases performance. I think `valval` is the only framework currently doing any performance work.

There is also vweb (the web framework in the v standard library) which has been worked on recently.

Do you need to park this merge request until the concurrency situation changes?

waghanza:
I have to investigate further. I have the same issue when running a higher concurrency, but only on `macos` (I mean on `linux` everything is fine), maybe a glitch between `docker-machine`, `docker` and the **v** server.

exastencil:
The goal with `vape` is not actually to have a traditional web server, but to slice your application into individual handlers, similar to functions used in serverless frameworks.

Then these handlers will be routed and bin-packed by a platform running on auto-scaled servers. While concurrency will allow more throughput to each handler, in a way that can also be counter-productive as individual handlers can occupy more server resources.

Eventually, though I will have concurrent handlers and rely on the bin-packing and load balancing to avoid overuse.

Since it's working, I'm happy to have it added and you can hide it for now. But if you'd rather not run it repeatedly, or skew the results, I'm happy to redo it later once the framework is more mature.

waghanza:
> The goal with vape is not actually to have a traditional web server, but to slice your application into individual handlers, similar to functions used in serverless frameworks.

Yeah, some **frameworks** use the same approach `here`. Afaic, this project is more about backend/server-side/whatsoever you called. Thus having your approach is not contradictory to what this project is about, and we are not speaking of traditional approach (any idea is new and could promoted `here`).

If you thing `vape` is 👌, I'm happy to include it `here`
+ If for you having a very low rank is **NOT** a problem, having it displayed by default is :ok:
+ If you think it show **NOT** be displayed by default (because of low rank), it's :ok:

Anyway, any **framework** (homogeneous set of software functions) have its place `here`. The call is to the author/community, it could be counter-productive to include in case of low rank

Should I include it, and show results by default ?

waghanza:
For me `vapie` is **working** :
+ on `macos`
```
docker-machine rm default --force && docker-machine create default && eval $(docker-machine env default) && export FRAMEWORK=v.vape && shards build && bin/make config --keep --without-sieger --driver docker-machine && bin/neph ${FRAMEWORK} && crystal spec
```
+ on `linux`

exastencil:
I'm happy for it to be added and shown, even if it is at the bottom. The Readme in the Vape repo explains the goal, so people can read that if they are interested in it.

👍

PR #2830, opened 2020-06-01
[Swift] Use environment variables for vapor

waghanza:
Hi,

This `PR` adds environment variables usage for https://github.com/vapor/vapor

It facilitate maintenance / homogenization for all frameworks.

Regards,

PR #2919, opened 2020-06-10
[Swift] Use Environment Variables for Vapor

0xTim:
Replaces #2830 

edit
--------
Closes #2830


swift/vapor-framework/Package.resolved
```diff
@@ -114,8 +114,8 @@
         "repositoryURL": "https://github.com/vapor/vapor.git",
```
waghanza, line 114: could you remove this file ?

swift/vapor-framework/Sources/server/main.swift
```diff
@@ -1,6 +1,6 @@
 import Vapor
 
-var env = Environment.production
+var env = Environment(name: Environment.get("VAPOR_ENV") ?? "production")
```
waghanza, line 3: should not it be **dev** by default ? I mean the environment variable will always be set `here`, but a common usage is to set to **dev** environment (at least in `ruby`)

swift/vapor-framework/Package.resolved
0xTim: Done!

swift/vapor-framework/Sources/server/main.swift
0xTim: Ah fair enough, done as well

waghanza:
last thing, could you remove **trailing space** in `config.yaml`
https://www.codefactor.io/repository/github/the-benchmarker/web-frameworks/pull/2919

0xTim:
@waghanza think that's done!

waghanza:
still missing one https://www.codefactor.io/repository/github/the-benchmarker/web-frameworks/pull/2919

PR #2920, opened 2020-06-10
Modify configuration information

lizhichao:
Modify configuration information


php/one/App/Config/protocol.php
```diff
@@ -16,7 +16,8 @@
         'sock_type'   => SWOOLE_SOCK_TCP,
         'ip'          => '0.0.0.0',
         'set'         => [
-            'worker_num'       => swoole_cpu_num(),
+            'enable_coroutine' => false,
+            'worker_num' => (int) shell_exec('nproc') ?? 32,
```
waghanza, line 20: :fire: you assume you are on `linux`

php/one/App/Config/protocol.php
```diff
@@ -16,7 +16,8 @@
         'sock_type'   => SWOOLE_SOCK_TCP,
         'ip'          => '0.0.0.0',
         'set'         => [
-            'worker_num'       => swoole_cpu_num(),
+            'enable_coroutine' => false,
```
waghanza, line 19: could you add this options to all swoole based frameworks `here`

sy-records:
@waghanza The base configuration I recommend is this.
```php
'worker_num'       => swoole_cpu_num(),
'enable_coroutine' => false,
```

@leocavalcante The role of `enable_coroutine` may not be clearly described in the English documentation, so I'll roughly describe it here.

The default value of `enable_coroutine` is `true`, Take `Swoole\Http\Server` as an example.

* When `enable_coroutine` is set to `true`, Swoole automatically creates a coroutine in the `onRequest` callback.
* Conversely, no coroutine is automatically created

> I think the coroutine is useful in an api which includes I/O operation. For example, querying database or calling another api.

He's right, no IO operations such as databases are used in this project, so there is no need to use coroutine.

leocavalcante:
@sy-records thank you very much for the explanation.

huangzhhui:
> enable_coroutine

There is no need to unify this configuration value, different values can be set for different scenarios

waghanza:
@huangzhhui to ensure results accuracy, and a _as fair as_ possible, comparison we **SHOULD** have the same configuration.

I'm in favour of what is used in production, even if it decrease performances (i.e -> enable_coroutine = true since most applications in production use a database)

The purpose is to have results that can be close (as close as possible) the real life

huangzhhui:
> @huangzhhui to ensure results accuracy, and a _as fair as_ possible, comparison we **SHOULD** have the same configuration.
> 
> I'm in favour of what is used in production, even if it decrease performances (i.e -> enable_coroutine = true since most applications in production use a database)
> 
> The purpose is to have results that can be close (as close as possible) the real life

Agree, but actually, the results of this project does not including the tests has I/O requests, so this configuration will not affect the performance results, it does not matter what value it is.

waghanza:
> Agree, but actually, the results of this project does not including the tests has I/O requests, so this configuration will not affect the performance results, it does not matter what value it is.

:ok_hand: so I suggest to leave the default.

Why ?

The idea, as you mention, is that this option (on the non default) could affect only IO operations. So no variation due to that `here`. :heart: 

However, leaving the default option has 3 advantages :
+ Add only code that is necessary (very important for `lean` code)
+ Could suite to any kind of scenario (more `flexible` for us)
+ Anyone (a beginner mostly) could take this code base as `inspiration`

Please vote :+1: to use the **default** and :-1: to set **false**

waghanza:
I leave this issue open since 19th of June, and I will act according to the vote

waghanza:
@leocavalcante the vote in in https://github.com/the-benchmarker/web-frameworks/pull/2920#issuecomment-642449138

waghanza:
@Yurunsoft `enable_coroutine` swoole option seems to be unavailable with **imi**. I have `404` if I add `enable_coroutine => false` in **mainServer.configs**

PR #2944, opened 2020-06-12
chubbyphp-roadrunner

dominikzogg:
chubbyphp-roadrunner is same as chubbyphp but using roadrunner instead of php-fpm.

### DIFF
--- a/php/chubbyphp-roadrunner/config.yaml
+++ b/php/chubbyphp-roadrunner/config.yaml
@@ -1,0 +1,8 @@
+framework:
+  github: chubbyphp/chubbyphp-framework
+  version: 2.8
+
+before_command:
+  - vendor/bin/rr get-binary
+
+command: ./rr serve -v -d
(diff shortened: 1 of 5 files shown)


php/chubbyphp-roadrunner/config.yaml
```diff
@@ -0,0 +1,8 @@
+framework:
+  github: chubbyphp/chubbyphp-framework
+  version: 2.8
+
+before_command:
+  - vendor/bin/rr get-binary
```
waghanza, line 6: @dominikzogg How can the command be different from `spiral` ?
https://github.com/the-benchmarker/web-frameworks/blob/cb5bd67bddd75db632828bb72f79d90d5347a013/php/spiral/config.yaml#L9

BTW, this is working, I have just curiosity in mind

dominikzogg: @waghanza spiral got some wrapper code, i use the plain roadrunner as suggested within the roadrunner documentation

waghanza: :ok_hand:

waghanza:
@dominikzogg I have updated results

PR #2951, opened 2020-06-13
[Kotlin] Update ktor to 1.3

waghanza:
Hi,

This `PR` updates https://github.com/ktorio/ktor to **1.3**

Moreover, I've seen that all `kotlin` frameworks use `org.jetbrains.kotlin:kotlin-stdlib-jdk8`. Is not recommended to use `jdk8` to compile and run then ?

Regards,


kotlin/ktor/gradle.properties
```diff
@@ -1,4 +1,4 @@
-ktor_version=1.2.+
+ktor_version=1.3.+
```
NikkyAI, line 1: would not use + here as that could lead to a ktor version for eap versions of kotlin being used
EAP versions and normal releases are binary incompatible

same for the kotlin version

waghanza: ah thanks for clarification.  🎉 The problem is `+`. How can I specify specify a range saying that I want any version in **1.3** then ?

waghanza: is there a way to avoid `eap` ? (In the case of a comparison, I don't think this is required)

NikkyAI: when you are fully specifying the ktor and kotlin version and made sure to remove kotlin-eap from the repositories it should be fine

NikkyAI: for version management i can recommend this though: https://github.com/jmfayard/refreshVersions
it does reduce a lot of the friction of looking up available versions but still allows full control over which version you pick
it only really starts to shine in big projects and multi project things though

waghanza: If I understand correctly, their is no way to specify a version range (dynamic versioning) in `kotlin`, since `1.3.2-1.4M2` is in `1.3.+` (or `[1.3,1.4)`)

It is not like for example `ruby` where I can use `~> 1.3` (to stay in **1.3.x**), since there a way to avoid `alpha` ... (`~>`)

I then use **strict** versioning , `renovatebot` will help to have up-to-date versions

NikkyAI: you still seem to have kotlin-stdlib-jdk8 on a dynamic version

and not sure if thats available in groovy, but when using `build.gradle.kts`
you can write `implementation(kotlin("stdlib"))`, it will use the kotlin version from the plugin
alternatively: `implementation(kotlin("stdlib", kotlinVersion))` assuming kotlinVersion is defined or loaded from `gradle.propeties`

dynamic version ranges at best just make it better because you do not have to look up all the things when you know major and minor
at worst.. it updates without noticing .. (or due to changing repositories) and you start searching and debugging and lookign what messed up your dependencies

waghanza: thanks for your clarifications @NikkyAI. As you can probably understand, I have no `kotlin`  knowledge (and does not intend to us it for any of my daily basis app, so this of your tips are very valuable :heart: 

my purposes are to :
+ use `ktor` **1.3**
+ use `kotlin` **1.3**
+ avoid any maintenance (or at least try to) even if their is a bot helping me
+ use the most used `jvm` in `kotlin` world (actually gradle use **11**, but `build.gradle` require  **8**)

PR #2981, opened 2020-06-16
[C#] Add Simplify.Web #2973

### LINKED ISSUE #2973: [C#] Add Simplify.Web
Hi @i4004,

Would you mind to add (or consent that we add) [Simplify.Web](https://github.com/SimplifyNet/Simplify.Web) in `here` ?

Regards,

panesofglass:
Closes #2973 

NOTE: I'm not familiar with this library, so I'm not sure that my route processing to retrieve the userId is standard or not.

### DIFF
--- a/csharp/simplify.web/Controllers/UserIdController.cs
+++ b/csharp/simplify.web/Controllers/UserIdController.cs
@@ -1,0 +1,15 @@
+﻿using Simplify.Web;
+using Simplify.Web.Attributes;
+
+namespace web.Controllers
+{
+	[Get("/user/{id}")]
+	public class UserIdController : Controller
+	{
+		public override ControllerResponse Invoke()
+		{
+			var content = this.Context.Route.Substring(6);
+			return Content(content);
+		}
+	}
+}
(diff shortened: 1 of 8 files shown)


i4004:
@panesofglass, @waghanza  the standard way is:
```csharp
public override ControllerResponse Invoke() => Content(RouteParameters.id);
```
or
```csharp
[Get("/user/{id:int}")]
public class UserIdController : Controller
{
	public override ControllerResponse Invoke() => Content(((int)RouteParameters.id).ToString());
}
```

csharp/simplify.web/Controllers/UserIdController.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Simplify.Web;
+using Simplify.Web.Attributes;
+
+namespace web.Controllers
+{
+	[Get("/user/{id}")]
+	public class UserIdController : Controller
+	{
+		public override ControllerResponse Invoke() => Content(RouteParameters.id);
```
panesofglass, line 9: @i4004 is this better?

i4004: @panesofglass yes, it is ok.

PR #3005, opened 2020-06-21
[Java] Add struts

### LINKED ISSUE #2976: [Java] Add struts
Hi @lukaszlenart @JCgH4164838Gh792C124B5

Would you mind to add (or consent that we add) [struts](https://struts.apache.org/) in `here` ?

Regards,

lukaszlenart:
Closes #2976

### DIFF
--- a/java/struts2/config.yaml
+++ b/java/struts2/config.yaml
@@ -3,7 +3,12 @@ framework:
   version: 2.5
 
 build:
-  mvn clean package
+  - mvn clean package
 
-command: >
-  mvn jetty:run
+bin_deps:
+  - jetty-runner
+
+files:
+  - target/struts2-1.0.war
+
+command: java  -XX:+UseNUMA -XX:+UseParallelGC -XX:+AggressiveOpts  -jar /usr/share/java/jetty-runner.jar --port 3000 target/struts2-1.0.war
--- a/java/struts2/config.yaml
+++ b/java/struts2/config.yaml
@@ -1,0 +1,6 @@
+framework:
+  website: struts.apache.org
+  version: 2.5.22
+
+command: >
+  mvn jetty:run
--- a/java/struts2/config.yaml
+++ b/java/struts2/config.yaml
@@ -2,5 +2,8 @@ framework:
   website: struts.apache.org
   version: 2.5
 
+build:
+  mvn clean package
+
 command: >
   mvn jetty:run
--- a/java/struts2/config.yaml
+++ b/java/struts2/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: struts.apache.org
-  version: 2.5.22
+  version: 2.5
 
 command: >
   mvn jetty:run
--- a/java/struts2/pom.xml
+++ b/java/struts2/pom.xml
@@ -10,7 +10,6 @@
 
     <properties>
         <struts2.version>2.5.22</struts2.version>
-        <log4j2.version>2.13.3</log4j2.version>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
     </properties>
 
@@ -20,18 +19,6 @@
             <groupId>org.apache.struts</groupId>
             <artifactId>struts2-core</artifactId>
             <version>${struts2.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-api</artifactId>
-            <version>${log4j2.version}</version>
-        </dependency>
-
-        <dependency>
-            <groupId>org.apache.logging.log4j</groupId>
-            <artifactId>log4j-core</artifactId>
-            <version>${log4j2.version}</version>
         </dependency>
 
         <dependency>
--- a/java/struts2/pom.xml
+++ b/java/struts2/pom.xml
@@ -49,20 +49,6 @@
                     <optimize>true</optimize>
                 </configuration>
             </plugin>
-            <plugin>
-                <groupId>org.eclipse.jetty</groupId>
-                <artifactId>jetty-maven-plugin</artifactId>
-                <version>9.4.30.v20200611</version>
-                <configuration>
-                    <stopKey>CTRL+C</stopKey>
-                    <stopPort>8999</stopPort>
-                    <scanIntervalSeconds>0</scanIntervalSeconds>
-                    <webAppSourceDirectory>${basedir}/src/main/webapp/</webAppSourceDirectory>
-                    <webAppConfig>
-                        <descriptor>${basedir}/src/main/webapp/WEB-INF/web.xml</descriptor>
-                    </webAppConfig>
-                </configuration>
-            </plugin>
         </plugins>
     </build>
 </project>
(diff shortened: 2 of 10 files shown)


java/struts2/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  website: struts.apache.org
+  version: 2.5.22
```
waghanza, line 3: could you use `X.Y` format (`2.5`), since it's clear for readers

java/struts2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>org.apache.struts.benchmarker</groupId>
+    <artifactId>struts2</artifactId>
+    <version>1.0-SNAPSHOT</version>
+    <packaging>war</packaging>
+    <name>struts2</name>
+
+    <properties>
+        <struts2.version>2.5.22</struts2.version>
```
waghanza, line 12: could you use an interval (I mean `[2.5,2.6)`)  ?

java/struts2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+        <log4j2.version>2.13.3</log4j2.version>
```
waghanza, line 13: is this required ? I mean to specify a version

java/struts2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
+                <configuration>
+                    <encoding>UTF-8</encoding>
+                    <source>1.8</source>
```
waghanza, line 61: :+1:

java/struts2/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+                    <target>1.8</target>
+                </configuration>
+            </plugin>
+            <plugin>
+                <groupId>org.eclipse.jetty</groupId>
+                <artifactId>jetty-maven-plugin</artifactId>
+                <version>9.4.30.v20200611</version>
```
waghanza, line 68: is this required ? I mean to specify a version ?

waghanza:
Thanks for you contribution :tada:  I have ask you some question to guarantee that all `java` **frameworks** run the same (at least the max possible) way.

PS : I've no `java` skill, so some questions could appears coming from a noob 

PSS : Could you also adds `structs2` to `.travis.yml` ? btw, why `struts2` and not `struts` ?

java/struts2/pom.xml
lukaszlenart: I tried but then Maven uses `2.6-SNAPSHOT` version which is a development version. Being specific in Maven based project is a good approach.

java/struts2/pom.xml
lukaszlenart: I can throw away the whole logging and just use what's provided.

java/struts2/config.yaml
lukaszlenart: 👍

java/struts2/pom.xml
lukaszlenart: Yes

lukaszlenart:
Done, `struts2` is a different framework then `struts1` 

https://en.wikipedia.org/wiki/Apache_Struts_1
https://en.wikipedia.org/wiki/Apache_Struts_2

java/struts2/pom.xml
waghanza: Ah ... I haven't the same problem with other `java` frameworks, but I have with `clojure` and `kotlin`. There is not way to specify a range **BUT** avoid any `development`, `rc` ... versions in **maven** ?

java/struts2/pom.xml
waghanza: yeah, it will be awesome

The logging will be **disabled** for all frameworks

java/struts2/pom.xml
lukaszlenart: It supposed to ignore `-SNAPSHOT` versions but somehow Maven includes it, I was testing locally and still the same.

```
Resolution of dependency ranges should not resolve to a snapshot (development version) unless it is included as an explicit boundary.
```

lukaszlenart: `mvn clean package`

```
dev-1:[1.8] ~/P/O/w/j/struts2 (master) > ll target/struts2-1.0/WEB-INF/lib
total 10936
-rw-r--r--+ 1 lukaszlenart  staff    72446 Apr 10 07:38 commons-fileupload-1.4.jar
-rw-r--r--+ 1 lukaszlenart  staff   214788 Apr 10 07:38 commons-io-2.6.jar
-rw-r--r--+ 1 lukaszlenart  staff   523372 Apr 19 10:23 commons-lang3-3.10.jar
-rw-r--r--+ 1 lukaszlenart  staff   207030 Apr 10 07:38 commons-text-1.8.jar
-rw-r--r--+ 1 lukaszlenart  staff  1702975 Apr 19 10:23 freemarker-2.3.30.jar
-rw-r--r--+ 1 lukaszlenart  staff   777727 Apr 10 07:39 javassist-3.24.1-GA.jar
-rw-r--r--+ 1 lukaszlenart  staff   292301 May 16 08:19 log4j-api-2.13.3.jar
-rw-r--r--+ 1 lukaszlenart  staff   264012 Apr 19 10:23 ognl-3.2.14.jar
-rw-r--r--+ 1 lukaszlenart  staff  1530698 Jun 22 08:12 struts2-core-2.6-SNAPSHOT.jar
```

lukaszlenart: Just added the version range as it works with 2.6 as well.

waghanza: I was tricked liked that before (either with `maven` or `gradle`)

java/struts2/config.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+  website: struts.apache.org
+  version: 2.5
+
+build:
+  mvn clean package
```
waghanza, line 6: must be an array (missing a **-**)

waghanza:
It will run but

+ we have `\n` for get /  
+ we have `\n\n0` for get /user/0
+ post on /user is failing

lukaszlenart:
Should be fixed

waghanza:
may I add you to `CODEOWNER` @lukaszlenart ? so as to ping you in case of modification of `struts2` here

java/struts2/src/main/webapp/WEB-INF/user-id.jsp
```diff
@@ -0,0 +1,3 @@
+<%@ taglib prefix="s" uri="/struts-tags" %>
+<%@ page contentType="text/text;charset=UTF-8" %>
```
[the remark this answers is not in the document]
lukaszlenart: Right, I overlooked that :(

lukaszlenart:
> may I add you to `CODEOWNER` @lukaszlenart ? so as to ping you in case of modification of `struts2` here

Sure :)

waghanza:
@lukaszlenart can we create a single binary or package embedding all dependencies ? (I think it's a **fatjar** or **uberjar** that I need)

The main idea is to avoid relying on a binary package for `jetty-runner`, but embed into the `pom.xml` (ensure the version ...) and it could :+1: for next version (on `cloud`)

lukaszlenart:
You would like to avoid using `jetty-runner` and have everything embedded in one fat jar? And the run it with `java -jar target/struts2.jar`?

PR #3054, opened 2020-06-28
[Java] Disable logging

waghanza:
Hi,

This `PR` disable logging on all `java` frameworks.

Logging is disabled for :

+ [x] act
+ [x] blade
+ [x] javalin
+ [x] jooby
+ [x] micronaut
+ [x] rapidoid
+ [x] spring-boot
+ [x] spring-framework
+ [x] struts2

Regards,

PR #3064, opened 2020-06-28
jooby: disable logging

jknack:
Ref #3054


waghanza:
Thanks @jknack. I re-integrate into #3054

PR #3077, opened 2020-07-02
Add light-4j benchmark test

### LINKED ISSUE #2180: Light-4j Java framework
+ Language : Java
+ Repo : https://github.com/networknt/light-4j
+ Author :

tsuilouis:
Closes #2180.

I successfully ran the (new) `bin/neph java.light-4j` job on my local machine. If desired, I can paste the `bin/db to_readme` result here but please note I only ran it once (along with a couple of other Java frameworks for comparison) due to how long it took, so it may not be representative.

Thanks for reviewing!


java/light-4j/src/main/resources/config/openapi-security.yml
```diff
@@ -0,0 +1,43 @@
...
+# Enable JWT verification flag.
+enableVerifyJwt: ${openapi-security.enableVerifyJwt:false}
+
+# Extract JWT scope token from the X-Scope-Token header and validate the JWT token
+enableExtractScopeToken: ${openapi-security.enableExtractScopeToken:true}
+
+# Enable JWT scope verification. Only valid when enableVerifyJwt is true.
+enableVerifyScope: ${openapi-security.enableVerifyScope:true}
+
+# Enable JWT scope verification.
```
waghanza, line 15: trailing whitespace

java/light-4j/config.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+  website: doc.networknt.com
+  version: 1.6.23
```
waghanza, line 3: could you use `1.6` ?

java/light-4j/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+
+build:
+  - mvn clean package -Prelease
```
waghanza, line 6: what does `-Prelease` mean ?

java/light-4j/pom.xml
```diff
@@ -0,0 +1,428 @@
...
+    <packaging>jar</packaging>
+    <name>benchmark</name>
+    <version>0.0.1</version>
+
+    <properties>
+        <java.version>1.8</java.version>
+        
+        <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <project.main.class>com.networknt.server.Server</project.main.class>
+        <version.light-4j>1.6.23</version.light-4j>
```
waghanza, line 15: could you use a range like `[1.6,1.7)` to enforce using last release in **1.6.x**

java/light-4j/pom.xml
```diff
@@ -0,0 +1,428 @@
...
+        <version.jackson>2.10.0</version.jackson>
```
waghanza, line 16: is there any restriction to use `*` ? I mean it could better to have results with last version, if stable, than having to ping each dependencies manually

java/light-4j/pom.xml
```diff
@@ -0,0 +1,428 @@
...
+        <version.junit>4.12</version.junit>
+        <version.undertow>2.0.29.Final</version.undertow>
+        <version.json-schema-validator>1.0.29</version.json-schema-validator>
+        <version.fastscanner>2.18.1</version.fastscanner>
+        
+        
+        
+        
+        
+        <versions.maven-version>2.4</versions.maven-version>
```
waghanza, line 29: Is this not provided globally ? I mean by https://github.com/the-benchmarker/web-frameworks/blob/master/java/Dockerfile (without any specific image in `config.yaml` mave **3.6** is used)

java/light-4j/pom.xml
```diff
@@ -0,0 +1,428 @@
...
+        </dependency>
+        
+    </dependencies>
+
+    <build>
+        <plugins>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.0</version>
```
waghanza, line 233: `maven-version` is pined. what does pining `maven-compiler-plugin` means in such case ?

java/light-4j/src/main/resources/config/openapi-security.yml
stevehu: These files are configuration files to enable or disable cross-cutting concerns in the request/response chain. They are checked during the server startup and it won't impact the runtime performance as the security is not enabled for this performance test.

java/light-4j/config.yaml
stevehu: Is there any requirement for the version of JDK? The 1.6.x is for JDK 8 and 2.0.x is for JDK 11. Most users are using the JDK11 version which is a little bit faster.

java/light-4j/config.yaml
stevehu: -Prelease is a profile in the generated pom.xml to build a fat jar. During the development cycle, we don't build fat jar at all to save time for developers. The fat jar will be built only before release.

java/light-4j/config.yaml
waghanza: understood, I prefer to keep on jdk 8 to be as-fair-as possible in `java` **frameworks**. somes, at least, `act` is not compatible with `11` yet

stevehu: That is Cool. Let's keep the JDK8 then. Thanks for the input.

tsuilouis:
Thanks for the feedback @waghanza. I'll make changes to relax the versioning in the pom. Also, I'll try to figure out if any files are not strictly necessary to run the application when it is being benchmarked.

waghanza:
@tsuilouis @stevehu I you want to add yourself in `.github/CODEOWNERS` feel free to do so for `light-4j` :heart: I'll then ping you on any modification

java/light-4j/pom.xml
tsuilouis: Done in b81fb2691f80630445b1add4f978e939c0180966.

java/light-4j/pom.xml
tsuilouis: This property name is a bit misleading; I believe the version is for the [Versions Maven Plugin](https://www.mojohaus.org/versions-maven-plugin/), not the version of Maven itself. I don't think it is actually necessary here so I've removed both the property and the plugin itself (in b81fb2691f80630445b1add4f978e939c0180966).

java/light-4j/pom.xml
tsuilouis: At least for Jackson, when I tried using version 3, the application failed to run due to a missing method. I did update framework versions to allow for newer versions.

tsuilouis:
@waghanza I discussed with @stevehu and I realized that I didn't need all the middleware handlers and configuration generated by default when benchmarking. I've removed those yaml config files and disabled those handlers. As a pleasant surprise, on my local machine the max req/s more than doubled after this change. 😄

java/light-4j/pom.xml
waghanza: perhaps `[2,3)` could work

tsuilouis: Okay, I managed to get it working with a range in 18bddd3079846cb9e0607fbf32d5045a33d70b1a.

waghanza:
@tsuilouis @stevehu I've update result in https://github.com/tsuilouis/web-frameworks/tree/add-light-4j#results.

I'll merge after `CI`

PR #3102, opened 2020-07-06
[Java] Update light4j to 2.0

waghanza:
Hi @stevehu @tsuilouis,

This `PR` update https://github.com/networknt/light-4j to **2.0**

As I notice we have to compile targeted VM, right ?

Regards,

### DIFF
--- a/java/light-4j/config.yaml
+++ b/java/light-4j/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: doc.networknt.com
-  version: 1.6
+  version: 2.0
 
 build:
   - mvn clean package -Prelease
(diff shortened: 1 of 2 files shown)


stevehu:
I thought we had agreed to stay with 1.6.x release as it is using Java 8 which is consistent with other frameworks. What do you think?

waghanza:
Indeed. However `act` could be used with `java` **11** (with last release), so I'm upgrading all `java` **frameworks** in https://github.com/the-benchmarker/web-frameworks/pull/3101 which this `PR` target :heart:

stevehu:
Got it. It is a good move to go with Java 11. Most of our users are using Java 11 these days.

waghanza:
The their is no change needed ? I mean this `PR` is OK ?

stevehu:
Should be OK if the maven command is OK.

PR #3112, opened 2020-07-08
Embedded jetty

lukaszlenart:
This PR uses an embedded Jetty instead of deprecated jetty-runner and also uses new PlainResult instead of JSPs

### DIFF
--- a/java/struts2/pom.xml
+++ b/java/struts2/pom.xml
@@ -5,11 +5,12 @@
     <groupId>org.apache.struts.benchmarker</groupId>
     <artifactId>struts2</artifactId>
     <version>1.0</version>
-    <packaging>war</packaging>
+    <packaging>jar</packaging>
     <name>struts2</name>
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+        <jetty.version>9.2.30.v20200428</jetty.version>
         <java.version>11</java.version>
     </properties>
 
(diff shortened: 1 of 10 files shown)


java/struts2/pom.xml
```diff
@@ -5,11 +5,12 @@
     <groupId>org.apache.struts.benchmarker</groupId>
     <artifactId>struts2</artifactId>
     <version>1.0</version>
-    <packaging>war</packaging>
+    <packaging>jar</packaging>
```
waghanza, line 8: Just for information, what is the difference between `war` and `jar` deployment method ?

Also, is the any universal deployment type, like for `sbt` in **scala** ?
https://github.com/the-benchmarker/web-frameworks/blob/master/scala/Dockerfile#L7

java/struts2/pom.xml
```diff
@@ -22,33 +23,86 @@
...
+            </resource>
+            <resource>
+                <directory>src/main/webapp</directory>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
+                <version>3.8.1</version>
```
waghanza, line 61: Is it require ? I mean there seems not to have any problems if this version is removed

java/struts2/pom.xml
```diff
@@ -22,33 +23,86 @@
...
                 <configuration>
                     <encoding>UTF-8</encoding>
                     <source>1.8</source>
```
waghanza, line 64: Should not in be `11` ?

waghanza:
`CI` is failing
https://travis-ci.com/github/the-benchmarker/web-frameworks/jobs/358711886#L367

java/struts2/pom.xml
lukaszlenart: `war` is used when you want deploy the package into a Java servlet container (like jetty-runner or standalone Jetty server). `jar` is a common Java package used for libraries and apps (executable). If you want to start your app with `java -jar` it has to be a proper jar. I replaced `war` with `jar` as `war` is not needed any more.

No, you must use maven-assembly-plugin, btw. `universal` is also a plugin https://www.scala-sbt.org/sbt-native-packager/formats/universal.html

java/struts2/pom.xml
lukaszlenart: Dropped by using a proper properties

java/struts2/pom.xml
lukaszlenart: Yes, fixed

lukaszlenart:
> `CI` is failing
> https://travis-ci.com/github/the-benchmarker/web-frameworks/jobs/358711886#L367

Ach... SNAPSHOT isn't available :( I will try to fix it other way

waghanza:
Thanks @lukaszlenart. Could you do the same for `spring-framework` ?

lukaszlenart:
Sorry, I haven't been using Spring for a very long time and I have no idea how annotations Spring magic works :)

waghanza:
Ok. No problem. I will reorganise to do both in the same time (I'll ping someone, since ain't java skills)

java/struts2/pom.xml
```diff
@@ -22,32 +24,74 @@
...
+            <resource>
+                <directory>src/main/webapp</directory>
+            </resource>
+        </resources>
         <plugins>
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
-                <artifactId>maven-compiler-plugin</artifactId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.4</version>
```
waghanza, line 62: Could we avoid versioning plugins ? I mean having version here requires to track updates

java/struts2/src/main/resources/struts.xml
```diff
@@ -14,14 +14,10 @@
         <default-action-ref name="index"/>
         <global-allowed-methods>index,view,create</global-allowed-methods>
 
-        <action name="index">
-            <result>/WEB-INF/index.jsp</result>
-        </action>
+        <action name="index" class="org.apache.struts.benchmarker.example.UserAction"/>
 
         <action name="user/*" class="org.apache.struts.benchmarker.example.UserAction">
```
waghanza, line 19: Does this mean that `/user/a/4` is working ? I mean the _*_ does not mean any character allowed ?

java/struts2/pom.xml
lukaszlenart:
```
[WARNING]
[WARNING] Some problems were encountered while building the effective model for org.apache.struts.benchmarker:struts2:jar:1.0
[WARNING] 'build.plugins.plugin.version' for org.apache.maven.plugins:maven-jar-plugin is missing. @ line 59, column 21
[WARNING]
[WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
[WARNING]
[WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
[WARNING]
```

java/struts2/src/main/resources/struts.xml
lukaszlenart: Yes, it's working but it will try to map `a/4` as a user ID

waghanza: In fact this project requires that `/user/1`, `/user/2` ... Works, but not any deeper

`/user/a/4` should not work

lukaszlenart: Hm... need to double check and maybe fix that

waghanza: Let's fix this an other time

PR #3113, opened 2020-07-08
Optimized code for hyperf.

limingxinleo:


limingxinleo:
`php sbin/hyperf.php` will generate cache

it will optimize the speed of start command.

PR #3148, opened 2020-07-15
[Crystal] update spider-gazelle to v3.3.0

stakach:
faster router


stakach:
@waghanza done :)

waghanza:
@stakach the results has been updated in https://github.com/the-benchmarker/web-frameworks#results, but there is neither increase nor decrease, weird but I won't focus on that (wast of time since docker is involved)

PR #3051, opened 2020-06-28
[Crystal] Disable logging

waghanza:
Hi,

This `PR` disable logging on all `crystal` frameworks.

Logging is disabled for :

+ [x] amber
+ [x] athena
+ [x] grip
+ [x] kemal
+ [x] lucky
+ [x] orion
+ [x] router.cr
+ [x] runcobo
+ [x] shivneri
+ [x] spider-gazelle
+ [x] toro

Regards,

PR #3184, opened 2020-07-24
Disable Shivneri logging

ujjwalguptaofficial:


crystal/shivneri/src/server.cr
```diff
@@ -7,6 +7,6 @@ module Server
   Shivneri.port = 3000
   Shivneri.host = "0.0.0.0"
   Shivneri.open do
```
waghanza, line 9: is the block (`do` ... `end`) required ?

ujjwalguptaofficial: have removed

waghanza:
Thanks @ujjwalguptaofficial. I integrate you work in #3051

Issue #3191, opened 2020-07-27
[Ruby] Fix pakyow


waghanza:
Hi @bryanp,

As I understand `pakyow` requires now `mini_racer`

Since this **gem** requires extra dependencies on **alpine**, the build is failing.

Regards,

bryanp:
Yes, Pakyow has required `mini_racer` since the v1.0 release a year ago. I'll dig in this week to see why the build is failing.

bryanp:
@waghanza Can you please point me to the failure that you're seeing? I went looking and couldn't find it.

waghanza:
Sorry for the delay @bryanp. I have not a huge bandwidth for this project this month :stuck_out_tongue: 

I thin the issue is related to `docker`

waghanza:
Seems that `mini_racer` is **NOT** usable (compilation issue) on `alpine`, since `pakyow` rely on `0.2.15`
https://github.com/pakyow/pakyow/blob/6ec68aa0ec7352e4ada704a8fbc238c275d26d47/pakyow-assets/pakyow-assets.gemspec#L30

btw, `mini_racer` `0.3.1` is :ok_hand: on `alpine`

PR #3216, opened 2020-08-04
Fix Pakyow, tune performance

bryanp:
@waghanza This should fix the Pakyow issue that we discussed in https://github.com/the-benchmarker/web-frameworks/issues/3191. I realized that the benchmark app shouldn't even be using the `assets` framework since it doesn't rely on assets at all. Removing the `assets` framework was as easy as removing the metagem from `Gemfile`, which simply defines the standard gemset as a project dependency.

I also changed the command to boot in `ludicrous` mode, which configures the application for benchmarking.

Still can't get the benchmark to run for me locally, but it appears to be some other issue:

```
Successfully tagged ruby.pakyow:latest
[Error]  -- STDERR(.neph/ruby.pakyow/log/log.err) --
  from ../../lib/db/src/db.cr:152:5 in 'build_database'
  from ../../lib/db/src/db.cr:148:5 in 'build_database'
  from ../../lib/db/src/db.cr:116:5 in 'open'
  from ../../tools/src/client.cr:33:5 in 'run'
  from ../../tools/src/client.cr:22:1 in 'parse_and_run'
  from ../../tools/src/client.cr:22:1 in 'run'
  from ../../tools/src/client.cr:116:1 in '__crystal_main'
  from ../../../../../../../usr/local/Cellar/crystal/0.35.1/src/crystal/main.cr:105:5 in 'main_user_code'
  from ../../../../../../../usr/local/Cellar/crystal/0.35.1/src/crystal/main.cr:91:7 in 'main'
  from ../../../../../../../usr/local/Cellar/crystal/0.35.1/src/crystal/main.cr:114:3 in 'main'
```

Other benchmarks, e.g. `ruby.roda` have the exact same issue on my end.

### DIFF
--- a/ruby/pakyow/config.yaml
+++ b/ruby/pakyow/config.yaml
@@ -2,4 +2,4 @@ framework:
   website: pakyow.com
   version: 1.0
 
-command: bundle exec pakyow boot -e production
+command: bundle exec pakyow boot -e ludicrous
--- a/ruby/pakyow/config.yaml
+++ b/ruby/pakyow/config.yaml
@@ -2,4 +2,4 @@ framework:
   website: pakyow.com
   version: 1.0
 
-command: bundle exec pakyow boot -e ludicrous
+command: bundle exec pakyow boot -e production
--- a/ruby/pakyow/Gemfile
+++ b/ruby/pakyow/Gemfile
@@ -1,5 +1,4 @@
 source "https://rubygems.org"
 
-gem "pakyow", "~> 1.0.0"
 gem "pakyow-core", "~> 1.0.0"
 gem "pakyow-routing", "~> 1.0.0"


ruby/pakyow/config.yaml
```diff
@@ -2,4 +2,4 @@ framework:
   website: pakyow.com
   version: 1.0
 
-command: bundle exec pakyow boot -e production
+command: bundle exec pakyow boot -e ludicrous
```
waghanza, line 5: what is `ludicrous` @bryanp ?

bryanp: It's a mode that configures an application for benchmarking (also a [reference to the movie "Spaceballs"](https://www.youtube.com/watch?v=oApAdwuqtn8) 😂)

waghanza: ah probably better to keep `production` mode. the idea is to reflect how can `pakyow` perform is real-world. wdyt ?

bryanp: Ah, good point. Reverted that commit: https://github.com/the-benchmarker/web-frameworks/pull/3216/commits/e839ea73050a5b2498801741bda63311ec515738

PR #3222, opened 2020-08-06
[Go] Update Gearbox Init Settings

abahmed:


abahmed:
@waghanza

go/gearbox/main.go
```diff
@@ -13,7 +13,8 @@ var sendID = func(c gearbox.Context) {
 
 func main() {
 	app := gearbox.New(&gearbox.Settings{
-		Prefork: true,
+		Prefork:               false,
+		DisableStartupMessage: true,
```
waghanza, line 17: I can see that `fiber` has the same kind of options, could you give some more informations about thos options ?

abahmed: **Prefork**: https://www.reddit.com/r/learnprogramming/comments/25vdm8/what_is_a_prefork_worker_model_for_a_server/


**DisableStartupMessage**: This is just an startup message printed in the console with the port used by service and app version

PR #3225, opened 2020-08-07
[PHP] add nano

laylatichy:
add nano

Nano - small micro api framework based on workerman that helps you write fast API

Thank you


php/nano/config.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+  gitlab: x.laylatichy.x/nano
```
waghanza, line 2: `gitlab` keyword is not available. you are the first one entirely on `gitlab`. I've open an issue so as this configuration could be written https://github.com/the-benchmarker/web-frameworks/issues/3226

laylatichy: I wanted to change this but I see you created a new branch for that already, cheers for that

didn't check toolsdir earlier for that scenario

PR #3228, opened 2020-08-07
[Python] Add django-ninja

vitalik:


python/django-ninja/requirements.txt
```diff
@@ -0,0 +1,4 @@
+django>=3.1.0,<3.2.0
+django-ninja==0.3.1
```
waghanza, line 2: could you use a more relax constraint ? e.g ->  `django-ninja>=0.3.1,<0.4.0`

python/django-ninja/requirements.txt
```diff
@@ -0,0 +1,4 @@
+django>=3.1.0,<3.2.0
```
waghanza, line 1: pining `django` is necessary here ?

python/django-ninja/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: http://django-ninja.rest-framework.com/
+  version: 0.3.1
```
waghanza, line 3: could you use `X.Y` format ? e.g : `0.3`

python/django-ninja/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: http://django-ninja.rest-framework.com/
```
waghanza, line 2: could you remote `http` -> will be added automatically

python/django-ninja/requirements.txt
vitalik: yes, 3.1+ comes with async views that are required

python/django-ninja/config.yaml
vitalik: done

python/django-ninja/config.yaml
vitalik: done

python/django-ninja/requirements.txt
vitalik: done

python/django-ninja/requirements.txt
waghanza: Is the 3.2 released ? 
I mean :
+ without any version pip will take the last version
+ this should not be done by `django-ninja` itself ?

vitalik: @waghanza , indeed - updated requirements

python/django-ninja/requirements.txt
```diff
@@ -0,0 +1,3 @@
+django-ninja==0.6.0
```
waghanza, line 1: better to use a range, no ?
`django-ninja>=0.6.0,<0.7.0`

waghanza (changes requested):

vitalik:
@waghanza  done (well I set range till 1.0 as I expect only minor changes and bug fixes before 1.0)

python/django-ninja/requirements.txt
```diff
@@ -0,0 +1,3 @@
+django-ninja>=0.6.0,<1.0.0
```
waghanza, line 1: could you update to `>=0.8.0,<0.9.0` ?

python/django-ninja/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: django-ninja.rest-framework.com
+  version: 0.3
```
waghanza, line 3: could you update to `0.8` ?

waghanza (changes requested):
could you also rebase (I've changed CI) @vitalik ?

vitalik:
done

python/django-ninja/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: django-ninja.rest-framework.com
+  version: 0.8.2
```
waghanza, line 3: could you use 0.8 instead of 0.8.2 ?

python/django-ninja/config.yaml
vitalik: sure, done

waghanza (approved):
I don't know why CI is stuck, but this is OK imho

Thanks @vitalik.

May I add you to `CODEOWNER` in order to ping you on each `django-ninja` update `here` ?

PR #3231, opened 2020-08-08
chubbyphp-workerman

dominikzogg:


waghanza:
Failing at https://travis-ci.com/github/the-benchmarker/web-frameworks/jobs/369999957. I suspect there is a multiprocessing involved. Need to investigate, `docker` does not like that (same thing when using **prefork** in `go`)

dominikzogg:
@waghanza i used the wrong port (8080 instead of 3000), sorry for that

waghanza:
Ah ok, it is not the problem I thought 😛 

------

Note that the results will **NOT** be updated this month

dominikzogg:
thx

PR #3242, opened 2020-08-14
[Python] Add index.py

abersheeran:
https://github.com/abersheeran/index.py

### DIFF
--- a/python/index.py/config.yaml
+++ b/python/index.py/config.yaml
@@ -1,0 +1,12 @@
+framework:
+  website: index-py.abersheeran.com
+  version: 0.12.1
+
+command: >
+  gunicorn  \
+    --log-level warning  \
+    --bind 0.0.0.0:3000  \
+    --reuse-port  \
+    --workers $(nproc)  \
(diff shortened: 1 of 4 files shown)


python/index.py/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  website: index-py.abersheeran.com
+  version: 0.12.2
```
waghanza, line 3: I prefer to use `X.Y` notation -> e.g : `0.12`

python/index.py/requirements.txt
```diff
@@ -0,0 +1 @@
+index.py[gunicorn]
```
waghanza, line 1: what is this notation ?
should not it be something like
```
index.py>=0.12.0,<0.13.0
gunicorn
uvicorn
```

abersheeran:
> Could you also add `index.py` into `.travis.yml` ?

Because the frame name has `.`, I don't know how to add it. Is it like this?

```
      - FRAMEWORK=python.index.py
```

PR #3257, opened 2020-08-24
[Perl] Specify version for dancer2

### LINKED ISSUE #3253: [Perl] Specify version for dancer
Hi,

The actual `cpanfile` does not contains any version.

Regards,

waghanza:
Hi,

This `PR` specify `dancer2` **version**

Regards,

-----------
Closes #3253

### DIFF
--- a/perl/dancer2/cpanfile
+++ b/perl/dancer2/cpanfile
@@ -1,4 +1,4 @@
-requires 'Dancer2';
+requires 'Dancer2', '> 0.3';
 # Gazelle or Starman
 requires 'Gazelle';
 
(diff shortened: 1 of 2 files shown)


perl/dancer2/cpanfile
```diff
@@ -1,4 +1,4 @@
-requires 'Dancer2';
+requires 'Dancer2', '> 0.3';
```
waghanza, line 1: Is it correct to right version as is @xsawyerx ?

I mean is there a standard way to say that I want to install any `0.3.x` ?

PR #3277, opened 2020-08-28
[PHP] Change process count for Mark

passwalls:
Change process count for Mark

### DIFF
--- a/php/mark/start.php
+++ b/php/mark/start.php
@@ -5,7 +5,7 @@ require 'vendor/autoload.php';
 
 $api = new SimpleApp('tcp://0.0.0.0:3000');
 
-$api->count = 8;
+$api->count = shell_exec('nproc') ? shell_exec('nproc') - 4 : 12;
 
 $api->get('/', function ($requst) {
     return '';


php/mark/start.php
```diff
@@ -5,7 +5,7 @@
 
 $api = new SimpleApp('tcp://0.0.0.0:3000');
 
-$api->count = 8;
+$api->count = shell_exec('nproc') ? shell_exec('nproc') - 4 : 12;
```
waghanza, line 8: why this number @passwalls ?

php/mark/start.php
```diff
@@ -5,7 +5,7 @@
...
 
 $api->get('/', function ($requst) {
```
waghanza, line 10: a typo ?

php/mark/start.php
passwalls: I want to reserve some CPU resources for wrk and the operating system, so the number of processes will be less than the number of CPUs.

waghanza: ah, I see. In **production** (soon) `wrk` will **NOT** run the same hardware.

Just for information, `workerman` use https://github.com/the-benchmarker/web-frameworks/blob/725e5367642c239fc10a67b50a144d3f8dc774ae/php/workerman/start.php#L7
or for  **TechEmpower** 

@walkor recommends 1-3 proc count
 https://github.com/walkor/Workerman/issues/445#issuecomment-521104641

passwalls: Ok, thanks for your information. Looking forward to the coming of **production**. 👍

waghanza: you can track https://github.com/the-benchmarker/web-frameworks/pull/3118 @passwalls

passwalls:
Closed and waiting the coming of production.

PR #3278, opened 2020-08-28
[PHP] Update Simps

sy-records:


php/simps/config/servers.php
```diff
@@ -19,8 +19,6 @@
         ],
         'settings' => [
             'worker_num' => swoole_cpu_num() * 2,
-            'open_tcp_nodelay' => true,
-            'enable_reuse_port' => true,
```
waghanza, line 23: why removing this ? I suppose the performance decrease for this use case ?

sy-records: I feel like it's useless because I only start one port.

waghanza: reuse_port is about port sharing (same port)
https://lwn.net/Articles/542629/

btw, it could be useless here, honestly I don't know how `swoole` works

any informations @limingxinleo @huangzhhui ?

limingxinleo: `enable_reuse_port` is useless for this point.

I think it could be removed from project.

huangzhhui:
lgtm

huangzhhui (approved):

php/simps/config/servers.php
waghanza: Is it not enabled by default ? I mean should we not set it to `false` ?

sy-records: The default is false, so there is no need to set

waghanza: ok

waghanza:
sorry for the delay @sy-records, I'll merge this before the next round (I'll try to standardize `swoole` setup for php frameworks listed here)

PR #3290, opened 2020-09-02
[PHP] Used php7.4 preload feature 

### LINKED ISSUE #3291: PHP7.4 is not optimized
According to my tests
benchmark of php7.4 - phalcon framework showed  10x better results then in basic configuration that was made in your repo.
My changes you can easily review in  my PR [https://github.com/the-benchmarker/web-frameworks/pull/3290/files](url)

lodry:
-----
Closes #3291


php/phalcon/Dockerfile
```diff
@@ -0,0 +1,62 @@
+FROM php:7.4-fpm-alpine
```
waghanza, line 1: This file is not only for `phalcon`, but for all **php** frameworks (should not be versioned)

php/phalcon/Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+  fastcgi_pass unix:/var/run/php-fpm.sock;\n\
+  fastcgi_param   SCRIPT_FILENAME         $document_root/index.php;\n\
+  include fastcgi_params;\n\
+}\n\
+
+  }\n'\
+  >> /etc/nginx/conf.d/www.conf
+  
+  RUN echo -e 'opcache.enable=1\n\
+  opcache.preload="/usr/src/app/preload.php"\n\
```
waghanza, line 52: nice tips, could it be used in any other framework ?

waghanza (commented):
Nice improvements 🎉 

Could this be applied to all php frameworks ?

php/phalcon/Dockerfile
lodry: yes,
 but should be configured for each framework

php/phalcon/Dockerfile
lodry: how then can I inject configs into php.ini?

waghanza: Unfortunately, you can't right now. The only exception is that this configuration is the same for all frameworks (in this case you could add entry in `php/Dockerfile`). I'll create a `PR` if needed

php/phalcon/Dockerfile
waghanza: sure, you can use `php/Dockerfile` (the idea is to have the same setup for all php frameworks)

php/phalcon/Dockerfile
lodry: yeah i can create such common config

waghanza:
@lodry We have changed a lot the code base, I'll then create an other `PR` to re-active this.

Could you confirm that this `PR` contains 2 main ideas ?
+ Add  `opcache.preload`
+ Setting the above config to a file that `require_once` document root

PR #3316, opened 2020-09-14
[Go] Add Goyave

### LINKED ISSUE #2647: Introduce goyave
Hi @System-Glitch,

Would you mind to add (or consent that we add) [goyave](https://github.com/System-Glitch/goyave) in `here` ?

Regards,

System-Glitch:
Added [Goyave](https://github.com/System-Glitch/goyave/).

Closes #2647

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -11,6 +11,7 @@
 /go/gearbox/ @muhammednagy
 /go/clevergo/ @razonyang
 /go/mars/ @roblillack
+/go/goyave/ @System-Glitch
 
 /ruby/roda/ @jeremyevans
 /ruby/pakyow/ @bryanp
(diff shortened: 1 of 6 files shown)


waghanza (changes requested):

waghanza (commented):
seems that the server is unreachable https://travis-ci.com/github/the-benchmarker/web-frameworks/jobs/386491970#L425

System-Glitch:
How does `benchmark_spec.cr` find the URI? I tried to dig in the code but couldn't find what the problem is.

waghanza:
> I think it doesn't work because the Dockerfile doesn't copy the configuration file (config.json)

Seems to, let me check

> How can we make a patch in the dockerfile to fix that?

Not sure. Let me add some commits on your branch. I don't remember any `go` frameworks that requiresd a config file to work, so I do not work in this direction

System-Glitch:
I think I fixed it by adding `files: - config.json`to `config.yaml`. You should be allowed to add commits to my branch.

waghanza:
Ok, this file is need at runtime, and not compile. 🎉

Issue #3303, opened 2020-09-09
[R] Add plumber


t-wojciech:
[plumber](https://www.rplumber.io) is the most popular API development tool in R. RestRserve has already been added in #3255.

> This project compare performance based on 3 routes :
> 
> - GET /, SHOULD return a successful status code and an empty string
> - GET /user/:id, SHOULD return a successful status code and the id
> - POST /user, SHOULD return a successful status code and an empty string

Gently ping authors: @trestletech, @schloerke, @cpsievert, @meztez

waghanza:
Thanks @t-wojciech.

Waiting for explicit consent, or no opposition from any author 😛 to start working on it

schloerke:
@waghanza 👍 

------------- 

The dev version supports async. If installing from master (or >= v1.0.0), don't know if adding async support as another test would be interesting to compare

Ex:
```r
function() {
	return(
	  promises::promise(~ {later::later(resolve(""), 0) })
	)
}
```

t-wojciech:
plumber 1.0.0 in on [CRAN](https://cran.r-project.org/web/packages/plumber/index.html) now.

vikram-rawat:
I was about to add this too... as RestRServe has already been added. I would definitely like to see the difference. But the benchmarks haven't been updated for over a month.

schloerke:
@t-wojciech Thinking on this some more, I want to be sure we are actually testing plumber and not the filters or serializers.

I've added a test file below to compare how fast the core of plumber is, with and without filters and serializers.  Routes that end in default test using filters and a serializer. The testing routes test the core plumber response time.

```r
# plumber.R

# GET /, SHOULD return a successful status code and an empty string
# GET /user/:id, SHOULD return a successful status code and the id
# POST /user, SHOULD return a successful status code and an empty string

named_empty_headers <- setNames(list(), character(0))
#* @preempt __first__
#* @get /
function(res) {
  # https://github.com/rstudio/httpuv/issues/195
  res$headers <- named_empty_headers
  res
}

#* @serializer contentType list(type = "text/plain")
#* @get /default
function() {
  ""
}


#* @preempt __first__
#* @get /user/<id>
function(res, id) {
  # https://github.com/rstudio/httpuv/issues/195
  res$headers <- named_empty_headers
  res$body <- id
  res
}

#* @serializer contentType list(type = "text/plain")
#* @get /user/<id>/default
function(id) {
  id
}


#* @preempt __first__
[...]

PR #3334, opened 2020-09-22
Add R `plumber` and `httpuv`

schloerke:
Fixes https://github.com/the-benchmarker/web-frameworks/issues/3303

Updates:
* Adds `r.plumber` framework
* Adds `r.httpuv` framework

---------------

Commands used to tests locally on macOS:

```bash
# install
brew install crystal wrk postgresql docker-machine
/usr/local/opt/postgres/bin/createuser -s postgres

# set up docker
docker-machine rm default --force
docker-machine create default
eval $(docker-machine env default) 

# build
shards build
bin/make config --driver docker-machine

# set up database
export DATABASE_URL=postgresql://postgres@localhost/benchmark
dropdb -U postgres benchmark
createdb -U postgres benchmark
psql -U postgres -d benchmark < .ci/dump.sql

# test
bin/neph r --seq 

# save output
bin/db to_readme
```

Relevant local test output:
| 1 | r (4.0)| [httpuv](https://https://github.com/rstudio/httpuv/) (1.5.4) | 987 | 1 012 | 835 |
| 2 | r (4.0)| [plumber](https://rplumber.io) (1.0.0) | 223 | 240 | 229 |
| 3 | r (4.0)| [restrserve](https://restrserve.org) (0.3) | 125 | 103 | 37 |

------------

cc @t-wojciech

### DIFF
--- a/r/httpuv/config.yaml
+++ b/r/httpuv/config.yaml
@@ -1,3 +1,3 @@
 framework:
-  website: https://github.com/rstudio/httpuv/
-  version: 1.5.4
+  github: rstudio/httpuv
+  version: 1.5
--- a/r/httpuv/config.yaml
+++ b/r/httpuv/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: https://github.com/rstudio/httpuv/
+  version: 1.5.4
--- a/r/plumber/config.yaml
+++ b/r/plumber/config.yaml
@@ -1,3 +1,3 @@
 framework:
   website: rplumber.io
-  version: 1.0.0
+  version: 1.0
--- a/r/plumber/config.yaml
+++ b/r/plumber/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: rplumber.io
+  version: 1.0.0
--- a/.travis.yml
+++ b/.travis.yml
@@ -131,6 +131,8 @@ env:
       - FRAMEWORK=java.javalin
       - FRAMEWORK=java.jooby
       - FRAMEWORK=java.light-4j
+      - FRAMEWORK=r.httpuv
+      - FRAMEWORK=r.plumber
       - FRAMEWORK=r.restrserve
       - FRAMEWORK=fsharp.falco
       - FRAMEWORK=fsharp.suave
--- a/r/httpuv/DESCRIPTION
+++ b/r/httpuv/DESCRIPTION
@@ -1,0 +1,2 @@
+Depends:
+  httpuv (>= 1.5.4)
--- a/r/httpuv/app.R
+++ b/r/httpuv/app.R
@@ -1,0 +1,26 @@
+s <- httpuv::startServer(host = "0.0.0.0", port = 3000,
+  app = list(
+    call = function(req) {
+      path <- req$PATH_INFO
+      body <-
+        if (path == "/" || path == "" || path == "/user") {
+          ""
+        } else {
+          # /user/id path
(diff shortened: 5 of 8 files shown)


t-wojciech:
Thanks for this PR. I am really surprised by the improved performance in the new version of `plumber` over `restrserve` in your local test. Good job!

waghanza (changes requested):

r/httpuv/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: https://github.com/rstudio/httpuv/
```
schloerke, line 2:
```suggestion
  github: rstudio/httpuv
```

r/httpuv/config.yaml
```diff
@@ -0,0 +1,3 @@
...
+  version: 1.5.4
```
schloerke, line 3:
```suggestion
  version: 1.5
```

r/plumber/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: rplumber.io
+  version: 1.0.0
```
schloerke, line 3:
```suggestion
  version: 1.0
```

vikram-rawat:
Great results @schloerke ... could you also post a benchmark with serialization added. so that We can see how much difference there actually is. Just for reference. Please

schloerke:
@vikram-rawat The plumber results above contain the default behavior (currently using the 4 default filters and using the content-type serializer).

Plumber sits on top of `httpuv`, so we can think that the speed would be upper bounded by `httpuv`'s speed.  

In my local testing in #3303, I found that plumber could be 2x faster if the filters were removed but this did not represent the typical user interaction. To be a better representation of what plumber is, I kept the filters in. (But speeding up the filters is something to address and look forward to!)

vikram-rawat:
Trust me I never expected this speed from plumber package. Plumber 1.0 has now certainly piqued my interest. I am really amazed. 👍 🥇  

It's twice as fast as RestRserve. All I want to know if there is any optimization that I missed in the code. @dselivanov should actually look into it. I would really like to know the result. as I am currently finalizing an API framework for next project.

schloerke:
I believe this comes from the default behavior of each routing package. The `hello-world` style testing done by `web-frameworks` is better to be done using a single thread (ex: `httpuv`, which `plumber` wraps). There is too much overhead to process "each request in a separate fork" (ex: `RServe`, which `RestRServe` wraps) when only returning static values.

`RestRServe` (with no specialized route code) will handle long execution processes in parallel, therefore will handle many long execution processes faster when compared to the default single-threaded processing of `plumber`. However, in `plumber` v1.0.0 `promises` (ex: `future::future()`) is now supported. So the long processing time performance comparison (https://restrserve.org/articles/benchmarks/Benchmarks.html#comparison-with-plumber) should be re-evaluated under comparable execution conditions (ex: same number of parallel processing engines).

dselivanov:
@schloerke you are right! Would you consider to send PR for fair benchmark against plumber 1.0? 

Actually overhead for processing each request in a separate fork is not huge. The issue is that request handler is getting JIT compiled beforehand in each fresh fork (https://github.com/rexyai/RestRserve/issues/149) . This will be fixed in the next release.

vikram-rawat:
Thanks for the info guys... I would love to see both frameworks grow... We really need you 👍

t-wojciech:
Great discussion, thank you for your efforts and I wish you next success!

PR #3381, opened 2020-10-18
Switch to SemaphoreCI

waghanza:
Hi,

Due to limitation (only 200 jobs are allowed on `travis`), we must switch to an other CI.

SemaphoreCI seems a good alternative.

@the-benchmarker/web-frameworks what do you think ?

Regards,


leocavalcante:
Sounds good to me. And, what about GH Actions?

waghanza:
Not against.

Honestly, I've never tried ... Will be happy to try later ...

PR #3390, opened 2020-10-28
[Python] Update uvicorn extras

abersheeran:


waghanza:
Thanks @abersheeran, I've notice that and include it in https://github.com/the-benchmarker/web-frameworks/pull/3381

PR #3430, opened 2020-11-08
Switch to make

waghanza:
Hi,

This `PR` switch from https://github.com/tbrand/neph to https://www.gnu.org/software/make. This changes is motivated by 3 reasons :
+ No need to parallelize tasks for this project
+ It is very helpful to have a verbose tool (easier debug)
+ Perhaps this project will move from `crystal`

Regards,

/cc @tbrand


tbrand:
:+1:

PR #3441, opened 2020-11-10
[CSharp] [FSharp] Update to .NET 5.0

panesofglass:


fsharp/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM fsharp:10.10-netcore AS build
```
panesofglass, line 1: This image is still based on dotnet-core.

waghanza: you can use `dotnet` image from microsoft to compile `F#` ?

panesofglass: Yes, I think the fsharp image is the build from fsharp.org, but F# is released as part of the dotnet-sdk, as well.

PR #3449, opened 2020-11-12
[Clojure] Add Yada #3012

### LINKED ISSUE #3012: [Clojure] Add Yada
@malcolmsparks would you mind adding (or helping me add) an entry for [yada](https://github.com/juxt/yada)?

panesofglass:
Closes #3012 

- [x] GET /
- [x] POST /user
- [x] GET /user/id


clojure/yada/src/server.clj
```diff
@@ -0,0 +1,39 @@
...
+(def getuser
+  (yada/resource
+    {:parameters {:path {:name String}}
+     :methods
+     {:get
+      {:produces "text/plain"
+       :response
+       (fn [ctx]
+         (let [name (get-in ctx [:parameters :path :name])]
+           name))}}}))
```
panesofglass, line 29: This resource is causing me some problems, apparently calling `nth` somehow that I don't understand.

panesofglass:
Both coast and yada appear to have issues in the test section where the execution attempts to generate a `cid.txt` and `ip.txt`. I don't see anything in the `ip.txt` if I try to run this locally on WSL Ubuntu 20.04. I suspect there are some differences in OS and other expectations, so this is unlikely to mean much.

PR #3478, opened 2020-11-20
[Scala] Disable logging

whiplash:
Not sure why It also added changes I did not make. But anyways, I've added `logback.xml` to Play file and set it to shut off logging


waghanza:
> Not sure why It also added changes I did not make

Since this `PR` target **master**, not the branch you create this `PR` from 😛

waghanza (commented):
Thanks @whiplash. 

However, I can see only `play` here. Are `http4s` and `akkahttp` already silent ?

whiplash:
My bad. Yeah, this is only Play

waghanza:
Please ping me when ready @whiplash

PR #3481, opened 2020-11-20
updated Luminus dependencies to version 3.87

yogthos:


waghanza (commented):
Thanks 🎉 

How can I check the version in `project.clj` ? (I mean I want to use a bot to take care of updates)

yogthos:
I could just add a `:luminus-version` key to `project.clj` then to help with that then.

PR #3482, opened 2020-11-20
[PHP] sw-fw-less performance optimization

luoxiaojun1992:


luoxiaojun1992:
1. Opened composer autoloader optimization switch
2. Removed unused provider configuration 
3. Add object pool configuration (to avoid creating new objects)

luoxiaojun1992:
Could you please review the PR ?  @waghanza  Thanks.

php/sw-fw-less/composer.json
```diff
@@ -21,5 +21,8 @@
       "App\\": "app/"
     }
   },
+  "config": {
+    "optimize-autoloader": true
```
waghanza, line 25: not need, this will be done for all php frameworks (using the command line)

luoxiaojun1992: Ok. It has been removed.

PR #3493, opened 2020-11-26
Use only 2xx http requests to compute `req/s`

### LINKED ISSUE #3484: Unverified 5xx or 4xx errors
Hi @waghanza 

See in this example below that there were 1048.14 requests per second, but **15791** responses is wrong.

```shell
# wrk -t8 -c64 -d15s --timeout 8s http://192.168.1.4:82/user/123
Running 15s test @ http://192.168.1.4:82/user/123
  8 threads and 64 connections
    Latency    53.39ms   30.83ms 371.49ms   76.14%
    Req/Sec   138.09     57.56   490.00     68.50%
  15791 requests in 15.07s, 4.52MB read
  Non-2xx or 3xx responses: 15791
Requests/sec:   1048.14
Transfer/sec:    307.03KB
```

See below that the script does not consider errors, so several frameworks may be with wrong scores.

https://github.com/the-benchmarker/web-frameworks/blob/03b28682a08a65ac0b563de70ec6f16b7a64815a/tools/src/client.cr#L84-L90

I came to this conclusion because Ubiquity (2.3) that runs with PHP-FPM had a high score compare other frameworks that run with PHP-SWOOLE. Not that Ubiquity is slow, but we know that the slowness is the PHP-FPM.

When investigating Ubiquity in Docker I noticed that he was not doing anything, as PHP-FPM was showing an error, there should be a configuration in /php/ubiquity/config.yaml referring to #slasheddocroot as it uses a php-fpm prefix different from the default. With that only NGINX was processing and returning error "502 Bad Gateway".

Here is a recommendation for correcting script web-frameworks/tools/src/client.cr and avoid erroneous scores:

```ruby
 output = stdout.to_s 
 lines = output.split("\n") 
 requests_errors = lines.grep(/Non-2xx/)
 if requests_errors.length > 0 # wrk found errors
     requests_errors = requests_errors.first.split(":").pop.strip.to_f 
 else # wrk found no errors
     requests_errors = 0 
 end
 requests_per_seconds = lines.grep(/Requests/).first.split(":").pop.strip.to_f - (requests_errors/flags.duration)
 
 # insert(framework_id, "request_duration", result[0]) 
 # insert(framework_id, "request_total", result[1]) 
 insert(framework_id, "request_per_second", requests_per_seconds, concurrency_level_id) 
```
Lines is correct? :see_no_evil:

Now, if have 4xx or 5xx errors, they will be subtracted from the total number of requests.

@leocavalcante i found the reason why Ubiquity made many requests with PHP-FPM and probably other frameworks have the same false positive :nerd_face:

waghanza:
Hi @RicardoSette,

Thanks for notification in #3484.

This `PR` takes count of `http errors` to compute `req/s`.

Regards,

/cc @panesofglass `C#` results are available in https://github.com/waghanza/web-frameworks/tree/handle_non_200x#results

-------
Partial update (missing ruby frameworks)

----------
Closes #3484


waghanza:
Some frameworks decrease (example `fiber`) due to error count.

I think since is de to `docker` layer, I will investigate when on cloud (no value added to this project to debug a dev/test tool)

waghanza:
@itsumura-h With following `--threadAnalysis:off --define: usePcreHeader --passL: "/usr/lib/libpcre.a"   --excessiveStackTrace:off  -d:release --threads:on` I have an **SIGSEGV: Illegal storage access. (Attempt to read from nil?)**

PR #3494, opened 2020-11-26
Update restheart with some performance improvements in service implementation

ujibang:
Also modified the configuration file. It includes two performance sensitive options

```yml
# Number of I/O threads created for non-blocking tasks. at least 2. suggested value: core*2
io-threads: 1

 # Number of threads created for blocking tasks (such as ones involving db access). suggested value: core*16
worker-threads: 4
```

On my local machine 1 and 2 respectively gets the maximum throughput on concurrency 64. I set it to 1 and 4 to cover bigger concurrency values. I guess this small values are better than suggested values because the test case is very trivial and does not require any *blocking* task.

However it really depends on the testing environment. Would be great if following values could be tried in test environment:

- 1,2
- 1,8
- 1,16
- 2,4
- 2,8
- 2,16
- 2,32


waghanza:
I see. You need to write a config file, in order to adjust thread numbers. Could this value be export in an environment value or something like that (it could be useful if the process is started by a process manager like systemd)

ujibang:
Ywa @waghanza 

the configuration file is `restheart.yml` and here the two parameters are defined.

The parameter in the configuration file can be overridden with environment variables like this:

```
RH_IO_THREADS=1 RH_WORKER_THREADS=1 java -jar benchmarkers.jar restheart.yml
```

You can fine more information on restheart documentation about [env variables](https://restheart.org/docs/setup/#environment-variables)

waghanza:
I'm glad to here that this is working with env vars. We will use env vars for the cloud version

java/restheart/pom.xml
```diff
@@ -10,7 +10,7 @@
     <mongodb.version>4.2</mongodb.version>
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
-    <restheart.version>[5.1.6,5.2)</restheart.version>
```
waghanza, line 13: could you use `[5.1,5.2)` ? no need to bump each patch level version

ujibang: done

waghanza (approved):

PR #3518, opened 2020-12-01
Add option to display gitlab url in readme

laylatichy:
@waghanza added gitlab url as I noticed it was missing on my nano framework, wasn't sure if that's the correct one we are using now


waghanza:
@laylatichy which framework do you refers ? sorry if I miss something

laylatichy:
@waghanza added myself as php nano codeowner and I manually fixed url in readme, it should be correct when you run next tests

waghanza (approved):

PR #3550, opened 2020-12-05
[Java] Add quarkus

waghanza:
Hi,

This `PR` adds [quarkus](https://quarkus.io) to this list.

All `java` based **frameworks** runs on [OpenJDK](https://openjdk.java.net) version `11` (**lts**).

Regards,


java/quarkus/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+  </dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-arc</artifactId>
+    </dependency>
```
geoand, line 40: This dependency isn't needed as it is a transitive dependency of almost every quarkus extension

java/quarkus/pom.xml
```diff
@@ -0,0 +1,125 @@
...
+        <version>${quarkus.platform.version}</version>
+        <type>pom</type>
+        <scope>import</scope>
+      </dependency>
+    </dependencies>
+  </dependencyManagement>
+  <dependencies>
+    <dependency>
+      <groupId>io.quarkus</groupId>
+      <artifactId>quarkus-resteasy</artifactId>
```
geoand, line 35: If you opt for version `1.11.0.Beta2` of Quarkus, you can use `quarkus-resteasy-reactive` instead of this. RESTeasy Reactive is much more performant that RESTEasy classic

java/quarkus/config.yaml
```diff
@@ -0,0 +1,11 @@
+framework:
+  website: quarkus.io
+  version: 1.10
+
+build:
+  - mvn clean package -Dquarkus.package.type=uber-jar -Pnative
```
geoand, line 6: The best option is probably use use the `fast-jar` here

java/quarkus/src/main/java/org/acme/BenchmarkResource.java
```diff
@@ -0,0 +1,33 @@
...
+    @Path("/")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String root() {
+        return "";
+    }
+
+    @GET
+    @Path("/user/{id}")
+    @Produces(MediaType.TEXT_PLAIN)
+    public String userId(@PathParam("id") String id) {
```
geoand, line 23: If you use `RESTEasy Reactive` (see comment above), you don't need `@PathParam` at all

java/quarkus/pom.xml
whiplash: Great to here. I'll remove it. To be clear, the `https://code.quarkus.io/` "application generator" was used to bootstrap this.

geoand: Understood, `code.quarkus.io` adds it to just to be on the safe side

java/quarkus/pom.xml
whiplash: Oh nice. I'll wait for that to be out of beta before making this change.

java/quarkus/config.yaml
whiplash: Just found out about this in `https://quarkus.io/guides/maven-tooling`. will have a look.

java/quarkus/src/main/java/org/acme/BenchmarkResource.java
whiplash: Nice!

java/quarkus/pom.xml
geoand: 👍🏼 

That should happen around the 15th of January

waghanza (commented):
is this `PR` ready @whiplash ?

do you mind **convert to draft**, if not

whiplash:
I'd wait for 15th of January

java/quarkus/pom.xml
whiplash: Hey @geoand any news on quarkus 1.11.x?

geoand: It should be released in a few days

PR #3574, opened 2020-12-08
[Java] Update jetty to v11

waghanza:
Hi,

This `PR` updates jetty to v11

Regards,


lukaszlenart (dismissed):
Looks good, just why no to use a parent pom in `java` and put all those version there or at least the shared one

waghanza:
>  just why no to use a parent pom in java

will be awesome, be my guest on that 👍

waghanza:
`sping` fails on https://the-benchmarker.semaphoreci.com/jobs/e51e15de-44ef-46b3-b4a5-59ed8b6bc0ef#L1682 @lukaszlenart

waghanza:
As I understand, when we talk about `spring`, we talk mostly about the **boot** version
@see https://www.baeldung.com/java-in-2019#spring-boot-adoption

Moreover, **TFB** references only **boot** in https://github.com/TechEmpower/FrameworkBenchmarks/tree/master/frameworks/Java/spring

Thus, I'm not sure it is very accurate to have `framework` and `boot` here

What do you think @lukaszlenart @the-benchmarker/web-frameworks ?

/cc @david-m-s

lukaszlenart:
There are two approaches with Spring:
- SpringBoot means you are using an embedded container (I think it's Apache Tomcat)
- Spring to build just a War file which can be deployed to a servlet container

In both cases the same framework is tested, just differ in how the apps are started.

lukaszlenart:
> `sping` fails on https://the-benchmarker.semaphoreci.com/jobs/e51e15de-44ef-46b3-b4a5-59ed8b6bc0ef#L1682 @lukaszlenart

Hm... Jetty v11 is using the new API which is not supported by Spring yet (or by this version of Spring). I don't know if there is a Spring version which supports the new JakartaEE API.

waghanza:
> In both cases the same framework is tested, just differ in how the apps are started.

Then this is not very accurate to have both `here`

I'll leave this issue open and will remove `spring-framework` in something like a week in their is no arguments for keeping it

PR #3590, opened 2020-12-12
[C++] Update drogon to v1.2.0

waghanza:
Hi @an-tao,

This `PR` updates https://github.com/an-tao/drogon to v1.2.0.

Regards,

--------
May I notify you on each `drogon` update here ?

------------
Is `drogon` already (or will become in the near future) setupable via a package manager ?


an-tao:
@waghanza Thanks so much. you are so kind 👍

an-tao (approved):

PR #3613, opened 2020-12-16
Removing Lucky Crystal framework

jwoertink:
This PR removes the Crystal Lucky framework from the benchmarks. I am one of the core maintainers on the Lucky project, and authorize the removal for a few reasons I'll explain here.

While we believe that speed is an important measure for a web framework, especially when many have come up against performance limits, it is not the only measure. Speed is much easier to quantify however, than metrics like maintainability, safety, or developer productivity. Despite these other metrics, we've found that many developers will look primarily at a single number (speed) to determine which framework to use.

It's also difficult to keep up with benchmarks as a small team. We are not the fastest web framework, and likely never will be _the_ fastest. But we know Lucky is fast enough for most applications, and in line with most other Crystal web frameworks. Instead of trying to seek out small speed boosts, we would rather focus our efforts on making Lucky better holistically.

However, if we don't dedicate the time needed to keep Lucky configured correctly, new developers interested in Lucky ask why we are one of the "slowest frameworks". In reality, this is not a reflection of the speed of Lucky, but of the effort put into configuring it correctly for these benchmarks. It leads to negative press for us, which we believe is misleading and is certainly frustrating as we continue to explain the issue to newcomers.

We also believe that the configuration of Lucky for these benchmarks doesn't indicate a realistic scenario for the framework's usage. Most of Lucky is stripped out to make the benchmark function. For example, this doesn't show time to render a homepage using HTML markup, or Database Queries, or how JSON serialization works in an API. The logging is removed, and the middleware is stripped down.

Finally, we're still quite a ways out from a 1.0.0 release. Each current release usually contains many breaking changes which will cause any number of fluctuations in benchmarking to occur.

We appreciate the effort the maintainers have put in to making these benchmarks and allowing people to see how things compare, and we would be open to adding it back at a later time.

Thank you!


waghanza:
Hi @jwoertink,

> This PR removes the Crystal Lucky framework from the benchmarks. I am one of the core maintainers on the Lucky project, and authorize the removal for a few reasons I'll explain here.
 
If you want to remove it, we (mostly me) are not going to oppose ❤️ 

However, let me explain the plan (I have to create a clear roadmap, but I've not enough bandwith now).

The idea is not to compare only speed (it will be explain in the README),. The long term goal is to have a comparison of frameworks with :
+ speed
+ build time
+ resource consumption
+ ...

I am totally aware that speed is not the only indicator for a framework, and other indicators will come (but I'm the only core developer on this project, so I understand any frustration about have `lucky` listed here)

/cc @paulcsmith

paulcsmith:
That sounds like a fantastic direction @waghanza! We're just spread super thing so need to take this off our plate for now. We can definitely revisit in the future though! Thanks for all your hard work!

paulcsmith (approved):

jwoertink:
@waghanza yeah, that direction will be super great! And I totally get not having the bandwidth. We can definitely appreciate all of the work you've put in to this. Thanks for taking a look!

waghanza:
@paulcsmith @jwoertink I am not sure I understand.

You want to remove `lucky` ?

jwoertink:
@waghanza Correct. We move forward with this PR to remove Lucky, and then we revisit adding it back in at a later time.

waghanza (approved):
Then so be it 🎉  

You always be welcome `here` ❤️

jwoertink:
Thank you! We really appreciate it! ❤️

PR #3617, opened 2020-12-17
[Python] Update blacksheep to 0.3

waghanza:
Hi @RobertoPrevato,

This `PR` updates `blacksheep` to **0.3**

Regards,

-----
May I notify you on each update that affect `blacksheep` here ?

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -24,6 +24,8 @@ blocks:
 - name: pony
   dependencies:
   - setup
+  run:
+    when: change_in('/pony/')
   task:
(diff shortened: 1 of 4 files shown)


RobertoPrevato:
Hi @waghanza,
Thank You! Yes you can notify me. 
The last release is important because it adds `whells` to the distribution package, so installing the package should take less time also in your CI builds (except for Python 3.8 on Windows 10 because there is an issue in Python 3.8 that prevents using the wheel).
In the next days I will release other improvements to my framework and probably will release a new version.

Best Regards,
Roberto

PR #3618, opened 2020-12-17
revert-psr7-implementation-for-php8-support

dominikzogg:


waghanza:
@dominikzogg I am not able to choose which implementation we choose use in `here` depending on the version of `php` in use.

Which one do you recommend for a **production grade** app ? I mean I prefer promoting your framework depending on **YOUR** recommandation

dominikzogg:
@waghanza within my skeleton i suggest the usage of the slim psr7 implementation https://github.com/chubbyphp/petstore/blob/chubbyphp/composer.json#L29 The reason why i changed that in the first place was to have a better comparison against `sunrise-router` using its implementation

waghanza:
I understand why then 😄

But in order to have meaningful results, I prefer that each implementation is **production-grade** (I mean you can safely use the same components for a real life app).

I'll add that on `README`.

Also, I'll close / recreate your PR (due to CI) ❤️

PR #3626, opened 2020-12-18
[Java] Remove spring framework

### LINKED ISSUE #3623: [Java] Remove spring-framework
Hi,

`spring-framework` and `spring-boot` tests the same framework under the hood.

As the most common use case is to use `spring-boot`, `spring-framework` needs to be removed.

Regards,

waghanza:
Hi,

This `PR` removes spring-framework from this list, and then keep only `boot`.

Since it is the same framework, no need to test both.

Regards,

------
Closes #3623


lukaszlenart (approved):
LGTM 👍

PR #3632, opened 2020-12-20
Update results

waghanza:
@lukaszlenart Is the command line in https://github.com/the-benchmarker/web-frameworks/blob/f494f073e1ec3d608848aa8b48005c67fb4d14b0/java/struts2/config.yaml#L11 still available ?


laylatichy:
@waghanza is slack workspace open to join? That one mentioned in readme

lukaszlenart:
It's valid, but looks like some APIs have changed in Jetty v11 and now the embedded app won't start, need to figure out a proper approach.

lukaszlenart:
I need to revert the Jetty upgrade, Struts 2.5.x won't support the latest JakaratEE Servlet API - I planned add support for it (if possible) in Struts 2.6

PR #3643, opened 2020-12-23
[Scala]: Add Finch

whiplash:
Scala Finch added.

https://finagle.github.io/finch/


waghanza:
Hi @sergeykolbasov @vkostyukov @rpless

Would you mind to add (or consent that we add) [finch](https://github.com/finagle/finch) in `here` ?

Regards,

PR #3645, opened 2020-12-24
[C++] Add nawa

jatofg:
Add the nawa framework (issue #3634). As mentioned in the issue, I was unable to test whether the integration into the benchmark suite works. Feel free to adapt the PR as you like.

To me, it looks like the suite is run on Ubuntu 18.04, so I've set the dependencies in config.yaml accordingly. Unfortunately, Ubuntu doesn't provide a meta package for the most recent versions of libboost-system and libboost-thread, so if it were to be run on another OS, the following binary dependencies would have to be adapted (by setting the version accordingly):
* libboost-system1.65.1 (would be libboost-system1.71.0 on 20.04)
* libboost-thread1.65.1 (would be libboost-thread1.71.0 on 20.04)

### DIFF
--- a/.semaphore/semaphore.yml
+++ b/.semaphore/semaphore.yml
@@ -919,6 +919,9 @@ blocks:
       commands:
       - cd cpp/evhtp && make build  -f .Makefile  && cd -
       - FRAMEWORK=cpp/evhtp bundle exec rspec .spec
+    - name: nawa
+      commands:
+      - cd cpp/nawa && make build  -f .Makefile  && cd -
 - name: swift
   dependencies:
   - setup
--- a/cpp/nawa/CMakeLists.txt
+++ b/cpp/nawa/CMakeLists.txt
@@ -1,0 +1,14 @@
+cmake_minimum_required(VERSION 3.13)
+project(nawa_benchmark)
+
+set(CMAKE_CXX_STANDARD 17)
+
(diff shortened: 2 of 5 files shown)


waghanza (commented):
Thanks for your contribution. I'll help next week

waghanza (commented):
We use alpine for containers. I'll create an other PR from this one (due to CI restrictions I set I'm currently the only one able to trigger it)

waghanza (commented):
Two more things @jatofg if you mind

+ Could you confirm that logging is disabled ? This kind of stuff could mess-up results
+ Is this suitable for production ? I mean in order to have meaningful results, we need to ensure the same quality for each implementation

jatofg:
Hi,
* there should be no logging, unless an error occurs (no config file provided / wrong path, socket cannot be opened, ...)
* what do you mean with "suitable for production", the implementation of the challenge? I've tested it (manually) and haven't encountered any issues, it just does what it's supposed to (also, I don't see what could go wrong, as the logic of the challenge is quite simple).

waghanza:
Hi @jatofg,

> what do you mean with "suitable for production"

I meant having a app that you would safely deploy in production. But you can ignore this comment (I think), it is more accurate for interpreted language.

For example, the comment I made is to make sure the service could be safely deployable (ensure that the app won't crash - due to a memory leak in ruby as example)

PR #3649, opened 2020-12-25
add(javascript): add nanoexpress

dalisoft:
As i got some time to maintain nanoexpress, i decided make project live again as it's ajv+swagger built-in feature in many my own apps makes backend building easier.

@waghanza What do you think about adding nanoexpress to list again?

Thanks.
Regards,
Davlik

### DIFF
--- a/javascript/nanoexpress/config.yaml
+++ b/javascript/nanoexpress/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: nanoexpress/nanoexpress
-  version: 2.4.4
+  version: 2.4
 
 environment:
   EXPERIMENTAL_FASTCALL: 1
--- a/javascript/nanoexpress/config.yaml
+++ b/javascript/nanoexpress/config.yaml
@@ -1,0 +1,12 @@
+framework:
+  github: nanoexpress/nanoexpress
+  version: 2.4.4
+
+environment:
+  EXPERIMENTAL_FASTCALL: 1
+
+build_deps:
+  - git
+
+fixes:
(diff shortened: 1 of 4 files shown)


waghanza:
You are more than welcome @dalisoft

dalisoft:
Thanks @waghanza

javascript/nanoexpress/config.yaml
```diff
@@ -0,0 +1,12 @@
+framework:
+  github: nanoexpress/nanoexpress
+  version: 2.4.4
```
waghanza, line 3: Could you use `X.Y` format ?

javascript/nanoexpress/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+
+environment:
+  EXPERIMENTAL_FASTCALL: 1
```
waghanza, line 6: Is this production ready with uWebSocket ?

waghanza (changes requested):

javascript/nanoexpress/config.yaml
dalisoft: Yes, it is production ready. With this env variable performance boost sometimes improves by 20% which is huge for me

javascript/nanoexpress/config.yaml
dalisoft: Like `2.x.y`?

waghanza: like `2.4` because 
+`~2.4.4` in `package.json` could setup `2.4.8`
+ `2.4` is clearer for this project audience (I think)

dalisoft: I fixed it, thanks

PR #3675, opened 2020-12-28
[Scala] Add finch

waghanza:
Hi @whiplash,

This is a relocation of you PR  #3643.

Regards,

------
Closes #3643


scala/finch/src/main/scala/yolo/Main.scala
```diff
@@ -0,0 +1,34 @@
...
+  }
+
+  val postUser = post("user") {
+    okEmpty
+  }
+
+  val getUserName = get("user" :: path[String]) { name: String => Ok(name) }
+
+  val api: Service[Request, Response] =
+    Bootstrap
```
sergeykolbasov, line 25: I believe most users construct a single endpoint before bootstrapping it:

```scala
Bootstrap
      .serve[Text.Plain](postUser :+: getUserName :+: root)
      .toService
```

whiplash (dismissed):

whiplash (approved):

scala/finch/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.4.4
```
waghanza, line 1: do we really need this file @whiplash ?

scala/finch/project/plugins.sbt
```diff
@@ -0,0 +1 @@
+addSbtPlugin("com.typesafe.sbt" % "sbt-native-packager" % "1.7.3")
```
waghanza, line 1: can we use a less strict constraint ? I mean the promise here, is to ensure having a specific version of `finch`, we probably do not need to specify version of `sbt-native-packager` (we can use the latest, I think) ?


the think is to have the same build tool for each `scala` based frameworks

scala/finch/project/build.properties
whiplash: It can be 1.4.6 (see https://github.com/the-benchmarker/web-frameworks/blob/master/scala/Dockerfile#L1)

scala/finch/project/plugins.sbt
whiplash: Latest is now 1.8.0... we can possibly use the usual "range" thing I guess. can you make that change?

scala/finch/project/build.properties
waghanza: can not it be removed ? I mean the docker image contains already an `sbt` version

scala/finch/project/plugins.sbt
waghanza: my question was more to avoid pin this dependency
for example, by using http://ant.apache.org/ivy/history/latest-milestone/ivyfile/dependency.html#revision

whiplash: and I replied that I agree... we can set that plugin version something like `[1.7,2.0]`

waghanza: let me rephrase : is there a way to remove this version ? or tell **sbt** to use the last available (not depending on a range) ?

whiplash: by reading [here](https://ant.apache.org/ivy/history/2.3.0/ivyfile/dependency.html#revision) I think we can "try" with `latest.integration`

waghanza: seems 🆗

waghanza: but `sbt` warns about
```
[warn] There may be incompatibilities among your library dependencies; run 'evicted' to see detailed eviction warnings.
```

PR #3680, opened 2020-12-29
[Java] Add undertow

waghanza:
-----
Closes #3677

### DIFF
--- a/java/undertow/config.yaml
+++ b/java/undertow/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: undertow.io
-  version: 2.0.33.Final
+  version: 2.0
 
 build:
   - mvn package
--- a/java/undertow/config.yaml
+++ b/java/undertow/config.yaml
@@ -1,0 +1,14 @@
+framework:
+  website: undertow.io
+  version: 2.0.33.Final
+
+build:
+  - mvn package
+
+files:
+  - target/server-jar-with-dependencies.jar
+
+binaries:
+  - target/server-jar-with-dependencies.jar
+
+command: /usr/bin/java -XX:-RelaxAccessControlCheck -XX:+UseNUMA -XX:+UseParallelGC -XX:+AggressiveOpts -jar /opt/web/target/server-jar-with-dependencies.jar
--- a/java/undertow/pom.xml
+++ b/java/undertow/pom.xml
@@ -12,7 +12,7 @@
   <name>benchmark</name>
 
   <properties>
-    <undertow.version>2.0.33.Final</undertow.version>
+    <undertow.version>[2.0,2.1)</undertow.version>
     <maven.compiler.source>11</maven.compiler.source>
     <maven.compiler.target>11</maven.compiler.target>
     <maven.compiler.parameters>true</maven.compiler.parameters>
--- a/java/undertow/pom.xml
+++ b/java/undertow/pom.xml
@@ -1,0 +1,68 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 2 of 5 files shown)


whiplash (dismissed):

java/undertow/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>benchmark.undertow</groupId>
+  <artifactId>benchmark</artifactId>
+  <version>1.0.0</version>
+
+  <name>benchmark</name>
+
+  <properties>
+    <undertow.version>2.0.33.Final</undertow.version>
```
waghanza, line 15: probably rather user `[2.0,.1)` to be sure we get the last patch of 2.0

java/undertow/config.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+  website: undertow.io
+  version: 2.0.33.Final
```
waghanza, line 3: please use 2.0 instead

waghanza:
Hi @fl4via,

Would you mind to add (or consent that we add) [undertow](https://github.com/undertow-io/undertow) in `here` ?

------------

Regards,

java/undertow/src/main/java/the/benchmarker/Main.java
```diff
@@ -0,0 +1,44 @@
...
+import io.undertow.util.Headers;
+
+import java.util.Deque;
+import java.util.Optional;
+
+public class Main {
+
+    public static void main(final String[] args) {
+        Undertow server = Undertow.builder()
+                .addHttpListener(3000, "localhost")
```
waghanza, line 16:
```suggestion
                .addHttpListener(3000, "0.0.0.0")
```

mfirry (dismissed):

whiplash (dismissed):

java/undertow/src/main/java/the/benchmarker/Main.java
```diff
@@ -0,0 +1,44 @@
...
+                            public void handleRequest(HttpServerExchange exchange) throws Exception {
+                                exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "text/plain");
+                                exchange.getResponseSender().send("");
+                            }
+                        })
+                        .get("/user/{name}", new HttpHandler() {
+                            @Override
+                            public void handleRequest(HttpServerExchange exchange) throws Exception {
+                                exchange.getResponseHeaders().put(Headers.CONTENT_TYPE, "text/plain");
+                                var x = Optional.ofNullable(exchange.getQueryParameters().get("name"))
```
stuartwdouglas, line 36: Note that using lambdas and functional programming here will be slower than just manually checking for null.

java/undertow/pom.xml
```diff
@@ -0,0 +1,68 @@
...
+
+  <dependencies>
+
+    <dependency>
+            <groupId>io.undertow</groupId>
+            <artifactId>undertow-core</artifactId>
+            <version>${undertow.version}</version>
+    </dependency>
+    <dependency>
+            <groupId>io.undertow</groupId>
```
stuartwdouglas, line 30: You don't need Servlet or websocket dependencies, as you are not using them.

java/undertow/src/main/java/the/benchmarker/Main.java
whiplash: I'm curious to know more about how this "littlle" thing can slow down this endpoint (curious in general, not for this specific case). Care to elaborate further?

stuartwdouglas: Basically just extra allocations when you don't really need them. Undertow has been optimised to avoid allocation as much as possible, so for these super simple style tests extra allocation inside the handler may make a noticeable difference. 

It's less of an issue in a 'real' app because the app generally generates a lot more garbage doing business logic, so a few extra allocations are unnoticeable.

PR #3688, opened 2020-12-30
[Java] Add jersey

waghanza:
Hi @jansupol,

Would you mind to add (or consent that we add) [jersey](https://github.com/eclipse-ee4j/jersey) in `here` ?

------------

Regards,


jansupol:
Hi @waghanza,
Eclipse Jersey is an open-source project licensed under EPL-2.0 + GPL-2.0 with CPE. As long as the license is honored, it can be used without extra permission.

Personally, I'd love to see as many usages of Jersey in benchmarks as possible. Potentially, it can find a possible weak spot in the implementation and help to improve it. 

I only can tell: Thank you for taking Jersey into consideration.

waghanza:
Hi @jansupol,

Thanks for this kinds word. Indeed, most of frameworks (all :-)) are opensource (with various license), and thus we can integrate it `here`.

However, there is 3 points that tell me explicit consent (or at least no opposition) is for the good all of us :
+ pay respect to author(s) / maintainer(s), someones don't want to see their apps included here -> which we accept even we do not understand
+ including it here could be out of subject (framework not maintained anymore, not target the scenario here ...)
+ having a review from people involved on creation is 👍 

I will merge this and ping you on results update tomorrow

PR #3690, opened 2020-12-30
[Rust] Fix gotham

waghanza:


rust/gotham/Cargo.toml
```diff
@@ -6,7 +6,7 @@ authors = ["Isaac Whitfield <iw@whitfin.io>"]
 [dependencies]
 gotham = "~0.5"
 gotham_derive = "~0.5"
-hyper = "*"
+hyper = "~0.13"
```
msrd0, line 9: I strongly recommend you remove this dependency and use the `gotham::hyper` re-export instead

waghanza: and is it the same for `serde` ?

msrd0: I'd recommend not using `*` as the version, but the serde version is independent from gotham, so it shouldn't be that much of a problem.

waghanza: so what do you recommend ? using a range constraint like `~1.0` ?

msrd0: Just put `serde = "1.0"`, cargo will treat that as `>= 1.0, < 2.0` so that should be fine

PR #3753, opened 2021-01-08
[PHP] Use libevent for workerman

waghanza:
Hi @walkor,

As I understand, it safer to use something like `libevent` when deploying `workerman` on **production**, right ?

Regards,

-------
Thanks for the tips @leocavalcante

### DIFF
--- a/php/Dockerfile
+++ b/php/Dockerfile
@@ -5,22 +5,22 @@ RUN docker-php-ext-install zip opcache
 
 WORKDIR /usr/src/app
 
+{{#build_deps}}
+  RUN apk add {{{.}}}
+{{/build_deps}}
+
+{{#php_mod}}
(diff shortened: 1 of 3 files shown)


leocavalcante:
Np ;)
More info:
Screen Shot 2021-01-08 at 15 57 47
https://l2hl2j3scgnnw5mujftxyollfe--doc-workerman-net.translate.goog/install/requirement.html

waghanza:
I suppose that all frameworks that use `workerman` requires also that configuration, https://github.com/the-benchmarker/web-frameworks/tree/master/php/chubbyphp-workerman for example

dominikzogg (approved):

walkor:
> As I understand, it safer to use something like libevent when deploying workerman on production, right ?

Yes. Use event extension for better performance.

PR #3754, opened 2021-01-08
[Clojure] Update luminus to 3.91

### LINKED ISSUE #3638: [Clojure] Update Luminus to 3.91

waghanza:
Hi @yogthos,

This `PR` update luminus to 3.91.

Could you confirm that we are using 3.91 ?

Could also review this ? I do not know about clojure but `project.cjl` seems to list a lot of dependencies, shouldn't this deps handle by luminus-template ?

Regards,

------
Closes #3638 

------
Without `[com.fasterxml.jackson.core/jackson-core "2.12.0"]` I have a `com.fasterxml.jackson.core.util.JacksonFeature` at compile time

### DIFF
--- a/clojure/luminus/config.yaml
+++ b/clojure/luminus/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: luminusweb.com
-  version: 1.0
+  version: 3.91
 
 image: clojure:openjdk-11-lein-slim-buster
 
--- a/clojure/luminus/project.clj
+++ b/clojure/luminus/project.clj
@@ -1,6 +1,7 @@
 (defproject luminus "0.1.0-SNAPSHOT"
 
-  :dependencies [[ch.qos.logback/logback-classic "1.2.3"]
+  :dependencies [[luminus/lein-template "3.91"]
+                 [ch.qos.logback/logback-classic "1.2.3"]
                  [cheshire "5.10.0"]
                  [clojure.java-time "0.3.2"]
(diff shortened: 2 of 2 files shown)


yogthos:
Hi,

The template simply generates an instance of the project, and `project.clj` is used to manage dependencies in the project. You don't want to remove selmer dependency from it as it's needed for HTML templating. Otherwise looks fine.

waghanza:
I thought the template will requires it as it is specified in https://github.com/luminus-framework/luminus-template/blob/26214411b812701f144a252e82fb48aa464371f9/project.clj#L9

yogthos:
The template is basically a program that generates the project given the command line arguments. Selmer is used internally by the template engine to inject runtime variables into the files it produces. The generated project is completely unrelated to it, and it uses its own `project.clj` file to track its dependencies.

waghanza:
The challenge here does not require html templating, btw

I think we could skip this step (the idea is to have the as-minimalist-as implementation we can (for exemple, to be read by newcomers).

Moreover, even with `selmer` we need to explicitly as `jackson-core`

yogthos:
We'd need to remove references to it from the code in the project, but yeah that wouldn't be hard to do. I could take a look.

waghanza:
> We'd need to remove references to it from the code in the project, but yeah that wouldn't be hard to do. I could take a look.

It will be awesome if you take it to minimal

PR #3764, opened 2021-01-10
Update to Symfony 5.2

Nyholm:
I use the files that comes from a standard symfony 5.2 application.


php/symfony/config.yaml
```diff
@@ -1,14 +1,11 @@
...
   - nginx
 
-environment:
-  APP_ENV: prod
-
 before_command:
   - mkdir -p var/log && chown www-data var/log
   - mkdir -p var/cache && chown www-data var/cache
-  - php bin/console cache:warmup
+  - php bin/console cache:warmup --no-debug
```
waghanza, line 11: I'm not sure we need a warmup here, this step is done with wrk 
https://github.com/the-benchmarker/web-frameworks/blob/f7f7984535d7dd43f0e034bdadc679ce67dc3fb4/lib/tasks/collect.rake#L36

php/symfony/composer.json
```diff
@@ -1,9 +1,9 @@
 {
   "require": {
-    "symfony/framework-bundle": "~5.1.0",
+    "symfony/framework-bundle": "^5.2",
```
waghanza, line 3: This could install 5.3
@see https://getcomposer.org/doc/articles/versions.md#caret-version-range-
and we must use the same version in `config.yaml`

I prefer using `~5.2.0` (not sure this is the standard in php word, btw)

php/symfony/config.yaml
Nyholm: Thank you for the review. 

Is that line running a http request before we actually start measuring the responses? If so, yes, you might be correct if docker allows me to write to the filesystem. 

This line is best practice to always run before you ship your project to production. I'll prefer to keep it.

php/symfony/composer.json
Nyholm: That is correct. I thought this was a feature =)

I'll revert.

php/symfony/config.yaml
waghanza: We can keep it, yep. However, `symfony` is the only (or one of) framework that have a built-in console and a warm up command. The above line is doing the same think (i think), it makes several HTTP requests so a all frameworks are warn for the real measurement

waghanza (approved):

Nyholm:
Thank you for merging.

PR #3783, opened 2021-01-15
Update simps

sy-records:


waghanza:
Thanks for this @sy-records. Could you make any recommendations about settings ? I mean any settings you could be confident to set for a `swoole` based app ?

/cc @leocavalcante @huangzhhui @limingxinleo @lizhichao

sy-records:
Hi @waghanza,I should have said before. We can set 'worker_num' and 'enable_coroutine'.
```
 'worker_num' => swoole_cpu_num() * 2,
 'enable_coroutine' => false,
```

Other configuration items parameters can be set according to the needs of the framework and do not have to be unified.

waghanza:
> Hi @waghanza,I should have said before. We can set 'worker_num' and 'enable_coroutine'.
> 
> ```
>  'worker_num' => swoole_cpu_num() * 2,
>  'enable_coroutine' => false,
> ```
> 
> Other configuration items parameters can be set according to the needs of the framework and do not have to be unified.

Thanks, will check ...

waghanza:
Ok for `worker_num`. However, I can not see the default value of `enable_coroutine` in https://www.swoole.co.uk/docs/modules/swoole-server/configuration

Does it need to be overwritten ?

sy-records:
The default is true. let's close it.

waghanza:
> The default is true. let's close it.

Ok. I think this documentation is missing on the official doc, btw

We also have to disable logs. I think redirecting logs to `/dev/null` and  setting log level to `TRACE` will be enough

sy-records:
> Ok. I think this documentation is missing on the official doc, btw

image

leocavalcante (approved):

sy-records:
goto #3786

PR #3791, opened 2021-01-16
[PHP] add antidot react framework

kpicaza:
Would you mind to add (or consent that we add) Antidot React in here ?

GET /, SHOULD return a successful status code and an empty string
GET /user/:id, SHOULD return a successful status code and the id
POST /user, SHOULD return a successful status code and an empty string
Regards,


php/antidot/composer.json
```diff
@@ -0,0 +1,87 @@
...
+  },
+  "require-dev": {
+    "franzl/whoops-middleware": "^2.0",
+    "phpro/grumphp": "^1.0",
+    "phpstan/phpstan": "^0.12",
+    "phpunit/phpunit": "^8.0|^9.0",
+    "roave/security-advisories": "dev-master",
+    "squizlabs/php_codesniffer": "^3.4",
+    "symfony/var-dumper": "^5.2",
+    "laminas/laminas-component-installer": "^2.1.2"
```
sy-records, line 41: dev dependencies are not necessary and are recommended to be removed

`php/antidot/public/favicon.ico` This one is also

sy-records (changes requested):

php/antidot/composer.json
kpicaza: Thanks! maybe will be correct to commit the cached config to avoid generating it in the first request?

waghanza: not needed. the warm-up phase will do this for you

php/antidot/composer.json
```diff
@@ -0,0 +1,61 @@
...
+    "antidot-fw/cli": "^1.1.0",
+    "antidot-fw/container": "^0.1.0",
+    "antidot-fw/dev-tools": "^0.1.2",
+    "antidot-fw/event-dispatcher": "^2.0.3",
+    "antidot-fw/framework": "^0.1.1",
+    "antidot-fw/logger": "^1.1.0",
+    "antidot-fw/symfony-config-translator": "^1.1.0",
+    "antidot-fw/yaml-config-provider": "^0.1.0",
+    "webmozart/assert": "^1.7.0",
+    "wshafer/psr11-monolog": "^3.0.0|@dev"
```
waghanza, line 31: do we need to specify each version ?
I mean is their not a main package, like `antidot-fw/framework` which require all components ?

waghanza: also I prefer using tilde ranges, to ensure that version is in-sync with what we show in `README.md`

php/antidot/composer.json
```diff
@@ -0,0 +1,61 @@
...
+        "antidot-fw/tactician-adapter",
+        "antidot-fw/message-queue",
+        "antidot-fw/cli",
+        "antidot-fw/fast-router-adapter",
+        "antidot-fw/event-dispatcher",
+        "antidot-fw/symfony-config-translator",
+        "wshafer/psr11-monolog",
+        "antidot-fw/antidot-react-http",
+        "antidot-fw/antidot-react-psr15"
+      ]
```
waghanza, line 58: is this required for on production ?

php/antidot/config.yaml
```diff
@@ -0,0 +1,9 @@
+framework:
+  website: antidotfw.io
+  version: 0.1.3
```
waghanza, line 3: better use `X.Y` format, clearer and requires less maintenance

php/antidot/config/services/dependencies.prod.yaml
```diff
@@ -0,0 +1,22 @@
...
+
+  some.command:
+    class: App\Application\Command\SomeCommandExample
+    tags:
+      - { name: 'console.command', command: 'my:custom:command' }
+
+parameters:
+  debug: false
+  config_cache_enabled: true
+  monolog:
```
waghanza, line 14: could the log been disabled ? this kind of operations could mess-up the results

php/antidot/var/cache/config-cache.php
```diff
@@ -0,0 +1,120 @@
+<?php
```
waghanza, line 1: is this file suitable for production ?

waghanza (changes requested):
Thanks for you contribution 🎉 

Please note that the implementation should be as-minimal-as it can to run on production

php/antidot/var/cache/config-cache.php
kpicaza: Is an auto-generated cache, not needed, I will remove it. When I apply al suggested changes I will request another revision, thank you for your help,- D

php/antidot/composer.json
kpicaza: Is the standard installation, which adds some functionality like event dispatcher, logger, cli, dev-tools... I will update the PR accordingly using only the minimal installation, thanks for the advice;- D

php/antidot/composer.json
```diff
@@ -0,0 +1,30 @@
...
+  "license": "BSD-2-Clause",
+  "authors": [
+    {
+      "name": "kpicaza"
+    }
+  ],
+  "require": {
+    "php": "^7.4|^8.0",
+    "ext-json": "*",
+    "antidot-fw/react-framework": "~1.0",
```
waghanza, line 14: should be on `X.Y.Z` to guarantee that we keep staying at `X.Y` (if no lock file)

waghanza (approved):

PR #3793, opened 2021-01-17
Rust with musl has a performance issue

octaltree:
https://github.com/rust-lang/rust/issues/70108
musl-libc seems to be slow when running multithreaded.
Is it a regulation every executable is portable?
I recommend to use x86_64-unknown-linux-gnu.

my environment: Linux version 5.10.2-2-MANJARO (builduser@LEGION) (gcc (GCC) 10.2.0, GNU ld (GNU Binutils) 2.35.1) #1 SMP PREEMPT Tue Dec 22 08:14:42 UTC 2020

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -2,21 +2,8 @@ FROM rust:1.49 AS build
 
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get -y install musl-tools --no-install-recommends \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
-
-RUN apt-get -y install musl-tools --no-install-recommends
-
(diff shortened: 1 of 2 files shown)


waghanza:
Thanks for notice @octaltree.

I understand your concern, but having only performances indicator in not the purpose here. I mean the initial purpose of this project is to reflect some indicators (performance is one of them) with best practices and conventions.

The conventions when building containers is to use `alpine`

octaltree:
>The conventions when building containers is to use alpine

Now that I understand the convention, I'll close this pull request.

https://w3techs.com/technologies/history_details/os-linux
Alpine is not used for the production server (it was CentOS a while ago), so why not change the running server?

PR #3797, opened 2021-01-17
Add basic vertx4 (core) and vertx4-web

### LINKED ISSUE #441: vertx java framework
+ Language : Java
+ Repo : https://vertx.io/ https://github.com/eclipse-vertx/vert.x
+ Author : https://github.com/eclipse-vertx

elliottback:
Closes #441


java/vertx4/config.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+  website: https://vertx.io/docs/vertx-core/java/
```
waghanza, line 2: no need to have `https://`
```suggestion
  website: vertx.io/docs/vertx-core/java/
```

java/vertx4/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  version: 4.0.0
```
waghanza, line 3: I prefer `X.Y` format
```suggestion
  version: 4.0
```

java/vertx4/pom.xml
```diff
@@ -0,0 +1,77 @@
...
+    <groupId>benchmark.vertx4</groupId>
+    <artifactId>benchmark</artifactId>
+    <version>1.0.0</version>
+    <packaging>jar</packaging>
+    <name>vertx4</name>
+    <description>Demo project for Vertx</description>
+
+    <properties>
+        <java.version>11</java.version>
+        <vertx.version>4.0.0</vertx.version>
```
waghanza, line 15: better use a range, to ensure we test the last version (as specify on `config.yaml`)
```suggestion
        <vertx.version>[4.0,4.1)</vertx.version>
```

java/vertx4web/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+build:
+  - mvn package
+
+files:
+  - target/server-jar-with-dependencies.jar
+
+binaries:
+  - target/server-jar-with-dependencies.jar
+
+command: /usr/bin/java -XX:-RelaxAccessControlCheck -XX:+UseNUMA -XX:+UseParallelGC -XX:+AggressiveOpts -jar /opt/web/target/server-jar-with-dependencies.jar
```
waghanza, line 14: could you explain those options ? the fact is I'm not a `java` dev trying to review `java` (I think it will be for the better if all jvm options in use, I mean for each frameworks, are the same)

java/vertx4web/src/main/java/benchmark/vertx4web/BenchmarkApplication.java
```diff
@@ -0,0 +1,22 @@
...
+public class BenchmarkApplication {
+    public static void main( String [] args )
+    {
+        Vertx vertx = Vertx.vertx();
+
+        Router router = Router.router(vertx);
+
+        router.get("/").respond( ctx -> ctx.response().setStatusCode(200).end());
+        router.post("/user").respond( ctx -> ctx.response().setStatusCode(200).end());
+        router.get("/user/:userId").respond( ctx -> ctx.response().setStatusCode(200).end(ctx.pathParam("userId")));
```
waghanza, line 16: could this be chained (I think it's clearer)
```suggestion
        router.
          get("/").respond( ctx -> ctx.response().setStatusCode(200).end())
          post("/user").respond( ctx -> ctx.response().setStatusCode(200).end())
          get("/user/:userId").respond( ctx -> ctx.response().setStatusCode(200).end(ctx.pathParam("userId")));
```

waghanza (changes requested):

java/vertx4/pom.xml
elliottback: Vert.x 4.1 does not exist

java/vertx4/config.yaml
elliottback: URLs should have schemes.

java/vertx4web/config.yaml
elliottback: They're taken from another java benchmark; unifying them to a property in /.../java seems reasonable but not in the scope of this PR.

java/vertx4web/src/main/java/benchmark/vertx4web/BenchmarkApplication.java
elliottback: It cannot, the API is not chainable: https://vertx.io/docs/apidocs/io/vertx/ext/web/Router.html#get--

elliottback:
I don't intend to spend any additional time in this space, so please feel free to use the above as a sketch for incorporating basic Vert.x benchmarks into this framework.

java/vertx4/pom.xml
waghanza: `[4.0,4.1)` means >= 4.0 and <= 4.1

waghanza:
could you repoen this PR, at least to incorporate you commits (I'll do the next stuffs)

PR #3798, opened 2021-01-17
Laravel v8

barryvdh:
- Upgrade to Laravel 8
- Remove default middleware for session/cookies etc
- Cache the config + routes as described in https://github.com/the-benchmarker/web-frameworks/issues/3393

Fixes https://github.com/the-benchmarker/web-frameworks/issues/3393

### DIFF
--- a/php/laravel/app/Http/Kernel.php
+++ b/php/laravel/app/Http/Kernel.php
@@ -14,11 +14,11 @@ class Kernel extends HttpKernel
      * @var array
      */
     protected $middleware = [
-        \App\Http\Middleware\TrustProxies::class,
-        \App\Http\Middleware\CheckForMaintenanceMode::class,
-        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
-        \App\Http\Middleware\TrimStrings::class,
-        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+//        \App\Http\Middleware\TrustProxies::class,
+//        \App\Http\Middleware\CheckForMaintenanceMode::class,
+//        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
+//        \App\Http\Middleware\TrimStrings::class,
+//        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
     ];
 
     /**
@@ -28,13 +28,13 @@ class Kernel extends HttpKernel
      */
     protected $middlewareGroups = [
         'web' => [
-            \App\Http\Middleware\EncryptCookies::class,
+            //\App\Http\Middleware\EncryptCookies::class,
             //\Illuminate\Cookie\Middleware\AddQueuedCookiesToResponse::class,
-            \Illuminate\Session\Middleware\StartSession::class,
+            //\Illuminate\Session\Middleware\StartSession::class,
             // \Illuminate\Session\Middleware\AuthenticateSession::class,
-            \Illuminate\View\Middleware\ShareErrorsFromSession::class,
+            //\Illuminate\View\Middleware\ShareErrorsFromSession::class,
             //\App\Http\Middleware\VerifyCsrfToken::class,
-            \Illuminate\Routing\Middleware\SubstituteBindings::class,
+            // \Illuminate\Routing\Middleware\SubstituteBindings::class,
         ],
 
         'api' => [
--- a/php/laravel/artisan
+++ b/php/laravel/artisan
@@ -1,0 +1,53 @@
+#!/usr/bin/env php
+<?php
+
+define('LARAVEL_START', microtime(true));
+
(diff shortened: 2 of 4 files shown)


php/laravel/app/Http/Kernel.php
```diff
@@ -14,11 +14,11 @@ class Kernel extends HttpKernel
...
-        \App\Http\Middleware\TrustProxies::class,
-        \App\Http\Middleware\CheckForMaintenanceMode::class,
-        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
-        \App\Http\Middleware\TrimStrings::class,
-        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
+//        \App\Http\Middleware\TrustProxies::class,
+//        \App\Http\Middleware\CheckForMaintenanceMode::class,
+//        \Illuminate\Foundation\Http\Middleware\ValidatePostSize::class,
+//        \App\Http\Middleware\TrimStrings::class,
+//        \Illuminate\Foundation\Http\Middleware\ConvertEmptyStringsToNull::class,
```
waghanza, line 21: I think it's better to remove this instead of commenting (to keep code as minimal as possible)

php/laravel/artisan
```diff
@@ -0,0 +1,53 @@
+#!/usr/bin/env php
```
waghanza, line 1: do we really need this file in production ? (except for warming caches)

php/laravel/composer.json
```diff
@@ -1,8 +1,6 @@
 {
   "require": {
-    "fideloper/proxy": "*",
-    "laravel/framework": "~7.27.0",
-    "laravel/tinker": "*"
+    "laravel/framework": "^8.12"
```
waghanza, line 3: Does this constraint ensure we are using 8.12 ?
```suggestion
    "laravel/framework": "~8.22.0"
```

php/laravel/config.yaml
```diff
@@ -17,3 +17,6 @@ before_command:
   - find /usr/src/app -type f -exec chmod 644 {} \;
   - find /usr/src/app -type d -exec chmod 755 {} \;
   - chmod -R ug+rwx storage /usr/src/app/bootstrap/cache
+
+  - php artisan config:cache
+  - php artisan route:cache
```
waghanza, line 22: is the warm-up not enough ?
https://github.com/the-benchmarker/web-frameworks/blob/b99ea681988502f59508df4062aaf48372e81dbd/.tasks/collect.rake#L37

waghanza (changes requested):
Thanks for this upgrade ❤️

php/laravel/artisan
barryvdh: The file isn't needed to run, but is required for caching. Not sure how to include them otherwise?

php/laravel/composer.json
barryvdh: No, this is a SemVer requirement, which will allow any 8.x version, but at least 8.12. It picks 8.22 now, but will upgrade to 8.23 etc (so all minor updates). This is the constraint used in laravel/laravel now.

php/laravel/config.yaml
barryvdh: No, warmup does not cache either the config or routes. This has to be done manually upon deployment.

php/laravel/app/Http/Kernel.php
barryvdh: We could, I left it to make the changes from laravel/laravel more visible.

waghanza: and I'm glad you've taken this path. you're right lets take implementation to minimal in an other `PR`

php/laravel/artisan
waghanza: I think we can remove it. Cache will be create on warm-up

php/laravel/composer.json
waghanza: yes, it's intended to keep `8.22`, to be aligned with the doc (a bot will take care of update)

php/laravel/config.yaml
waghanza: ok, so forgot what I said about `artisan` binary

waghanza (changes requested):

php/laravel/composer.json
barryvdh: Ok but because it's SemVer, any 8.x release should work. So ^8 would perhaps be easier then letting the bot update each time.

waghanza (dismissed):

waghanza (approved):

barryvdh:
@KaungZawHtet relax, it will updated in due time. I can't benchmark the results myself.

I've also removed some code/stuff that was not used, not sure how much the impact is. https://github.com/the-benchmarker/web-frameworks/pull/3799

PR #3786, opened 2021-01-16
Update swoole config

sy-records:
#### Tasks

+ Set worker number to `swoole_cpu_num() * 2`
+ Disable coroutine
+ Disable logs

#### Frameworks

+ [x] chubbyphp-swoole
+ [x] hyperf
+ [x] imi
+ [x] one
+ [x] siler-swoole
+ [x] simps
+ [x] slim-swoole
+ [x] sw-fw-less
+ [x] swoft
+ [x] yii-swoole

PR #3800, opened 2021-01-18
[PHP] Added DriftPHP as a Framework

mmoreram:


mmoreram:
I have a question

- This framework is based on a single-PHP-thread set of components named ReactPHP. Some other frameworks are actually using Nginx, which can work with several PHP threads at the same time, taking the maximum power of the server. Is there any rule here?

waghanza:
not for this, yet

this should be clarified on `README`

leocavalcante (approved):

leocavalcante:
> I have a question
> 
> * This framework is based on a single-PHP-thread set of components named ReactPHP. Some other frameworks are actually using Nginx, which can work with several PHP threads at the same time, taking the maximum power of the server. Is there any rule here?

ReactPHP itself says [you don't need threads](https://github.com/reactphp/reactphp/wiki/FAQ#what-about-threads), it is part of the runtime's architectural decision, if they say it could be efficient in a single-thread IMHO I don't see it as a competitive disadvantage to be solved by rules.
And actually, after #3786 with Swoole running with `SWOOLE_BASE`, it will run as a single-threaded server... also like other *competitors* like Node.js and Dart.

mmoreram:
@leocavalcante yes, ReactPHP is so efficient. I'm not talking in terms of absolute values, but in relative values (like Swoole, or NodeJS). Adding resources rules could be a nice addition, basically to provide all frameworks the same available resources (let's say 4 parallel CPUs, bringing the possibility to 1-single-thread servers to balance between these CPUs. Just an idea :)

Thanks for your work, BTW. Really cool project :)

leocavalcante:
Oh, I see. Yes, would be nice to have the used resources specified and ensured that all frameworks are running under the same conditions and then what the framework does with the resources is about it's features.

Ps.: Not sure if you are thanking me for this project or you forgot to mention @waghanza, but to clarify, he is the one putting way more efforts, I'm just new here.

mmoreram:
@leocavalcante btw, about your reference, and after reviewing the documentation, and **not being 100% sure about this**, Swoole works as a PHP extension, having a configuration called `swoole_cpu_num`. The same extension allows you to work with this CPU concurrency.

mmoreram:
@leocavalcante Oh, I saw that you approved this PR, so I thought that. Sorry cc/ @waghanza

BTW, about your reference, and after reviewing the documentation, and not being 100% sure about this, Swoole works as a PHP extension, having a configuration called swoole_cpu_num. The same extension allows you to work with this CPU concurrency.

leocavalcante:
@mmoreram unfortunately, the english version of Swoole's documentation isn't complete as the Chinese wiki, but Google Translator can help: https://translate.google.com/translate?sl=auto&tl=en&u=https://wiki.swoole.com/wiki/page/353.html
There is also this issue: https://github.com/swoole/swoole-src/issues/1085 simple, but says exactly that.

waghanza:
About resources : when using cloud resources (plan but have to be written in the roadmap) any framework will be set-up in an isolate VM (of course same VMs for all), it up-to the code to spread on all CPU units

mmoreram:
@waghanza when you mean the code, you mean the same framework, right? I mean, if DriftPHP relies in the user balancing instead of internal framework balancing, could I add an external layer of balancing?

waghanza:
@mmoreram Short answer, no. Long answer, perhaps. 

As you know, infrastructure has a cost. The first step is to spot one VM per framework, depending on that we can plan any feature like that

mmoreram:
@waghanza I think that I explained wrong. When I talk about external balancing, I mean that the entrypoint of the framework is a balancer among X (let's say 4) instances of the framework. I you understood that, then OK with that answer :)

waghanza:
I had understood before, but probably mis-explained 😛  there will not be external load load balancing. What's the framework is doing inside (balance between threads or reuse_port) is out of this project scope

mmoreram:
Good :) Thanks!

php/driftphp/composer.json
```diff
@@ -0,0 +1,33 @@
...
+    "description": "Skeleton repository ofr DriftPHP",
+    "type": "project",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Marc Morera",
+            "email": "yuhu@mmoreram.com"
+        }
+    ],
+    "require": {
```
waghanza, line 12: seems that their is duplicate

php/driftphp/composer.json
```diff
@@ -0,0 +1,33 @@
...
+        "php": "^7.3",
+        "symfony/console": "^5.0.0",
```
waghanza, line 14: already required in https://github.com/driftphp/server/blob/c36d51be73afd36bc9c6d7bd1070ce1676a698b3/composer.json#L17

php/driftphp/composer.json
```diff
@@ -0,0 +1,33 @@
...
+    "type": "project",
+    "license": "MIT",
+    "authors": [
+        {
+            "name": "Marc Morera",
+            "email": "yuhu@mmoreram.com"
+        }
+    ],
+    "require": {
+        "php": "^7.3",
```
waghanza, line 13: does `driftphp` support php 8 ?

php/driftphp/composer.json
```diff
@@ -0,0 +1,33 @@
...
+        "symfony/console": "^5.0.0",
+        "symfony/framework-bundle": "^5.0.0",
+        "symfony/dotenv": "^5.0.0",
+        "symfony/stopwatch": "^5.0.0",
+        "symfony/debug": "^4.4.0",
+
+        "drift/http-kernel": "0.1.*, >=0.1.7",
```
waghanza, line 20: this is not very consistent with the constraint above, I suggest using `~` instead of `^`

php/driftphp/config.yaml
```diff
@@ -0,0 +1,15 @@
+framework:
+  github: driftphp/driftphp
+  version: 1.0
```
waghanza, line 3: it is not `0.1` for now ? (https://renovatebot.com/ will take care of update)

php/driftphp/config.yaml
```diff
@@ -0,0 +1,15 @@
...
+
+provider:
+  default:
+    language: 7.4
```
waghanza, line 7: no need `here`

waghanza (changes requested):

php/driftphp/composer.json
mmoreram: Yes, should.

php/driftphp/composer.json
mmoreram: I remove this dependency because, in this case, the console is not used directly. This dependency must be present in the skeleton, as the console is used directly there, outside the server.

php/driftphp/composer.json
mmoreram: I don't see the inconsistency here. Can you explain a bit more?

php/driftphp/composer.json
mmoreram: I don't understand :(

php/driftphp/composer.json
mmoreram: The base docker image is still on top of php7.4. As soon as is updated, I can update here the image.

php/driftphp/composer.json
waghanza: I mean that some subdependencies are listed. For example, some dependencies of https://github.com/driftphp/server/blob/master/composer.json are here, which seems to be an error

php/driftphp/composer.json
waghanza: The image is created on-the-fly 😛  
Could you set something like
```suggestion
        "php": "~7.4.0|~8.0.0",
```
we need to unify this constraint for all php frameworks, but could be in an other `PR`

I'm not very safe with the `^` constraint, despite it is the default in composer

php/driftphp/composer.json
waghanza: `^4.4.0` and `0.1.*, >=0.1.7` are quite similar in meaning for composer, but is not very clear (for reader), better use the same semantic

php/driftphp/composer.json
mmoreram: But these are different packages. In this package we have the dependencies this package needs, and some other dependencies, which can be the same ones, are listed in the server composer file because the server, in an isolated way, needs them.

php/driftphp/composer.json
mmoreram: This will only work if the composer install is done with `--ignore-platform-reqs` flag, basically because some dependencies might require **only** php^7 and that would make composer install crash.

php/driftphp/composer.json
mmoreram: Updated,

php/driftphp/composer.json
waghanza: seems this could lead to version mismatch for `driftphp` ... but anyone we could consider it at resolved

php/driftphp/Drift/config/routes.yml
```diff
@@ -0,0 +1,15 @@
+# Routes
+ping:
+  path: /
+  controller: App\Controller\DefaultController:ping
```
waghanza, line 4: better name this `index`

waghanza (changes requested):

php/driftphp/Drift/config/routes.yml
mmoreram: Done.

waghanza (approved):

PR #3806, opened 2021-01-19
[Clojure] Add donkey

whiplash:
Hi @yaronel,

Would you mind to add (or consent that we add) [donkey](https://github.com/AppsFlyer/donkey) in `here` ?

------------

Regards,


yaronel:
Hi. 
Yeah, go ahead and add it. 
just please add the command line arguments as they appear here:
https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/Clojure/donkey/donkey.dockerfile

waghanza:
You mean https://github.com/TechEmpower/FrameworkBenchmarks/blob/6d43ad15d76b3dc1c51bcf988b30d86fd876fd22/frameworks/Clojure/donkey/donkey.dockerfile#L12, right ?

`vertx` options are OK, but I'm not sure about `Xsm2G`. is it for memory limit ?

whiplash:
`Xms` specifies the initial memory allocation size

mfirry:
Not sure `UseStringDeduplication` and `UseNUMA` will make much difference in this case.

yaronel:
The vertx options are the most important. 
Setting the min and max memory to the same value avoids committing more memory during runtime.

clojure/donkey/config.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+  website: https://github.com/AppsFlyer/donkey
```
waghanza, line 2: clearer 😛 

```suggestion
  github: AppsFlyer/donkey
```

clojure/donkey/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+  version: 0.4.0
```
waghanza, line 3: better
```suggestion
  version: 0.4
```

waghanza (changes requested):
minor typos

clojure/donkey/project.clj
```diff
@@ -0,0 +1,8 @@
+(defproject the-benchmarker-donkey "0.1.0-SNAPSHOT"
+  :dependencies [[org.clojure/clojure "1.10.1"]
+                 [com.appsflyer/donkey "0.4.0"]]
```
waghanza, line 3: is it required to pin the exactas  version in `clojure` ? could not we use a range, like `[0.4,0.5)` @yaronel

whiplash: Not sure how leiningen handles ranges. Will investigate.

waghanza (approved):

PR #3813, opened 2021-01-20
[Java] Add quarkus

whiplash:


java/quarkus/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+  <properties>
+    <compiler-plugin.version>3.8.1</compiler-plugin.version>
+    <maven.compiler.parameters>true</maven.compiler.parameters>
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
+    <quarkus-plugin.version>1.11.0.Final</quarkus-plugin.version>
```
waghanza, line 17: is there any parity between `quarkus.platform.version` and `quarkus-plugin.version` ?

java/quarkus/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+    <quarkus.platform.artifact-id>quarkus-universe-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>1.11.0.Final</quarkus.platform.version>
```
waghanza, line 20: better use a range I think
```suggestion
    <quarkus.platform.version>[1.11,1.12)</quarkus.platform.version>
```

java/quarkus/config.yaml
```diff
@@ -0,0 +1,11 @@
+framework:
+  website: quarkus.io
+  version: 1.11
+
+build:
+  - mvn clean package -Dquarkus.package.type=uber-jar -Dquarkus.http.port=3000 -Pnative
```
waghanza, line 6: this is not duplicated ? I mean this configuration is already in `application.resources`

waghanza (changes requested):

java/quarkus/pom.xml
whiplash: I would think so, yeah. Maybe @geoand can confirm.

java/quarkus/config.yaml
whiplash: Good point. For some reasons in my tests `quarkus.http.port=3000` wasn't being picked up from the conf file. Not sure why.

java/quarkus/pom.xml
geoand: Yes, those need to match

java/quarkus/config.yaml
whiplash: Maybe @geoand knows what's going on here.

geoand: It added to `application.properties`, it should definitely work. It however you are trying to change the port used for testing, then the property is `quarkus.http.test-port`.

java/quarkus/pom.xml
waghanza: then you could use only one version (`quarkus.platform.version`) instead of two vars in `pom.xml` I suppose @whiplash

geoand: Yeah, that's pretty common for users to do

java/quarkus/config.yaml
whiplash: Just tried again. Have a look: https://gist.github.com/whiplash/8fd3004593c8f9440d32caafd13fe2a5

java/quarkus/pom.xml
whiplash: Done

java/quarkus/config.yaml
geoand: I'll check it out

geoand: I can't reproduce the problem. I just created a hello world application, set `quarkus.http.port=3000` in `application.properties` and when I launched the application is correctly started on port `3000`.

mfirry: I'll provide a repro. Later today. Thanks so much.

whiplash: I'll provide a repro repo later today. Thanks so much anyways!

whiplash: Have a look https://github.com/whiplash/quarkus-port-repro

java/quarkus/pom.xml
```diff
@@ -0,0 +1,93 @@
...
+    <maven.compiler.source>11</maven.compiler.source>
+    <maven.compiler.target>11</maven.compiler.target>
+
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+    <project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+
+
+    <quarkus.platform.artifact-id>quarkus-universe-bom</quarkus.platform.artifact-id>
+    <quarkus.platform.group-id>io.quarkus</quarkus.platform.group-id>
+    <quarkus.platform.version>[1.11,1.12)</quarkus.platform.version>
```
whiplash, line 20: This doesn't work for me for some reasons. I think it has to do with the fact that the version is with a `Final` string at the end.

java/quarkus/config.yaml
geoand: The problem is that you have named the config file `application.resources` instead of `application.properties`

java/quarkus/pom.xml
waghanza: any idea why @geoand ?
https://github.com/the-benchmarker/web-frameworks/pull/3813/checks?check_run_id=1754048979#step:6:7772

java/quarkus/config.yaml
whiplash: Thanks. Sorry about this.

java/quarkus/pom.xml
geoand: Not really. We've never really tested Maven ranges

java/quarkus/config.yaml
geoand: NP

java/quarkus/pom.xml
whiplash: That's what I thought. I've never seen them used (but @waghanza likes it so much 😛 )

PR #3816, opened 2021-01-21
reduce number of threads for RESTHeart

ujibang:
Each performance test request requires microseconds to execute, so a couple of working threads should be enough for RESTHeart.


java/restheart/config.yaml
```diff
@@ -10,7 +10,7 @@ files:
 
 environment:
   RH_IO_THREADS: 1
-  RH_WORKER_THREADS: 4
+  RH_WORKER_THREADS: 2
```
waghanza, line 13: could you provide a dynamic value ? I mean setting something like : `nproc` (you can assume that we run on linux)

waghanza (changes requested):

java/restheart/config.yaml
ujibang: we tested it on different systems and always got best performance with 2 working threads.

we also noticed that test implementation of undertow (RESTHeart is built on top of it) just uses 1 io_thread, not even dispatching the request to a working thread

we realized that the test is so trivial (a request takes few microseconds) that enabling more threads leads to having an overhead.

having more working threads has great advantages when the request do execute some real work, like interacting with a database. for instance, if we add 100ms wait in each call, with 64 working threads and `wrk -c 64 -t 64` we get the best results!

so `nproc` is not needed!

ujibang: May I suggest to add a different performance test?

Basically each request should block for 100ms to simulate a more realistic request processing.

With this delay the maximum throughput is 10 request/second per thread (or similar concurrency mean) 

If `wrk` runs with 64 threads, the maximum throughput is 640 request per seconds.

RESTHeart (with 64 working threads) has a  throughput of ~620 requests/second reaching the 96,9% of the maximum. 

In this case 3.1% is the overhead associated with request routing and networking.

waghanza: > May I suggest to add a different performance test?

I will open a discussion this week end. The actual implementation does not reflect a **real-life** usage

Please note that this project target **production** usage, I mean, tuning performances is counter-productive. We want to reflect what the framework could when deployed in **production** (so not only with performance in mind, but safety, best practices ...)

ujibang: Yes, for instance handling at least basic authentication is paramount IMO for a more realistic test

waghanza: I'll merge this one.

waghanza (approved):

ujibang:
uh! I just noticed the latest results and throughput felt from 74k/sec to 56k/sec

maybe reducing the number of threads wasn't that clever! :)

PR #3819, opened 2021-01-21
Improved DriftPHP performance. Added multi-thread feature.

mmoreram:
After last results, I implemented this small feature in the server in order to start using all available threads. Fingers crossed.

### DIFF
--- a/php/driftphp/composer.json
+++ b/php/driftphp/composer.json
@@ -9,9 +9,9 @@
     ],
     "require": {
         "php": "~7.4.0|~8.0.0",
-        "symfony/framework-bundle": "^5.0.0",
-        "drift/http-kernel": "^0.1.7",
-        "drift/server": "^0.1.7"
+        "symfony/framework-bundle": "~5.0.0",
+        "drift/http-kernel": "~0.1.7",
+        "drift/server": "~0.1.17"
     },
     "autoload": {
         "psr-4": {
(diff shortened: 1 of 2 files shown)


php/driftphp/config.yaml
```diff
@@ -10,4 +10,4 @@ php_mod:
   - posix
   - sockets
 
-command: php vendor/bin/server run 0.0.0.0:3000 --env=prod --quiet
\ No newline at end of file
+command: php vendor/bin/server run 0.0.0.0:3000 --env=prod --quiet --workers=-1
```
waghanza, line 13: is it more more interesting to spread across all CPU ? instead of using 1 worker

please note that I'm not sure that is useful, since I don't know how reactphp react 😛

php/driftphp/composer.json
```diff
@@ -11,7 +11,7 @@
         "php": "~7.4.0|~8.0.0",
         "symfony/framework-bundle": "^5.0.0",
         "drift/http-kernel": "^0.1.7",
-        "drift/server": "^0.1.7"
+        "drift/server": "^0.1.17"
```
waghanza, line 14: Please use `~` instead of `^`

waghanza (changes requested):
`composer.lock` should not be versioned

php/driftphp/composer.json
mmoreram: Done

waghanza (approved):

PR #3835, opened 2021-01-23
[Scala] Add Finatra

mfirry:
Adding [Finatra](twitter.github.io/finatra/)


whiplash:
Hi @cacoco,

Would you mind adding (or consent that we add) [finatra](https://twitter.github.io/finatra/) in `here` ?

------------

Regards,

cacoco:
@waghanza I think this is OK but note there are some performance gains in turning off certain default behaviors of the underlying Finagle server (which you can see here: https://github.com/TechEmpower/FrameworkBenchmarks/blob/96bfbd7998e9b1239f9ffc253e6deb831d70544d/frameworks/Scala/finatra/src/main/scala/Main.scala#L12).

Additionally, we generally release a new version of the framework monthly so I would ask what is the plan for updating the version used here?

whiplash:
hi @cacoco and thanks for the quick feedback. I've added the `configureHttpServer` part as suggested (curious to know what's the performance gain... will test with and without just to know).

about the updates, this project uses https://github.com/renovatebot/renovate so I guess we'll catch up either with it or by checking every couple of months.

cacoco:
@whiplash sounds good. Also there are JVM arguments as well that help if you're trying to get a sense of the raw performance. See: https://github.com/TechEmpower/FrameworkBenchmarks/blob/96bfbd7998e9b1239f9ffc253e6deb831d70544d/frameworks/Scala/finatra/finatra.dockerfile#L10.

Thanks for adding Finatra here. We appreciate it.

waghanza:
Yep @cacoco. `renovate` will track update (all `scala` based frameworks are update this way). As long as `scala` version (in containers definition).

\May I notify you (or someone else) on update ?

waghanza (approved):

PR #3841, opened 2021-01-24
Clarify documentation

waghanza:
Hi,

This `PR` enhance homepage, `README`.

The purposes are :
+ clarify intentions / purposes of this project
+ show how results are produced
+ show usages

Any feedback is :heart: 

Regards,


README.md
```diff
@@ -1,60 +1,87 @@
...
+   <a href="https://github.com/the-benchmarker/web-frameworks/blob/master/LICENSE" target="_blank">
+      <img src="https://img.shields.io/github/license/the-benchmarker/web-frameworks" alt="License">
+   </a>
+</p>
 
-![CI](https://github.com/the-benchmarker/web-frameworks/workflows/CI/badge.svg)
+## Motivation
 
-[![Chat with US](https://img.shields.io/badge/slack-Chat_with_us-blueviolet)](https://thebenchmarker.slack.com)
+It exists plenty of frameworks. Each one has its advantages, and drawbacks. The idea of this project is to list those frameworks and to expose metrics for each one listed `here` (performance is only the first one)
```
ahopkins, line 19: "There are many frameworks, each one having its own advantages and drawbacks. The purpose of this project is to identify them and attempt to measure their differences (performance is only one metric)."

README.md
```diff
@@ -1,60 +1,87 @@
...
 
-This project aims to be a load benchmarking suite, no more, no less
+#### What is a framewok ?
```
ahopkins, line 21: "framework"

README.md
```diff
@@ -1,60 +1,87 @@
...
 
-> Measuring response times (routing times) for each framework (middleware).
+A framework is a set of components working together. There is a **huge** variety of frameworks, from full-stack (when deling with multiple layers) to micro (when providing only routing part)
```
ahopkins, line 23: "dealing"

I am not sure that this is a sufficient definition.

waghanza: yeah need probably more details, but try to keep it simple

will take time to clarify

README.md
```diff
@@ -1,60 +1,87 @@
-# Which is the fastest?
+# Which is the fastest ?
```
System-Glitch, line 1: Some text and headers suggest only the speed is measured, but some other say that other metrics are also measured. It's not very consistent. I think it would be best to make it clear everywhere that this benchmark is not only about speed.

README.md
System-Glitch, line 19: Not really related sorry, but I had a random idea when reading this part: if possible, that would be cool to also add a list of pros and cons to each framework alongside the benchmarking results, so this project would become the one go-to source with everything you need to know to pick the framework you need. These lists would have to be filled by either the maintainer of the framework (that would be best IMO) or by experienced developers working with that framework.

That may be worth opening a new discussion.

README.md
```diff
@@ -1,60 +1,87 @@
...
+<details>
+  <summary><strong>Technical details</strong></summary>
+  <ul>
+   <li>CPU : 8 Cores (AMD FX-8320E Eight-Core Processor)</li>
+   <li>RAM : 16 Gb</li>
+   <li>OS : Fedora</li>
+   <li><pre>Docker version 20.10.0-rc1, build 5cc2396
+</pre></li>
+  </ul>
+</details>
```
System-Glitch, line 79: Very nice addition there. This was definitely something missing. 👍🏻

README.md
ahopkins: There was a discussion about this before where the thought was to build a frontend project that would be powered partially with data from here with some of this information.

PR #3847, opened 2021-01-25
Upgrade to Masonite 3

vaibhavmule:
cc @josephmancuso


python/masonite/requirements.txt
```diff
@@ -1,3 +1,3 @@
-masonite>=2.3.1,<2.4.0
+masonite>=3.0,<4.0
```
waghanza, line 1: that's a huge range. This one not guarantee that we are in 3.0, right ?
```suggestion
masonite>=3.0.0,<3.1.0
```

waghanza (approved):

python/masonite/requirements.txt
vaibhavmule: @waghanza I don't have an idea about the roadmap, @josephmancuso please guide here.

waghanza: @vaibhavmule what I meant is that we show version on `README` (based on `config.yaml`). in order to be in sync, we must restrict the range

josephmancuso: I'm 99% sure Masonite 4 will be released after Masonite 3 but putting `masonite>=3.0.0,<3.1.0` is a safe bet for now.

waghanza: that leave 1% uncertainty @josephmancuso 😛

PR #3877, opened 2021-01-29
[PHP] Update swoole mode

sy-records:
@waghanza I noticed that some frameworks are not included in the results, What's the reason? For example, [imi](https://github.com/the-benchmarker/web-frameworks/tree/master/php/imi)


waghanza:
indeed @sy-records, I've not created an issue, but I've notice (imi is skipped on results)

I've created https://github.com/the-benchmarker/web-frameworks/issues/3879 so we can investigate

PS : Specs are OK for `imi`, btw

php/swoole/start.php
```diff
@@ -4,7 +4,7 @@
 use Swoole\Http\Request;
 use Swoole\Http\Response;
 
-$server = new Server('0.0.0.0', 3000);
+$server = new Server('0.0.0.0', 3000, SWOOLE_BASE);
```
waghanza, line 7: is there a default value for this param @sy-records ?

sy-records: SWOOLE_PROCESS

waghanza (approved):

PR #3926, opened 2021-02-06
Athena Upgrade

Blacksmoke16:
* Upgrade Athena to `0.13.x`
* Bump Dockerfile to use Crystal `0.36.1`
* Add new generated files to .gitignore
* Add `jq` to requirements section in README

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,7 @@
 .neph
 neph.yaml
 shard.lock
+Gemfile.lock
 
 # git temporary files
 *.orig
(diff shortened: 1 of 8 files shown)


Blacksmoke16:
IDK why CI is failing.  I'm deff returning an empty string now.

waghanza:
seems to returns `""` not an empty string

https://github.com/the-benchmarker/web-frameworks/pull/3926/checks?check_run_id=1846883774#step:7:15

Blacksmoke16:
@waghanza I was returning `nil` before and CI still was failing.  Something doesn't seem right here.

waghanza:
Indeed `Nil` has the right behavior. Seems weird for me to have a block that returns a nullable object when speaking of HTTP, btw

Blacksmoke16:
@waghanza In Athena having an explicit `Nil` return type maps to a `204 No Content` response.

Issue #3922, opened 2021-02-06
[PHP] Sunrise is failing with roadrunner


waghanza:
Hi @fenric,

I've updated CI (I notice it was useless) and then it appear that `sunrise` is failing with `roadrunner`

https://github.com/the-benchmarker/web-frameworks/runs/1845194145?check_suite_focus=true#step:7:10

Regards,

fenric:
Hi! I think I understand why it happened. The problem isn't with the router, but with the test... I'm cooking a fix.

fenric:
@waghanza #3928

PR #3928, opened 2021-02-08
[PHP] Sunrise Router / improved code

fenric:
#3922

### DIFF
--- a/php/sunrise-router-annotations/composer.json
+++ b/php/sunrise-router-annotations/composer.json
@@ -2,7 +2,7 @@
   "require": {
     "sunrise/http-router": "~2.6.0",
     "sunrise/http-factory": "*",
-    "symfony/cache": "^5.0"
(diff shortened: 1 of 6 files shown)


php/sunrise-router-annotations/composer.json
```diff
@@ -1,7 +1,8 @@
 {
   "require": {
     "sunrise/http-router": "~2.6.0",
-    "sunrise/http-factory": "*"
+    "sunrise/http-factory": "*",
+    "symfony/cache": "^5.0"
```
waghanza, line 5: I find mixing `~` and `^` confusion, and not very accurate in this case
```suggestion
    "symfony/cache": "*"
```

fenric:
how can I install php apc cache? this is needed for an annotated instance...

fenric:
I see, thx... but apcu is needed for router-annotations... :)

PR #3967, opened 2021-02-16
simplify benchmark implementation of RESTHeart

ujibang:
Simplified  benchmark implementation of RESTHeart

Also put working threads configuration back to 4 (due to latest decreasing performance)


java/restheart/config.yaml
```diff
@@ -10,7 +10,7 @@ files:
 
 environment:
   RH_IO_THREADS: 1
-  RH_WORKER_THREADS: 2
+  RH_WORKER_THREADS: 4
```
waghanza, line 13: you will probably see the performance decrease, if running on cloud.

I think the better here is to configure what you recommend for a **production** environment

if their is a formula based on hardware resource, you can probably put any bash expression (I'll take care on evaluation)

java/restheart/restheart.yml
```diff
@@ -31,9 +31,9 @@ instance-name: restheart-benchmarker
 #### Performance Options
 
 # Number of I/O threads created for non-blocking tasks. at least 2. suggested value: core*2
-io-threads:
-  1
-  # Number of threads created for blocking tasks (such as ones involving db access). suggested value: core*16
+io-threads: 1
```
waghanza, line 34: is it not duplicate with env var `RH_IO_THREADS` set above ?

ujibang: restheart.yml is the configuration file, the env variables _override_ the values in the conf file.

java/restheart/config.yaml
ujibang: The problem is that the test is so trivial (the execution of the actual service last for microseconds) that having many threads is actually counterproductive. Having more threads boosts performances if the service is doing some actual work, for instance interacting with a db. I also suggested you to add a different, more realistic performance test than your "routing test".

the formula is described in the configuration file

```
# Number of I/O threads created for non-blocking tasks. at least 2. suggested value: core*2
io-threads: 1

# Number of threads created for blocking tasks (such as ones involving db access). suggested value: core*16
worker-threads: 4
```

On my development environment I get the best performances with io-threads=1 and working-threads=2

However we felt from 70k req/set to 35k in last results, and thought this was the cause.

waghanza: the main idea (despite what is actually displayed on `REDME`) is not to have only **performances**, but also resources ... this project should reflect how a framework will behave in **production**

we have to discuss about an expressive implementation

waghanza (approved):

PR #3655, opened 2020-12-26
[PHP] Upgrade to 8.0

waghanza:
Hi,

This `PR` replaces #3606, it upgrade PHP to **8.0**

Regards,

PR #3982, opened 2021-02-19
Ice, don't get from pecl until we get a stable version

mruz:
https://github.com/the-benchmarker/web-frameworks/pull/3655#issuecomment-780432282


waghanza (approved):
🎉

PR #4039, opened 2021-03-05
Add Swift Hummingbird framework

adam-fowler:
New swift web framework

### DIFF
--- a/swift/hummingbird-framework/Package.swift
+++ b/swift/hummingbird-framework/Package.swift
@@ -1,0 +1,27 @@
+// swift-tools-version:5.3
+// The swift-tools-version declares the minimum version of Swift required to build this package.
+
+import PackageDescription
+
+let package = Package(
+    name: "server",
+    products: [
+        .executable(name: "server", targets: ["server"])
+    ],
+    dependencies: [
+        .package(url: "https://github.com/hummingbird-project/hummingbird.git", .upToNextMinor(from: "0.5.0")),
+    ],
+    targets: [
+        .target(name: "server",
+            dependencies: [
+                .product(name: "Hummingbird", package: "hummingbird"),
+            ],
+            swiftSettings: [
+                // Enable better optimizations when building in Release configuration. Despite the use of
+                // the `.unsafeFlags` construct required by SwiftPM, this flag is recommended for Release
+                // builds. See <https://github.com/swift-server/guides#building-for-production> for details.
+                .unsafeFlags(["-cross-module-optimization"], .when(configuration: .release))
+            ]
+        ),
+    ]
+)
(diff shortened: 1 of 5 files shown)


swift/hummingbird-framework/Package.swift
```diff
@@ -0,0 +1,27 @@
...
+    ],
+    dependencies: [
+        .package(url: "https://github.com/hummingbird-project/hummingbird.git", .upToNextMinor(from: "0.6.0")),
+    ],
+    targets: [
+        .target(name: "server",
+            dependencies: [
+                .product(name: "Hummingbird", package: "hummingbird"),
+            ],
+            swiftSettings: [
```
waghanza, line 19: this is awesome 🎉 

perhaps a good idea to use this for all `swift` **frameworks** @0xTim

swift/hummingbird-framework/config.yaml
```diff
@@ -0,0 +1,8 @@
+framework:
+  website: https://github.com/hummingbird-project
```
waghanza, line 2: better use the `github` shortcut
```suggestion
  github: hummingbird-project/hummingbird
```
this will link to https://github.com/hummingbird-project/hummingbird

swift/hummingbird-framework/config.yaml
```diff
@@ -0,0 +1,8 @@
...
+  version: 0.6.0
```
waghanza, line 3: Better set `0.6`, `upToNextMinor` will allow `0.6.1` tho

swift/hummingbird-framework/config.yaml
```diff
@@ -0,0 +1,8 @@
...
+  name: hummingbird
```
waghanza, line 4: is this already require @0xTim ? I mean to have a directory not named `hummingbird`

swift/hummingbird-framework/config.yaml
```diff
@@ -0,0 +1,8 @@
...
+
+environment:
+  SERVER_HOSTNAME: 0.0.0.0
+  SERVER_PORT: 8080
```
waghanza, line 8: we use 3000 here
```suggestion
  SERVER_PORT: 3000
```

waghanza (changes requested):

swift/hummingbird-framework/config.yaml
adam-fowler: It's a limitation of the swift package manager. It gets confused if you have a dependency with the same name as the folder you are in. It thinks there is a cyclical dependency.

waghanza (changes requested):
could you run `bundle exec rake ci:config`, in order to add `humingbird` to CI ?

swift/hummingbird-framework/Package.swift
0xTim: Yes we should probably add it to all Swift frameworks that work with it

PR #3958, opened 2021-02-14
Multiple engines

waghanza:
Hi,

This `PR` allows to test each frameworks on various setups.

###### Actual

All frameworks are tested once. For example, `ruby` **framework** are tested with `puma` (since widely spread in `ruby` _communities_)

###### Purpose

Test framework with multiple engines (for the sake of this feature we keep focus on engines, not compilation options, vms, ...). The idea the is to test frameworks with various engines, when accurate indeed.

-------
List of engines

-------

+ Ruby
   + https://puma.io
   + https://github.com/socketry/falcon
   + https://github.com/boazsegev/iodine
+ Javascript
   + https://github.com/nodejs/node
   + https://deno.land
   + https://github.com/uNetworking/uWebSockets.js
+ Python
   + https://gitlab.com/pgjones/hypercorn
   + https://github.com/django/daphne
   + https://github.com/encode/uvicorn
   + https://github.com/benoitc/gunicorn
   + https://github.com/jonashaag/bjoern
   + https://github.com/Pylons/waitress
+ PHP
   + https://github.com/php/php-src, aka _php-fpm_
   + https://github.com/swoole/swoole-src
   + https://github.com/walkor/Workerman
   + https://github.com/reactphp/reactphp
   + https://github.com/amphp/amp
   + https://roadrunner.dev/

PR #4041, opened 2021-03-06
Add Laravel + Servel

gregorip02:


waghanza:
Thanks for your awesome contribution ❤️ 

If I may, `servel` is not a framework, but an adapter for `workerman`. As such, I think having it here is not accurate.

However, I can use it for https://github.com/the-benchmarker/web-frameworks/pull/3958, this will allow `laravel` to be use either with `php/fmp` or with `workerman` (we have also `swoole` coming).

Do you agree ?

gregorip02:
Sure, I agree with you, I'll be watching when this project can test variants (:

waghanza:
I can do the `PR` :-) and will will cc you

waghanza (approved):

PR #4058, opened 2021-03-10
remove tokio from salvo dependencies

chrislearn:


waghanza (changes requested):
`Cargo.lock` is not needed

could you also update `config.yaml` with `salvo-rs/salvo` (since github repo has moved)

as I understand `tokio` is not needed, I don't understand the `tokio` usage described in https://salvo.rs/book/quick-start/hello_world/

waghanza (approved):

chrislearn:
@waghanza I think I just give a way to avoid add tokio to dependiences in final projects. but the truth is salvo and hyper are base on tokio.

chrislearn:
@waghanza is that possible to add Cargo. lock to git ignore file?

PR #4090, opened 2021-03-21
[PHP] Add workerman for laravel

waghanza:
Hi @gregorip02,

When running `php artisan servel start`, I have the following error message 
```php
Servel 0.0.1 by Gregori Piñeres and contributors.

In Servel.php line 29:
                                                                                                                  
  Argument 1 passed to Servel\Servel::Servel\{closure}() must be of the type array, string given, called in /usr  
  /src/app/vendor/laravel/framework/src/Illuminate/Collections/Traits/EnumeratesValues.php on line 234 
```

Regards,

### DIFF
--- a/php/laravel/config/servel.php
+++ b/php/laravel/config/servel.php
@@ -1,0 +1,8 @@
+<?php
+
+return [
+    'servers' => [
+        'name' => 'http://0.0.0.0:80',
+        'count' => 16
+    ]
+];
--- a/php/laravel/config.yaml
+++ b/php/laravel/config.yaml
@@ -2,21 +2,40 @@ framework:
   website: laravel.com
   version: 8.33
 
-environment:
-  APP_ENV: production
-  APP_DEBUG: false
-  APP_KEY: base64:txfHNf/SOo222Rm8I39Urb9SmvUy+nuAF98t/ukF0lk=
+  environment:
+    APP_ENV: production
+    APP_DEBUG: false
+    APP_KEY: base64:txfHNf/SOo222Rm8I39Urb9SmvUy+nuAF98t/ukF0lk=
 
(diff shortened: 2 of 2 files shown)


php/laravel/config/servel.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return [
+    'servers' => [
+        'name' => 'http://0.0.0.0:80',
+        'count' => 16
```
waghanza, line 6: it should depends on hardware ressources

php/laravel/config/servel.php
```diff
@@ -0,0 +1,8 @@
+<?php
+
+return [
+    'servers' => [
+        'name' => 'http://0.0.0.0:80',
```
waghanza, line 5: should use `3000`

gregorip02: Sure, the configuration I posted has a bug. Please change the content of `config/servel.php` to the following.

```php
<?php

return [
    'servers' => [
        [
            'name' => 'http://0.0.0.0:3000',
            'count' => 16 // It depends on the hardware, change it if necessary.
        ]
    ]
];
```

This will solve the execution problem, thanks.

gregorip02 (changes requested):
`config/servel.php`

```php
<?php

return [
    'servers' => [
        [
            'name' => 'http://0.0.0.0:3000',
            'count' => 16 // It depends on the hardware, change it if necessary.
        ]
    ]
];
```

waghanza:
Most important is the above error @gregorip02

gregorip02:
> Most important is the above error @gregorip02

Yes, in fact it would be resolved with this change.

waghanza:
Same error

PR #4091, opened 2021-03-21
[PHP] Add variants support

waghanza:
/cc @gregorip02


gregorip02 (changes requested):
We still have the same problem, the initial configuration I added has a problem. The index "servers" must be an array with a unique configuration for a "server" because we could build a websocket or http server.

Use the following settings in the file `config/servel.php`

```php
<?php

return [
    'servers' => [
        ['name' => 'http://0.0.0.0:3000', 'count' => shell_exec('nproc')]
    ]
];
```

waghanza:
My feedback :
- should not above configuration be in `config/app.php` ?
- should not in be startd by `php artisan servel:start` instead of `php artisan servel start` ?

gregorip02:
- Laravel recommends that the external package configuration be in a dedicated file.
- Unfortunately `servel:start` doesn't support workerman's input arguments. So I had to follow their convention 🤷‍♂

PR #4094, opened 2021-03-21
Do not test Rust on musl

msrd0:
I just noticed that you compile all rust frameworks for the `x86_64-unknown-linux-musl` target which is known to be much slower for most Rust programms due to something related to musl's allocator. Changing the target to debian, I saw performance uplift by factor 5 for the `rust/gotham` framework, which is probably more than what is caused by unpredictable Ryzen boosting behaviour:

**Before** (`alpine`)

```
benchmark=# select avg(value) from values where key_id = 3;
  avg  
-------
 90456
(1 row)
```

**After** (`debian:buster-slim`)

```
benchmark=# select avg(value) from values where key_id = 3;
  avg   
--------
 453782
(1 row)
```

### DIFF
--- a/rust/Dockerfile
+++ b/rust/Dockerfile
@@ -1,22 +1,13 @@
-FROM rust:1.50 AS build
+FROM rust:1.50-slim AS build
 
 WORKDIR /usr/src/app
 
-RUN apt-get update && apt-get -y install musl-tools --no-install-recommends \
- && apt-get clean \
- && rm -rf /var/lib/apt/lists/*
-
-RUN apt-get -y install musl-tools --no-install-recommends
-
-RUN rustup target add x86_64-unknown-linux-musl
-
 COPY . ./
 
-RUN cargo build --target x86_64-unknown-linux-musl --release
+RUN cargo build --release && strip target/release/server
 
-# hadolint ignore=DL3006
-FROM alpine
+FROM debian:buster-slim
 
-COPY --from=build /usr/src/app/target/x86_64-unknown-linux-musl/release/server /usr/src/app/target/x86_64-unknown-linux-musl/release/server
+COPY --from=build /usr/src/app/target/release/server /usr/src/app/target/release/server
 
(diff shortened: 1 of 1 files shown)


waghanza:
Hi @msrd0,

Thanks for checking. The idea behind this is to use `best-practices`. I mean since `docker` has switched to `alpine`, I'd rather use `alpine` (but this will be the case only for kubernetes, and maybe not for VMs).

The main question I have is : how about usage in `rust` world ? I mean which target is used ? Is `rust` deployed on containers (vs VMs) on **production**.

Regards,

msrd0:
> The main question I have is : how about usage in rust world ? I mean which target is used ? Is rust deployed on containers (vs VMs) on production.

I can only speak for myself, but I deploy all rust services in docker containers based on a debian-slim base image, using the default gnu toolchain. I've been using alpine in the past (with alpine's own rust compiler, not the static one available on rustup) but switched to debian when I noticed that building times (as the compiler also runs on musl) are about twice as fast on debian than on alpine. I'd love to use alpine as a base image, but to me the smaller container size does not make up for worse performance. Using a static toolchain like you are currently doing was never an option for most of my projects because I needed to dynamically link to `libpq.so`.

One alternative might be to switch the allocator to jemalloc, which used to be the default allocator for most targets before rust decided to use the system allocator by default in version 1.32. The rust compiler itself is still using jemalloc for its own code on gnu targets, but considers jemalloc unstable on musl. Besides that, I feel like I'd rather not play around with allocators, especially on non-embedded targets like docker containers.

waghanza:
may I ask @robjtede @JohnTitor @sezna @colinbankier @chrislearn for any insight ?

the question is : how do rust community (at least you guys) deploy `rust` on **production**.

my first take was to take `alpine`, but I prefer to change if this not aligned with `rust` **practices**

robjtede:
Personally and professionally use `bitnami/minideb` as base for docker images. Musl might have improved since I made that decision but I found some edge case incompatibilities with one or two -sys crates at the time.

waghanza (approved):

PR #4098, opened 2021-03-23
[Crystal] Upgrade to 1.0

waghanza:
Hi,

Finally, `crystal` has released **1.0** status 🎉 

###### Frameworks :

+ [x] shivneri
+ [x] orion
+ [x] amber /cc @drujensen @eliasjpr @elorest @faustinoaq @fridgerator @marksiemers @robacarp
+ [x] router.cr
+ [x] runcobo
+ [x] toro
+ [x] spider-gazelle
+ [x] grip
+ [x] kemal
+ [x] athena

Regards,


soveran:
@waghanza Thanks for the heads up! Toro is already compatible with Crystal 1.0, let me know if I have to do something else.

waghanza:
@soveran `toro` might be compatible with `crystal` **1.0** (even if `--ignore-crystal-version` could be required to compile any app base on it)
However, it is mostly about `seg` and `crotest`.
When using `shards`, I have :
```crystal
Resolving dependencies
Fetching https://github.com/soveran/seg.git
Fetching https://github.com/emancu/crotest.git
Unable to satisfy the following requirements:

- `crystal (< 1.0.0)` required by `seg 0.1.0+git.commit.8580a8b398300ad58088ff71c061515cf4be5149`
- `crystal (< 1.0.0)` required by `crotest 1.0.0+git.commit.61bdf96c29494b63f4d5e70d25eb999ad40b097c`
Failed to resolve dependencies, try updating incompatible shards or use --ignore-crystal-version as a workaround if no update is available.
```

soveran:
@waghanza I fixed it for Seg, but I can't modify Crotest myself. I have contacted the author to let him know about his issue. It should be an easy fix for him because it is already compatible with Crystal 1.0. That said, for my libraries Crotest is only a development dependency.

waghanza:
However, I do not see the point of https://github.com/soveran/toro/blob/bc96e97570ad22622b5ed9553f81296a4a1ee3df/.github/workflows/crystal.yml#L16 @soveran. 

Sadly, if appears that CI is ignoring a kind of protection offered by `shards`, so as we have to do the same when using `toro`

soveran:
@waghanza Absolutely. Is there any other way to deal with this particular situation? To sum up: a development dependency is 1.0 compatible but has not updated the target Crystal version in its shard.yml yet.

waghanza:
@soveran I think having something like `crystal: 0.36.1, ~> 1.0.0` in `shard.yml` could do the trick

Having `crystal: ~> 1.0.0` could be also done, but the advantage of above solution is to avoid **breaking** any _compilation_ for existing users

waghanza:
@soveran could you publish a new release of `toro` (to allow compilation in 1.0) ?

soveran:
@waghanza Done!

PR #4109, opened 2021-03-26
[R] Update plumber to v1.1

waghanza:
Hi @schloerke,

This `PR` update **plumber** in here.

May I notify you on each update that can affect `plumber` ?

Regards,

### DIFF
--- a/r/plumber/DESCRIPTION
+++ b/r/plumber/DESCRIPTION
@@ -1,2 +1,2 @@
 Depends:
-  plumber (>= 1.0.0)
+  plumber (>= 1.1.0)
--- a/r/plumber/config.yaml
+++ b/r/plumber/config.yaml
@@ -1,3 +1,3 @@
 framework:
   website: rplumber.io
-  version: 1.0
+  version: 1.1


schloerke:
@waghanza Sure. Thank you for reaching out.

The speed test looks bad. 😞  I don't know how we can have -84 req/s when wrk is sending 512. 🤯  I wonder if `plumber` should be removed.

waghanza:
Hi @schloerke ,

> The speed test looks bad. 😞 I don't know how we can have -84 req/s when wrk is sending 512. 🤯

Seems to be an internal issue here.

I wanted to take this `PR` to also fix that (plumber` seems not to be the only framework affected)

I have something like
```
Running 15s test @ http://172.17.0.2:3000
  8 threads and 256 connections
    Latency   546.78ms   64.95ms 613.80ms   96.08%
    Req/Sec    92.83     84.55   323.00     73.53%
  6907 requests in 15.04s, 681.26KB read
Requests/sec:    459.16
Transfer/sec:     45.29KB
```

but no successful queries @ 512 concurrency level

The results are bad 
| master  | r (4.0)| [plumber](https://rplumber.io) (1.0) | 286.11 | 298.58 | -84.76 |
| here  | r (4.0)| [plumber](https://rplumber.io) (1.1) | 285.57 | 304.22 | -92.13 |

for `plumber`

> I wonder if `plumber` should be removed.

Any time feel free to ping me, if you want to remove `plumber`.

PR #4136, opened 2021-04-02
[Python] Use built-in server for tornado

waghanza:
Hi,

This `PR` switch from `gunicorn` worker to `tornado` _built-in_ server.

The idea is to use what frameworks are, and only rely of external lib if non is built-in (and suitable for **production**).

Same idea for `sanic`.

Regards,

@bdarnell Can you confirm the built-in server is **production** _ready_ ? What is your recommandation, if not ?

### DIFF
--- a/python/tornado/config.yaml
+++ b/python/tornado/config.yaml
@@ -2,11 +2,4 @@ framework:
   website: tornadoweb.org
   version: 6.1
 
-command: >
-  gunicorn  \
-    --log-level warning  \
-    --bind 0.0.0.0:3000  \
-    --reuse-port  \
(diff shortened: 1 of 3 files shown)


ahopkins (approved):

bdarnell:
The built-in server is production ready so this is fine in that respect (in fact, I'm pretty sure gunicorn just uses the tornado http server in its tornado workers). However, the gunicorn configuration was multi-process and you've replaced it with a single process. To match the previous configuration, you need to replace the `app.listen(3000)` line with

```
server = tornado.httpserver.HTTPServer(app)
server.bind(3000)
server.start(0)
```

PR #4159, opened 2021-04-07
DriftPHP to persistent connections

mmoreram:
### Edited from this original

Somehow, only 1 worker is being used for this benchmark. By setting this value to 8, all available cores should be used now.

Added as well a more relaxed quiet mode for checking errors

### DIFF
--- a/php/driftphp/composer.json
+++ b/php/driftphp/composer.json
@@ -11,7 +11,8 @@
     "symfony/framework-bundle": "*",
     "symfony/yaml": "*",
     "drift/http-kernel": "~0.1.7",
-    "drift/server": "~0.1.17"
+    "drift/server": "~0.1.23",
+    "react/http": "dev-master#7aa08f01583c765261d08d1191f0239324f18012 as 1.2.1"
   },
   "autoload": {
     "psr-4": {
--- a/php/driftphp/config.yaml
+++ b/php/driftphp/config.yaml
@@ -10,4 +10,4 @@ php_mod:
   - posix
   - sockets
 
-command: php vendor/bin/server run 0.0.0.0:3000 --env=prod --quiet --workers=-1
+command: php vendor/bin/server run 3000 --env=prod --quiet --workers=-1


waghanza:
> Somehow, only 1 worker is being used for this benchmark. By setting this value to 8, all available cores should be used now.

You configured `driftpjp` like this :-)
https://github.com/the-benchmarker/web-frameworks/pull/3819/files#diff-e159b7f16679097dc8746a4c56a8f40a9cfe1ef0affce32d77d66aaa94dd5e47R13

php/driftphp/config.yaml
```diff
@@ -10,4 +10,4 @@ php_mod:
   - posix
   - sockets
 
-command: php vendor/bin/server run 0.0.0.0:3000 --env=prod --quiet --workers=-1
+command: php vendor/bin/server run 0.0.0.0:3000 --env=prod --quiet --workers=8
```
waghanza, line 13: better use a dynamic calculation, no ?

I mean either `nproc` (for the cli) or any trick to calculate the core count in php

waghanza (approved):

php/driftphp/config.yaml
mmoreram: -1 means "core count" with nproc.

waghanza: do not understand. you set `-1` to use all cores, should be 8 (benchmarks are done a 8 CPU xeon)

mmoreram: Exactly, but seems that something is not working. This framework is a high performance PHP framework and is listed almost the last PHP one. These numbers make sense with 1 single core, but not with 8. I will try with specifically 8 cores, so we can detect some kind of issue in that `-1`.

waghanza: I convert this to draft -> meaning (for me), that this requires investigation. I'll try to post you results here, with both configuration

mmoreram: This server configuration is set as `--quiet`, which means that there will be nothing in the console. I could add a new option that allow us to know the server header and all errors. That could work,

waghanza: I mean I will output results : 8 workers vs current configuration

mmoreram: I updated the PR with this change. Now is not quiet anymore, but we'll be able to see errors and the server header

waghanza:
@mmoreram With your last modification, I have computed the following :

|                                   |  Speed (64) | Speed (256) | Speed (512) |
|---------------------|-------------- | -----------|-------------|
| with **-1** workers |  3 432.47 | 3 983.21 | 4 766.13 |
|  with **8** workers  | 3 409.90 | 4 263.83 | 4 634.95 |

mmoreram:
@waghanza do you have the output? In my local, using just one CPU I'm having the same number. I'm sure that it's only using 1 core instead of N.

waghanza:
figure are indeed weird, let me take a look.

Raw **wrk** says `9815.53` for `-1`

waghanza:
with a manual execution, I have :
+ for `GET` on `/` => `wrk -H 'Connection: keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 --script pipeline.lua http://172.17.0.2:3000`
```
Running 15s test @ http://172.17.0.2:3000
  8 threads and 64 connections
    Latency     7.50ms    6.13ms  66.16ms   78.89%
    Req/Sec   723.26    526.72     1.75k    69.72%
  84918 requests in 15.10s, 9.23MB read
Requests/sec:   5623.23
Transfer/sec:    626.02KB
```
+ for `POST` on `/user` => `wrk -H 'Connection: keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 --script pipeline_post.lua http://172.17.0.2:3000/user`
```
Running 15s test @ http://172.17.0.2:3000/user
  8 threads and 64 connections
    Latency     6.11ms    4.22ms  47.42ms   74.36%
    Req/Sec     1.08k   495.67     1.74k    62.15%
  127873 requests in 15.10s, 13.90MB read
Requests/sec:   8468.72
Transfer/sec:      0.92MB
```
+ for `GET` on `/user/0` => `wrk -H 'Connection: keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 --script pipeline.lua http://172.17.0.2:3000/user/0`
```
Running 15s test @ http://172.17.0.2:3000/user/0
  8 threads and 64 connections
    Latency     7.00ms    5.65ms  57.17ms   80.89%
    Req/Sec     0.87k   566.39     2.57k    51.48%
  101956 requests in 15.09s, 11.18MB read
Requests/sec:   6754.78
Transfer/sec:    758.59KB
```

mmoreram:
That's mine. Okay then. We will assume that there's a difference then. I don't know what could cause this 2/3 of performance reduction, but ok.

```
wrk -t 8 -d 15 --timeout 1 -H 'Connection: keep-alive'  -c 64 http://localhost:3000/
Running 15s test @ http://localhost:3000/
  8 threads and 64 connections
    Latency     2.58ms    2.33ms  15.23ms   80.39%
    Req/Sec     3.60k   356.45     9.00k    86.70%
  431128 requests in 15.10s, 46.87MB read
Requests/sec:  28552.18
Transfer/sec:      3.10MB
```

waghanza:
- when using a **container**
```
Running 15s test @ http://172.17.0.2:3000/
  8 threads and 64 connections
    Latency     6.25ms    4.85ms  46.28ms   78.81%
    Req/Sec     1.04k   570.00     1.92k    57.16%
  123269 requests in 15.08s, 13.40MB read
Requests/sec:   8173.32
Transfer/sec:      0.89MB
```
- when using local
```
Running 15s test @ http://127.0.0.1:3000/
  8 threads and 64 connections
    Latency     5.03ms    3.36ms  40.20ms   71.10%
    Req/Sec     1.58k   108.55     2.08k    73.41%
  189184 requests in 15.03s, 20.57MB read
Requests/sec:  12586.27
Transfer/sec:      1.37MB
```
- the most important thing is that we need to use a cloud provider (real network, real hardware ...) to get some strong conclusions about any tools in this list

waghanza:
can we close for now @mmoreram ?

mmoreram:
Reusing the PR. The number of workers did not change, but we updated dependencies and started reusing connections.

waghanza:
I think then :
- that using `-1` for worker num is better (results will be not be computed ever on 8 CPU machines)
- that using `almost-quiet` could affect results

mmoreram:
Yes, `--workers=-1` is back, and I turned the server completely `--quiet`. Ready then :dagger: 

🤞

waghanza:
> Yes, `--workers=-1` is back, and I turned the server completely `--quiet`. Ready then :D

Did you forget to push ? I see no modification

mmoreram:
@waghanza LOL. I was pushing to master.
Sorry!
Done :)

php/driftphp/composer.json
```diff
@@ -11,7 +11,8 @@
     "symfony/framework-bundle": "*",
     "symfony/yaml": "*",
     "drift/http-kernel": "~0.1.7",
-    "drift/server": "~0.1.17"
+    "drift/server": "~0.1.23",
+    "react/http": "dev-master#7aa08f01583c765261d08d1191f0239324f18012 as 1.2.1"
```
waghanza, line 15: do we need this @mmoreram ?

waghanza (changes requested):

php/driftphp/composer.json
mmoreram: Not anymore. There is a new version release from yesterday, in the next result set it will be automatically updated.

Closing the PR.

PR #4167, opened 2021-04-11
[R] Update rserve to 1.8

waghanza:
Hi @s-u,

I noticed that a new version of https://github.com/s-u/Rserve exists.

This `PR` update to **1.8**.

May I notify you on each change that can affect https://github.com/s-u/Rserve ?

Regards,

### DIFF
--- a/r/rserve/DESCRIPTION
+++ b/r/rserve/DESCRIPTION
@@ -1,2 +1,2 @@
 Depends:
-  Rserve (>= 1.7-3)
+  Rserve (>= 1.8)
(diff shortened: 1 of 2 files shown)


s-u:
@waghanza Although 1.8 series is the latest version and it's available from http://RFroge.net, the CRAN version is still on 1.7 (for historical reasons), so I wouldn't push up the requirement as it would exclude the CRAN version. Useers can still install the latests version. Performance-wise I don't think there should be a difference.

waghanza:
how can user install the latest version @s-u ?

PS : I have no skill about r` **ecosystem**

s-u:
@waghanza same ay as any other - just use `https://rforge.net` as the repository/mirror, so

```r
install.packages("Rserve",,"https://rforge.net")
```

waghanza:
and could the mirror be specified in `DESCRIPTION` file @s-u ?

PR #4174, opened 2021-04-13
Also export (results) as json

waghanza:
Hi,

This `PR` export results as json, file `data.json`.

I also update results in here.

###### Frameworks to rerun :
+ [x] roda
+ [x] luminus
+ [x] laravel
+ [x] driftphp
+ [x] quarkus

Regards,

/cc @SuspiciousLookingOwl

PR #4178, opened 2021-04-15
fix typo minimim to minimum

SuspiciousLookingOwl:
From #4174, I noticed there are output typo on the `.json` file, metric label `minimim_latency` supposed to be `minimum_latency`. This PR fixes the typo.


waghanza (approved):
❤️

PR #4192, opened 2021-04-20
[Nim] Update mike to 1.0

waghanza:
Hi @ire4ever1190,

This `PR` update https://github.com/ire4ever1190/mike to 1.0

May I notify you on each update @ire4ever1190 ?

Regards,

### DIFF
--- a/nim/mike/config.yaml
+++ b/nim/mike/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: ire4ever1190/mike
-  version: 0.5
+  version: 1.0
 
 files:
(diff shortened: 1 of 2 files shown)


ire4ever1190:
Yes, thanks for notifying me because Mike has changed a lot in the new update and the old code won't work. Planning another update real quick to add in the missing helper functions real soon

Thanks again for this 👍

waghanza:
I plan a new update today (I mean results update), I need then to skip `mike` for this round (next one is planned for next week)

PR #4204, opened 2021-04-22
README - very minor change

tylercaceres:


waghanza (changes requested):
Thanks 🎉 

Could you replicate this on `README.mustache.md`, the current file is generated

PR #4208, opened 2021-04-24
[Nim] Update mike to 1.1

ire4ever1190:
This supersedes  #4192  

this updates https://github.com/ire4ever1190/mike to 1.1.0 and updates the file here to reflect the new changes

### DIFF
--- a/nim/mike/config.yaml
+++ b/nim/mike/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: ire4ever1190/mike
-  version: 0.5
+  version: 1.1.0
 
 files:
   - /usr/src/app/server
--- a/nim/mike/config.yaml
+++ b/nim/mike/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: ire4ever1190/mike
-  version: 1.1.0
+  version: 1.1
 
 files:
   - /usr/src/app/server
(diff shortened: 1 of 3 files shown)


nim/mike/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: ire4ever1190/mike
-  version: 0.5
+  version: 1.1.0
```
waghanza, line 3: I prefer to use `X.Y` format
```suggestion
  version: 1.1
```

nim/mike/server.nimble
```diff
@@ -7,5 +7,5 @@ license = "MIT"
 
 # Dependencies
 
-requires "nim >= 1.2.2"
-requires "https://github.com/ire4ever1190/mike.git >= 0.5 & < 0.6"
+requires "nim >= 1.4.2"
+requires "https://github.com/ire4ever1190/mike.git == 1.1.0"
```
waghanza, line 11: I think we should allow patch level fix

```suggestion
requires "https://github.com/ire4ever1190/mike.git >= 1.1 & < 1.2"
```

waghanza (changes requested):

waghanza (approved):

PR #4209, opened 2021-04-24
[Nim] Add more optimisations

ire4ever1190:
* Add -d:danger to remove more runtime checks
* Add link time optimisation

### DIFF
--- a/nim/Dockerfile
+++ b/nim/Dockerfile
@@ -19,6 +19,9 @@ RUN nimble install -y
 RUN nim c {{#build_opts}} {{{.}}} {{/build_opts}} \
     --excessiveStackTrace:off \
     -d:release \
+    -d:danger \
+    --passC:-flto \
+    --passL:-flto \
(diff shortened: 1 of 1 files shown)


ire4ever1190:
The CI seems to fail since it is still trying to use the older version of mike which won't be a problem once #4208 is merged

waghanza (commented):
Thank for this. Honestly, I can not propose any solution / confirm we should use this 😛 

What is a common usage in `nim` ? I mean, is it really safe to use something like `-d:danger` in **production** ?

ire4ever1190:
`-d:danger` turns off checks (so assert statements and things like that). It's usually recommended for production builds so that you have the most performance
But no it **wouldn't be** safe in production since it removes asserts so I'll remove it

waghanza:
could you merge (or rebase) to trigger CI with `mike` **1.1** ?

waghanza (approved):

PR #4211, opened 2021-04-24
[Javascript] Add tinyhttp

waghanza:
Hi @talentlessguy,

Following https://github.com/talentlessguy/tinyhttp/issues/246, this `PR` adds https://github.com/talentlessguy/tinyhttp in `here`.

Actually, this implementation is very simple, but will change in the future.

Regards,

### DIFF
--- a/javascript/tinyhttp/app.js
+++ b/javascript/tinyhttp/app.js
@@ -1,0 +1,15 @@
+import { App } from '@tinyhttp/app'
+
+const app = new App()
+
+app
+  .get('/', (_, res) => {
+    res.send('')
+  })
+  .get('/user/:id', (req, res) => {
+    res.send(`${req.params.id}`)
+  })
+  .post('/user', (req, res) => {
+    res.end("");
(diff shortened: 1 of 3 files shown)


talentlessguy:
to make it even more performant u can pass an empty function to `applyExtensions`:

```js
new App({ applyExtensions: (req, res, next) => void next() })
```

waghanza:
what is your recommend for a **production** usage @talentlessguy ? despite, this kind of metrics shown, to have a fair (at least _as-fair-as_ we can), it could be better to reflect what the frameworks can do in **production**

talentlessguy:
@waghanza I think 90% of tinyhttp users don't use that feature so probably no need for custom `applyExtensions` function. Leave it as it is then :D

talentlessguy:
@waghanza for some reason it's not in readme 🤔

waghanza:
in a few days, I need to compute results. I try to do so once a week at least

PR #4240, opened 2021-05-02
Disables OGNL SecurityManager on init to improve performance

lukaszlenart:
Hopefully this should improve performance a bit, current place is 181 - we will see :) 
Relates to https://github.com/jkuhnert/ognl/pull/69

### DIFF
--- a/java/struts2/config.yaml
+++ b/java/struts2/config.yaml
@@ -8,4 +8,4 @@ build:
 binaries:
   - target/benchmark-1.0.0-standalone.jar
 
-command: java -XX:+UseNUMA -XX:+UseParallelGC -jar target/benchmark-1.0.0-standalone.jar --port 3000
+command: java -Dognl.security.manager=forceDisableOnInit -XX:+UseNUMA -XX:+UseParallelGC -jar target/benchmark-1.0.0-standalone.jar --port 3000


waghanza:
What OGNL stands for ? I'm not sure this should be included, afaik.

Correct me if I'm wrong, seems related to security, and then not recommended to disable this in **production**.

I mean tunning for performance is great, but I prefer this project reflect what could be in **production**

lukaszlenart:
OGNL is an expression language used by Struts to evaluate almost everything, eg. applying parameters to actions, calling actions, and so on.

And by default Security Manager isn't configured, each user has to [enable](https://struts.apache.org/security/#run-ognl-expressions-inside-sandbox) it by yourself and define appropriate policies. This setting just disables checking if SM is enabled and then calling each evaluation in a sandbox - with this option on, the whole boilerplate is skipped.

I'm not sure if I could disable this by default in Struts itself.

waghanza:
Ok. So, as I understand, this is mostly for `struts`. I let you give any recommendation for this (having in mind this implementation should be **production-ready** ...).

3 things however :
- having some options saying `disable` mean that something could be unsafe, imho
- I think this should be included in `struts` code base, if this is only related to `struts`
- better have same options as jvm args (disclaimer : work should be done in that direction)

lukaszlenart:
This isn't Struts specific, [OGNL](https://github.com/jkuhnert/ognl) is an external dependency used by many other libraries/frameworks and this option cannot be set in Struts codebase because of nature how JVM initialises static sections.
Anyway, I will try figure it out how to change this in OGNL itself.

PR #4304, opened 2021-05-17
corrects blacksheep

RobertoPrevato:


waghanza (changes requested):
seems to fail @RobertoPrevato 
https://the-benchmarker.semaphoreci.com/jobs/6f99d0c4-c1b0-4bff-befa-704c44aa11b1

waghanza:
seems to be related to dependency version between https://github.com/encode/uvicorn/blob/0.13.0/setup.py#L54 and https://github.com/Neoteroi/BlackSheep/blob/v1.0.5/setup.py#L87

waghanza:
waiting for https://github.com/encode/uvicorn/pull/1024
/cc @tomchristie

RobertoPrevato:
@waghanza this is weird, when I tested locally a few hours ago, I installed the dependencies using the `requirements.txt` in the repository, and I confirm it works with `uvicorn 0.13.4` and the last version of my web framework (I just double checked).
I doubt it's a problem of dependencies, I am using the latest version of `pip` (21.1.1). I suspect `pip` gets confused because of the missing wheels for `alpine OS`.

I try to reproduce the case with `python:3.9-alpine` in docker.

waghanza:
did you test with `uvicorn` ? or `uvicorn[standard]` ? @RobertoPrevato 

`httptools` is only required (on `uvicorn`) when using `[standard]` => the **cython** based tools

RobertoPrevato:
@waghanza I tested with `uvicorn[standard]`: the problem doesn't occur on Ubuntu 18.04, I am not sure why, but I suspect it's because the `wheel` packages handle the different versions in the dependencies, and the provided wheels work for Ubuntu, Windows, and Mac, but not Alpine.

I just verified that downgrading `blacksheep` to version `1.0.3` works on `python:3.9-alpine`. 
* blacksheep since 1.0.4 requires `httptools` 0.2.0 simply because it adds compatibility with `Python 3.10`

Let's please try again a build downgrading to `blacksheep==1.0.3` (I just pushed this change).

RobertoPrevato:
@waghanza I locked the dependency to 1.0.3 before you confirmed this is the desired approach. I see this last build fails, but not for blacksheep.

PR #4334, opened 2021-05-24
[PHP] BasicPHP v0.9.9

ray-ang:


php/basicphp/index.php
```diff
@@ -0,0 +1,21 @@
+<?php
```
waghanza, line 1: This file should be in a subdirectory name `public`

php/basicphp/config.yaml
```diff
@@ -1,5 +1,5 @@
 framework:
   github: ray-ang/basicphp
-  version: 0.9
+  version: 0.9.9
```
waghanza, line 3: please avoid patch level

waghanza (changes requested):

php/basicphp/index.php
ray-ang: Thanks. I placed the front controller file inside the public folder.

php/basicphp/config.yaml
ray-ang: Thanks. I reverted the line back to version level -> 0.9

waghanza (approved):

PR #4352, opened 2021-05-28
[Java] Update restheart to v5.4

waghanza:


ujibang:
Fine to update to 5.4, but we'd like to update to 6 (that requires java 16, in preparation for upcoming LTS java 17 expected in September)

waghanza:
I understand, but I prefer to stick on LTS (`node`,`jdk` ...)
- very hard to maintain
- meaningful to have frameworks using the same version

Issue #4405, opened 2021-06-16
Missing java / activej in results


eduard-vasinskyi:
It seems like after the recent run (2021-06-03), ActiveJ, as well as some other frameworks, are missing from [the test results page](https://web-frameworks-benchmark.netlify.app/result)

waghanza:
Hi @eduard-vasinskyi,

I've seen ... that some frameworks are missing (not only `activej`). I'll fix this for the next update (this week).

Thanks for reminder ❤️

Issue #4384, opened 2021-06-06
Missing ruby / rails in results


pitis:
Rails is one of the biggest frameworks out there, and we had results before. Can someone add it?

PR #4408, opened 2021-06-16
Results update

### LINKED ISSUE #4378: Missing nim / scorper in results

waghanza:
Hi,

This `PR` will update results.

Regards,

-------
Closes #4378, #4384, #4405, #4437


waghanza:
@dom96 @mattaylor for both `httpbeast` and `whip`. I have a `Address family not supported by protocol [OSError]` on **alpine** (perhaps **alpine** is not the right distro for `nim`, I have no insights to choose).

Compilation options are the same for both frameworks 

```dockerfile
FROM nimlang/nim:1.4.8-alpine

WORKDIR /usr/src/app

COPY . ./



ENV PATH $PATH:/root/.nimble/bin

RUN nimble install -y

RUN nim c  \
    --excessiveStackTrace:off \
    -d:release \
    --passC:-flto \
    --passL:-flto \
    server.nim

FROM alpine

COPY --from=0 /usr/src/app/server /usr/src/app/server

CMD /usr/src/app/server
```

waghanza:
@greenlaw110 When running `act` I have
```
WARNING: An illegal reflective access operation has occurred
WARNING: Illegal reflective access by com.esotericsoftware.reflectasm.AccessClassLoader (file:/opt/web/lib/reflectasm-1.11.9.jar) to method java.lang.ClassLoader.defineClass(java.lang.String,byte[],int,int,java.security.ProtectionDomain)
WARNING: Please consider reporting this to the maintainers of com.esotericsoftware.reflectasm.AccessClassLoader
WARNING: Use --illegal-access=warn to enable warnings of further illegal reflective access operations
WARNING: All illegal access operations will be denied in a future release
```

waghanza:
@dom96 @mattaylor I'll skip `httpbeast` and `whip` for this update

dom96:
@waghanza that's weird. Did httpbeast work fine before? Maybe it's the new version that has a regression, can you try 0.2.2?

waghanza:
We are still using 0.2.x 😛  If we solve this issue this week end, I can update `httpbeast`, I just prefer not to block a lot of frameworks due to 2 of them ❤️

waghanza:
@dom96 it is indeed a regression 0.3. Using `requires "httpbeast >= 0.2 & < 0.3"` fix the issue

mattaylor:
I can help look at this in the morning.

jeremyevans:
The Ruby issue is unrelated to any of the frameworks, it's an issue with reaching the rubygems repository during bundle install.  Only causes I can think of for that are a temporary issue with rubygems itself, or the docker instance not being able to reach rubygems due to a network issue.

dom96:
If you'll be using Debian can you switch back to httpbeast 0.3.0?

waghanza:
> If you'll be using Debian can you switch back to httpbeast 0.3.0?

I'll test this today :-)

dom96:
> PSS : The best for this project, imho, is to stick to want mostly used (in term of distro), do you have any insight for me @dom96 ? (I mean what is mostly used in nim community)

I think it should be whatever is most popular among all devs. Debian-based seems like would be most popular. I personally use Ubuntu the most (and in all my CIs too IIRC).

Definitely weird we're seeing problems on Alpine though, I'm guessing it's because of musl, would be nice to figure out what the change that caused the issue is.

waghanza:
@dom96 seems that the error was on my side (https://www.avahi.org disable `dns` on **containers** due to security / selinux)

My solution is to use https://podman.io

dom96:
Debian + 0.3.0 doesn't work?

waghanza:
> Debian + 0.3.0 doesn't work?

Nop. Same issue. I have opened a gh issue to track this subject

PR #4415, opened 2021-06-19
Fix website for Flask

ralexander-phi:
See https://www.pocoo.org/. Flask is now hosted by palletsprojects.com


waghanza (approved):
:heart:

PR #4471, opened 2021-07-15
Reduce containers size

waghanza:
Hi,

This `PR` will drastically decrease build time (at least on result publication, not CI).

The idea is to build extensions in separate containers.

For example, `go` produce only one binary, and this binary is upload in a final containers.

In `php`, all extensions are compiled in final container. The idea is to compile those extensions in intermediate containers, in order to :
+ leverage docker caching on publication process
+ have cleaner / slimmer final containers

Regards,

/cc @sy-records @leocavalcante

### DIFF
--- a/php/Dockerfile
+++ b/php/Dockerfile
@@ -1,3 +1,9 @@
+{{#php_ext}}
+  FROM php:7.4-fpm-alpine AS {{{.}}}
+  RUN apk --no-cache add pcre-dev ${PHPIZE_DEPS}
+  RUN pecl install {{{.}}}
+{{/php_ext}}
+
 FROM php:7.4-fpm-alpine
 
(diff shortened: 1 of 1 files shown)


sy-records (commented):
LGTM, But there are some extensions that fail to install, such as
```
configure: error: Please reinstall the event library, or provide the installation prefix via --with-event-libevent-dir option00:15
ERROR: `/tmp/pear/temp/event/configure --with-php-config=/usr/local/bin/php-config --enable-event-debug=no --enable-event-sockets=yes --with-event-libevent-dir=/usr --with-event-pthreads=no --with-event-extra --with-event-openssl --with-event-ns=no --with-openssl-dir=no' failed00:15
The command '/bin/sh -c pecl install event' returned a non-zero code: 100:15
make: *** [.Makefile:2: build] Error 1
```

waghanza:
indeed, `event` depends on `libevent` (actually we say that the framework depends on it -> but it's more the setup that depends on it)

PR #4475, opened 2021-07-16
Add Prologue

Willyboar:
PR adds Prologue (one of the most famous nim web frameworks) to the benchmark

@xflywind agreed to add it, but I would like to review it too

### DIFF
--- a/nim/prologue/server.nim
+++ b/nim/prologue/server.nim
@@ -1,0 +1,19 @@
+import prologue
+
+proc simpleGet*(ctx: Context) {.async.} =
+  resp ""
+
+proc userGet*(ctx: Context) {.async.} =
+  resp ctx.getPathParams("id")
+
+proc simplePost*(ctx: Context) {.async.} =
+  resp ""
+
+let settings = newSettings(port = Port(3000))
+
+let app = newApp(settings = settings)
+
+app.get("/", simpleGet)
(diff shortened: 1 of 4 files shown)


nim/prologue/server.nimble
```diff
@@ -0,0 +1,11 @@
...
+
+version = "0.1.0"
+author = "Willyboar"
+description = "Prologue Implementation"
+license = "MIT"
+
+# Dependencies
+
+requires "nim >= 1.4.0"
+requires "prologue >= 0.4"
```
waghanza, line 11: to be sure we are still on 0.4.x, we should use something like
```suggestion
requires "prologue >= 0.4 & < 0.5"
```

waghanza (approved):

nim/prologue/server.nim
```diff
@@ -0,0 +1,19 @@
...
+proc simpleGet*(ctx: Context) {.async.} =
+  resp ""
+
+proc userGet*(ctx: Context) {.async.} =
+  resp ctx.getPathParams("id")
+
+proc simplePost*(ctx: Context) {.async.} =
+  resp ""
+
+let settings = newSettings(port = Port(3000))
```
xflywind, line 12: Use `let app = newApp(settings = settings, debug = false)` to disable logging.

PR #4487, opened 2021-07-19
[Go] Add Buffalo

D-system:


waghanza:
Hi @D-system,

Thanks for your contribution.

However, I can not accept it. 

We must have consent (or no opposition) from tech team / author, and I'm afraid @markbates does not want to have `buffalo` included here.

Regards,

D-system:
Hi @waghanza 
Thank you for your prompt reply and clarification. I wasn't aware of the opposition of the author.

Best

PR #4552, opened 2021-08-14
[Go] Add apirouter

### LINKED ISSUE #4508: [Go] Add ApiRouter
apirouter provides a lightning fast RESTful api router.

Author: cnotch
Link: https://github.com/cnotch/apirouter/issues/1

waghanza:
Hi @cnotch,

This `PR` adds https://github.com/cnotch/apirouter to our benchmark.

I'm not specialized in `go` so I could find what I need to write a proper implementation\.

Feel free to correct me.

---------
Closes #4508

### DIFF
--- a/go/apirouter/main.go
+++ b/go/apirouter/main.go
@@ -1,22 +1,25 @@
 package main
 
 import (
-	"fmt"
 	"github.com/cnotch/apirouter"
 	"net/http"
+	"io"
 )
+
 
 func main() {
 	r := apirouter.New(
 		apirouter.API("GET", "/", func(w http.ResponseWriter, r *http.Request, ps apirouter.Params) {
-			fmt.Fprint(w, "")
+			io.WriteString(w, "")
+
 		}),
 		apirouter.API("GET", "/user/:id", func(w http.ResponseWriter, r *http.Request, ps apirouter.Params) {
 			id := ps.ByName("id")
-			fmt.Fprint(w, id)
+			io.WriteString(w, id)
 		}),
 		apirouter.API("POST", "/user", func(w http.ResponseWriter, r *http.Request, ps apirouter.Params) {
-			fmt.Fprint(w, "")
+			io.WriteString(w, "")
+
 		}),
 	)
 
--- a/go/apirouter/main.go
+++ b/go/apirouter/main.go
@@ -1,0 +1,18 @@
+package main
+
+import "github.com/cnotch/apirouter"
+
+	r := apirouter.New(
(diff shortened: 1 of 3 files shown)


waghanza (commented):
How can (or is it possible to) use `go` **modules** ?

System-Glitch:
> 
> 
> How can (or is it possible to) use `go` **modules** ?

It looks like you are already using it as module. What do you want to know about go modules?

waghanza:
I use `go` **cli** to generate module version. However, it looks like a commit hash

Is there not a standard way to publish package in `go` (tagging or else) ?

System-Glitch:
It is using the commit hash because this library doesn't have any git tag (release), so go modules falls back to a commit hash to know exactly which version is required.

If you want all the details, read [this](https://blog.golang.org/using-go-modules).

cnotch:
Hi @waghanza  ,Thank you for your work. 
Since Apirouter is a module for another project, I haven't released any versions. I'm releasing a version now，for easy reference to modules.

waghanza:
Thanks @cnotch. Not this `PR` could be included, but could you review it, plz ?

I'm no `go` dev, but seems that `ResponseWriter` could output directly bytes, instead of relying on `fmt` module.

I don't know which is considered as a best-practice in `go`

cnotch:
@waghanza , Just keep it consistent with httprouter for the benchmark test. The production environment may writes to the buffer and then outputs directly bytes to w. Such as:
``` go
func jsonTo(w io.Writer, o interface{}) error {
	formatted := buffers.Get().(*bytes.Buffer)
	formatted.Reset()
	defer buffers.Put(formatted)

	body, err := json.Marshal(o)
	if err != nil {
		return err
	}

	if err := json.Indent(formatted, body, "", "\t"); err != nil {
		return err
	}

	if _, err := w.Write(formatted.Bytes()); err != nil {
		return err
	}
	return nil
}
```

go/apirouter/main.go
```diff
@@ -0,0 +1,27 @@
+package main
+
+import (
+	"github.com/cnotch/apirouter"
+	"net/http"
+	"io"
+)
+
+
+func main() {
```
waghanza, line 10: @cnotch @System-Glitch I've changed the main method. As I understand this way is considered as "standard usage" in `go` 
https://pkg.go.dev/net/http#example-ResponseWriter-Trailers

Feel free to correct me if I'm wrong

waghanza: PS : There is no need to output `json` (at least yet). The idea is to output `plain/text`

cnotch: Direct IO theoretically provides better performance when formatting is not required

PR #4564, opened 2021-08-21
Update imi 2.0

Yurunsoft:


Yurunsoft:
@waghanza There is a limit on the number of CI, why not use Github Acitons?

waghanza:
> @waghanza There is a limit on the number of CI, why not use Github Acitons?

I had try GA once, but not very satisfied (network issues ...)

However, this is not be an issue anymore after https://github.com/the-benchmarker/web-frameworks/pull/3958

PR #4571, opened 2021-08-22
[PHP] Antidot: fix dependency issues in composer json

kpicaza:
Hi @waghanza, I see that there is some issue running Antidot Framework inner the benchmark. 
image

I updated the composer.json file to get at least the minimum required versions.

Thanks for your great work and for maintaining this excellent project;-D

### DIFF
--- a/php/antidot/composer.json
+++ b/php/antidot/composer.json
@@ -18,7 +18,8 @@
     "antidot-fw/fast-router-adapter": "*",
     "antidot-fw/react-framework": "^1.1",
     "antidot-fw/symfony-config-translator": "^1.1",
-    "antidot-fw/yaml-config-provider": "*"
+    "antidot-fw/yaml-config-provider": "*",
+    "react/http": "1.2.0"
   },
   "autoload": {
     "psr-4": {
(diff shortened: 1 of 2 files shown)


php/antidot/composer.json
```diff
@@ -13,13 +13,12 @@
     "ext-json": "*",
     "antidot-fw/framework": "~0.2.1",
     "antidot-fw/antidot-react-psr15": "*",
-    "antidot-fw/cli": "*",
+    "antidot-fw/cli": "^1.1",
```
waghanza, line 16: might the keyword `@stable` could work ?

waghanza (commented):
I will update results, after merge

php/antidot/composer.json
waghanza: nevermind, could not

waghanza (commented):
if you may invite me @kpicaza, I'll update results

kpicaza:
Sorry @waghanza, I'm debugging in more depth because the server falls after some stress tests, If you don't mind I prefer to check better the issue before merging, and for the moment moving the PR to draft, sorry for the inconvenience.

image

kpicaza:
I think the issue is related to keep-alive connections on React HTTP server as described here https://github.com/reactphp/http/pull/405#issuecomment-820388257. As a workaround, I will keep the react/http version fixed to v1.2.0. 

The performance is 2x better after v1.3.0, from 20k to 40k in Antidot Framework. But after some WRK hits the server crashes;-(.

waghanza (approved):

PR #4581, opened 2021-08-25
[C++] Update drogon to v1.7.2

waghanza:
/cc @marty1885


cpp/drogon/config.yaml
```diff
@@ -24,7 +24,7 @@ binaries:
   - build/drogon_benchmark
 
 download:
-  - git clone --branch v1.7.1 https://github.com/an-tao/drogon
+  - git clone --branch v1.7.2 https://github.com/an-tao/drogon
```
marty1885, line 27: Thanks for updating the version for us! May you update the URL to `https://github.com/drogon-framework/drogon`? We moved to organizations about a month ago. The old URL should still work but who knows when GitHub will change that.

Thanks!

marty1885:
@waghanza  Crap. By bad. `https://github.com/drogonframework/drogon`. I added an extra hyphen.

waghanza:
done @marty1885 + update results

PR #4601, opened 2021-08-30
Update mixphp fast_mode

onanying:


php/mixphp-workerman/src/Vega.php
```diff
@@ -14,6 +14,7 @@ class Vega
     public static function new(): Engine
     {
         $vega = new Engine();
+        $vega->mode(Engine::FAST_MODE);
```
waghanza, line 17: what is **fast mode** @onanying ?

onanying: @waghanza In fast mode, the user $request object is allowed to directly use Swoole or Workerman native request object to obtain parameters, which has better performance

onanying: No other changes

waghanza (approved):

PR #4640, opened 2021-09-11
[PHP] Antidot react: Fix max concurrency issues

kpicaza:
Hi @waghanza, after some research, I met that the allowed max concurrency were overloading PHP's memory since it breaks, I make some test locally and I look nice performance improvement by incrementing the PHP memory limit to 256M and by limiting the React-PHP HTTP server max concurrency to 512 requests.

image

Thanks again for your great work;-D.

### DIFF
--- a/php/antidot/config.yaml
+++ b/php/antidot/config.yaml
@@ -20,4 +20,4 @@ before_command:
   - mkdir -p var/cache && chmod 777 var/cache
   - mkdir -p var/log && chmod 777 var/log
 
-command: bin/console server:run
+command: bin/console server:run --quiet
--- a/php/antidot/config/config.php
+++ b/php/antidot/config/config.php
@@ -5,7 +5,8 @@ declare(strict_types=1);
 use Antidot\SymfonyConfigTranslator\Container\Config\ConfigAggregator;
 use Antidot\Yaml\YamlConfigProvider;
 use Laminas\ConfigAggregator\ArrayProvider;
-use Laminas\ConfigAggregator\PhpFileProvider;
+
+ini_set('memory_limit', '256M');
 
 $cacheConfig = [
     'config_cache_path' => 'var/cache/config-cache.php',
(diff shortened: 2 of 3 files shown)


php/antidot/composer.json
```diff
@@ -12,11 +12,11 @@
     "ext-intl": "*",
     "ext-json": "*",
     "antidot-fw/framework": "~0.2.1",
+    "antidot-fw/react-framework": "~1.1",
```
kpicaza, line 15: Added at least the latest antidot react integration version for better performance.

php/antidot/config.yaml
```diff
@@ -8,9 +8,6 @@ build_deps:
   - libicu-dev
   - libssl-dev
 
-php_ext:
```
kpicaza, line 11: Removed because it fails on the build process.

image

kpicaza:
Currently, I have to dig more in-depth to find what piece is creating the memory leak between requests.

php/antidot/composer.json
waghanza: now that reactphp is > 1, might you use `@stable` ?

php/antidot/config.yaml
kpicaza: > antidot does not rely on libevent ? I thought it was a requirement for reactphp

React-PHP checks for extensions enabled to select the event loop implementation. When any of the described extensions aren't available, it relies on the `StreamSelectLoop` class, based on the  `stream_select` function. For more info about the different event loop implementations, you can take a look at React-PHP Event Loop Docs https://github.com/reactphp/event-loop#loop-implementations.

kpicaza: The warning disappears using the Alpine image, and we can enable the event extension again, thanks;-D

kpicaza:
Hi @waghanza, I think that we can merge this PR while I seek the real problem. With this workaround, the application will respond well to the WRK tests without reaching the memory limit. Thnks again for your great work ;-D

waghanza:
Hi @kpicaza,

Thanks for your investigation.

I'll merge as-is. 

However, I can suggest you to find any formula for `max_concurrency`. Lemme explain myself a bit. 
Actually results are made on my `workstation` (depsite the hardware is close to what we use in **production**) and this _WILL_ vary (and then the number used actually could be irrelevant in the future is static).

Regards,

waghanza (approved):

PR #4656, opened 2021-09-16
chubbyjs-uwebsockets

dominikzogg:


waghanza (changes requested):
needs
```yml
build_deps:
  - git

bin_deps:
  - gcompat # required by uWebSocket
```
in `config` to use **uWebSocket**

dominikzogg:
@waghanza saw it, copied it from `fyrejet/fyrejet`

javascript/chubbyjs-uwebsockets/tsconfig.json
```diff
@@ -0,0 +1,10 @@
+{
+    "compilerOptions": {
+      "target": "es2019",
+      "module": "commonjs",
+      "outDir": "./",
+      "strict": true,
+      "declaration": true,
+    },
+    "exclude": ["node_modules"]
+  }
```
waghanza, line 10: I see that `nest` also use `tsconfig.json`, and seems to have differences (for example `nest` use **es2017**). I wonder if there will be differences on the final binary that may affects performances ....

dominikzogg: @waghanza code is build and tested against  es2019. there are features which aren't back portable without rewrites.

waghanza: my concern is not about how it is written, but about how it could affect what we want to compare

 is it OK in `javascript` world to compare es2017 to es2019 based code ?

dominikzogg: I don't think it's an issue, but if you do, you would need to upgrade the others, otherwise there is s big chance it's not working or could break at any time.

waghanza: honestly, I have not enough `javascript` skills to say so. let's just merge this for now

waghanza (approved):

javascript/chubbyjs-uwebsockets/tsconfig.json
dominikzogg: It could be a drawback for mine, cause the classic object model in JS (using function assigning prototypes instead of classes) is still that happens behind the scene.

PR #4685, opened 2021-09-28
[javascript] Update benchmarks for fyrejet

schamberg97:
* Disables marketing `X-Powered-By` header
* Uses slightly faster version of `send` function to better showcase the underlying performance
* Updates uWebSockets.js

### DIFF
--- a/javascript/fyrejet-uwebsockets/app.js
+++ b/javascript/fyrejet-uwebsockets/app.js
@@ -6,17 +6,18 @@ var app = fyrejet({
 });
 
 app.set("etag", false);
+app.set('x-powered-by', false)
 
 app.get("/", function (req, res) {
-  res.send("");
+  res.sendLite("");
 });
 
 app.get("/user/:id", function (req, res) {
-  res.send(req.params.id);
+  res.sendLite(req.params.id);
 });
 
 app.post("/user", function (req, res) {
-  res.send("");
+  res.sendLite("");
 });
 
 app.listen(3000, function () {});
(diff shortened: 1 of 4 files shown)


javascript/fyrejet-uwebsockets/app.js
```diff
@@ -6,17 +6,18 @@ var app = fyrejet({
 });
 
 app.set("etag", false);
+app.set('x-powered-by', false)
 
 app.get("/", function (req, res) {
-  res.send("");
+  res.sendLite("");
```
waghanza, line 12: is this intended to be used in **production** ?

I mean, what are the differences between `send` and `sendLite` ?

schamberg97: Yes. Fyrejet is heavily based on restana with res.sendLite being the renamed restana’s res.send. 

The goal of the framework, however, is to provide Express-like API with significantly improved performance. So it also implements most of Express API, besides a few deprecated features.

 The default res.send is thus a highly modified function of the same name from express.js, which, in comparison to res.sendLite also provides some features related to ETag, caching, etc. 

TL;DR: difference between res.sendLite and res.send is that the former has less features, but is much faster

schamberg97:
@waghanza meanwhile, are there any other frameworks that might need uWS to be updated?

waghanza (approved):

waghanza:
@schamberg97 Some other frameworks uses `uwebsocket` indeed, but the version is upto them
for an example in `chubbyjs` the version of `uwebsocket` is handle by the framework itself
https://github.com/chubbyjs/chubbyjs-uwebsockets-psr-http-message-bridge/blob/4796d8b789a2128862047b6bc19dcc14853dd0a0/package.json#L56

btw, @dominikzogg, how can you ensure to use an up-to-date `uwebsocket` release ? (I'm not even sure it's a problem for `chubbyjs`)

dominikzogg:
@waghanza ^1.2.1 upgrades to the same version

PR #4704, opened 2021-10-05
Upgrade baize to 0.11

abersheeran:


waghanza (changes requested):
Thanks 🎉 

But can you update also `config.yaml` ?

abersheeran:
https://github.com/the-benchmarker/web-frameworks/blob/master/python/baize-asgi/config.yaml#L3

😀It seems that the robot submitted the wrong version number last time, so I don't need to change it now.

waghanza (approved):
Thanks for double checking

PR #4785, opened 2021-10-30
[Crystal] Update athena to v0.15

waghanza:
Hi @Blacksmoke16,

This `PR` updates `athena` to **0.15**

Regards,


Blacksmoke16:
Thanks @waghanza, there is a breaking change that will need addressed tho. Tl;dr replace `ART` with `ATH` and `ARTA` with `ATHA` in https://github.com/the-benchmarker/web-frameworks/blob/master/crystal/athena/src/server.cr.

PR #4800, opened 2021-11-04
Update server.php

amirfaramarzi:


php/tower/config/server.php
```diff
@@ -4,7 +4,7 @@
     'listen' => 'http://0.0.0.0:3000',
     'transport' => 'tcp',
     'context' => [],
-    'count' => cpuCoreCount() * 2,
+    'count' => 8,
```
waghanza, line 7: why fix the CPU core num ? I mean that in the near futur,  we are going to upgrade infrastructure were benchmark runs (and then computer could have more than 8 core)

waghanza (changes requested):

php/tower/config/server.php
amirfaramarzi: hi @waghanza 

okay thanks
 
So please merge my new commit

PR #4815, opened 2021-11-08
[C#,F#] Update dotnet to 6.0

panesofglass:


panesofglass:
`Simplify.Web` appears to fail consistently during the `test` phase. I don't know why.

I also see random failures by one or more other projects, but these are inconsistent, and I cannot determine why they would fail. All build successfully and pass all tests on different runs.

panesofglass:
@i4004 any suggestions as to how to get Simplify.Web to respond with the expected empty string rather than HTML?

i4004:
@panesofglass 1. Ok, I'll check.
2. Do you have any example responses when it fails?

waghanza:
@i4004 seems to fail on each routes
https://github.com/the-benchmarker/web-frameworks/runs/4144664560?check_suite_focus=true

waghanza:
@64J0 @Krzysztof-Cieslak `saturn` is also failing

i4004:
So, I found the problem.

# Cause

Simplify.Web, by default, tries to set  `Thread.CurrentThread.CurrentCulture` to `en` and gets exception: `CultureNotFoundException - Culture is not supported. Parameter name: name en is an invalid culture identifier. `

That's because `aspnet:6.0-alpine` does not have package with cultures installed, so only `iv` (Invariant) culture is available.

If we try to set `<InvariantGlobalization>false</InvariantGlobalization>` parameter in csproj, then in runtime we got: `Process terminated. Couldn't find a valid ICU package installed on the system. Please install libicu using your package manager and try again. Alternatively you can set the configuration flag System.Globalization.Invariant to true if you want to run with no globalization support. Please see https://aka.ms/dotnet-missing-libicu for more information.`

# Solution

Install `libicu` in docker file in app run step

```docker
RUN apk add icu
```
It should solve the problem for all frameworks.

panesofglass:
> @64J0 @Krzysztof-Cieslak `saturn` is also failing

This is one that passes sometimes and fails others. If you look back through the tests, you'll see at least one seems to randomly fail to build.

panesofglass:
@i4004 seems to continue to fail even after the `Dockerfile` update.

panesofglass:
@i4004 what else might I be missing? Here's the link to the failed tests: https://github.com/the-benchmarker/web-frameworks/runs/4155752258?check_suite_focus=true

waghanza:
out of curiosity, is `alpine` relevant in `C#` / `F#` communities ?

I mean we should stick to what is used in **production**, and as I remember we have not challenged this choice (to use `alpine`)

waghanza:
all tests passes @panesofglass, but not sure about which platform (OS) we should use

panesofglass:
@waghanza both are valid. I use alpine in production, but I also use debian-based images.

panesofglass:
@davidfowl this work for you?

i4004:
> 
> 
> @i4004 what else might I be missing? Here's the link to the failed tests: https://github.com/the-benchmarker/web-frameworks/runs/4155752258?check_suite_focus=true

It is the same error, i will check again.

waghanza:
> It is the same error, i will check again.

this is then located on `alpine` @i4004, default image works

davidfowl:
cc @sebastienros

waghanza:
I'll merge this @panesofglass. Feel free to give any feedback @sebastienros, will be gladly appreciated

waghanza (approved):

PR #4844, opened 2021-11-16
Update results on 16th of November 2021

waghanza:
Failings taks :
+ [x] Micronaut

We have to check, probably this are temporary failures


whiplash:
For `micronaut` there are known issues using `gradle` 7.3. Suggestion for now seems to be to use 7.2
See https://github.com/micronaut-projects/micronaut-core/issues/6536

waghanza:
Thanks @whiplash. I'll pin `gradle` to **7.2** then, until this issue is solved

whiplash:
`struts` is failing now
```
curl: (7) Failed to connect to 172.17.0.2 port 3000: No route to host
```
not sure what goes on

waghanza:
false positive ....

actually all frameworks are tested in this kind of PR. since it is an update `PR` (no changes except result update), we need to check change, not test each implementations.

in order for me to be quicker, I choose this solution (and I check false positive by hand)

PR #4896, opened 2021-12-03
Update results

waghanza:
This `PR` updates results

#### Failing frameworks

+ `grape` https://github.com/ruby-grape/grape/issues/2205 ℹ️ use **master**

#### To run again

+ `bunrouter`


waghanza:
@mmoreram `driftphp` fails https://github.com/the-benchmarker/web-frameworks/runs/4411761336?check_suite_focus=true#step:7:16

waghanza:
some components (`symfony` based) are not on the same major
```php
  - Downloading symfony/http-foundation (v6.0.0)
  - Downloading symfony/event-dispatcher-contracts (v2.5.0)
  - Downloading symfony/event-dispatcher (v6.0.0)
  - Downloading symfony/var-dumper (v6.0.0)
  - Downloading symfony/error-handler (v6.0.0)
  - Downloading symfony/http-kernel (v5.4.0)
 ```
is it intended ?

waghanza:
@beetlex-io https://github.com/the-benchmarker/web-frameworks/runs/4443471017?check_suite_focus=true

@amirfaramarzi there is a **breaking change** for tower in https://github.com/the-benchmarker/web-frameworks/runs/4443483689?check_suite_focus=true#step:6:2194. I think it might be a good idea to `v1.17.0` instead of a `v1.16.2`

amirfaramarzi:
hi @waghanza 
This error is because your computer does not support php 8.1
Please upgrade your php version
And if possible, check that jit and opcache are enabled via the phpinfo function

And if you still need to change the version of the tower, send a message
Thanks

waghanza:
Hi @amirfaramarzi,

The point is that the php version required was changed (by a minor) is the same major of your framework, and then a major contains a **breaking change**.

With this change a user of `tower` (with a constraint like `^1.16.0` or `~1.16.0`) could break due to version change, its a common practice to upgrade major in such a case

amirfaramarzi:
@waghanza 

The tower version changed to 1.17.0

waghanza:
@panesofglass Any hand to solve https://github.com/the-benchmarker/web-frameworks/runs/4502361398?check_suite_focus=true#step:6:59 ?

waghanza:
@bung87 `scorper` is failing in https://github.com/the-benchmarker/web-frameworks/runs/4502852750?check_suite_focus=true

Willyboar:
Imo it will be better exclude failing frameworks and add a require fix label on the website so scheduled results don't effected.

waghanza:
Yeah, I plan to do this if I can not publish results / fix those frameworks for sunday

daviddenton:
The version of http4k is waaaaaaaay out of date so am not surprised it doesn’t work. 4.17.3.0 is the latest.

waghanza:
Thanks @daviddenton, having an outdated version couldn’t help, indeed (it should have been caught by `renovatebot` btw)

If you have time to create a new implementation, it will be very appreciated

daviddenton:
you should just be able to drop the new version in and it will work ok :)

PR #4928, opened 2021-12-15
Update httpbeast to 0.4

ajusa:
https://github.com/dom96/httpbeast

### DIFF
--- a/nim/httpbeast/config.yaml
+++ b/nim/httpbeast/config.yaml
@@ -1,3 +1,3 @@
 framework:
   github: dom96/httpbeast
-  version: 0.3
+  version: 0.4
--- a/nim/httpbeast/server.nimble
+++ b/nim/httpbeast/server.nimble
@@ -8,4 +8,4 @@ license = "MIT"
 # Dependencies
 
 requires "nim >= 0.19.0"
-requires "httpbeast >= 0.3 & < 0.4"
+requires "httpbeast >= 0.4 & < 0.5"


Willyboar:
Great! @waghanza do we need to update frameworks that using httpbeast?

waghanza:
sure it will be great @Willyboar

waghanza (changes requested):
missing `config.yaml`

Willyboar:
I think only whip needed do be done

waghanza (commented):
@ajusa https://github.com/ajusa/web-frameworks/pull/1

dom96:
Jester too, add `requires "httpbeast >= 0.4.0"` into https://github.com/the-benchmarker/web-frameworks/blob/master/nim/jester/server.nimble#L10

ajusa:
Thanks for the PR @waghanza, config.yml should exist now.
@dom96 Do I need to force httpbeast in this nimble file? I thought the way it worked is that Jester will use the version of httpbeast specified in its nimble file, so you'd need to cut a new release?

waghanza:
whip does not specify version, I assume `nimble` will take the last
https://github.com/mattaylor/whip/blob/a9fd1f04ba0167727dd0dbf9235791f708147992/whip.nimble#L12

I think indeed that `jester` need a release
https://github.com/dom96/jester/blob/a21b36a02b7745d6cdcda32d4ab3fba328cda17a/jester.nimble#L16

seems to be done in **master** tho

ajusa:
Yes, so `nimble` by default takes the latest release. I think that would mean that as long as this is ran on a blank slate, everything should use the latest version of httpbeast? Both whip and jester just ask for a minimum version of httpbeast, not a maximum, so they should be able to automatically use the newer one. 

I think this can be merged now, and a followup PR can happen if for some reason Jester and Whip don't find the updated httpbeast.

waghanza:
Could you track that @ajusa. Unfortunately I could monitor all frameworks

ajusa:
Sure. I think this is blocked from merging because of the change you requested, which you ended up making in a PR. Could you mark it as fixed?

waghanza (approved):

dom96:
Yeah, it should pick the latest. I don't know how these benchmarks are executed though, so I would suggest being explicit where possible (if httpbeast 0.2.2 happens to already have been installed then it will be used, this could happen if another benchmark installs that version, although I guess each benchmark gets its own clean slate so hopefully not a problem).

waghanza:
In this benchmark all things are built from scratch (I mean no cache, no lock files ...) so the latest should be picked.

That said, imho, the version should probably be explicitly written at `jester` level @dom96 (but not sure since the cost of maintenance and version sync could be very high).

PS : If anyone want to add `nimble` support in https://github.com/renovatebot/renovate it will be awesome. This is the only way to ensure version sync for few languages in there

dom96:
Indeed, ideally the versions should be locked. Nimble is getting lock file support so we should be able to do this soon.

waghanza:
that said we do not use lock files at all (for all languages), `renovatebot` could track version from manifest not only lock files

PR #4961, opened 2021-12-26
Upgrade/masonite 4

josephmancuso:


josephmancuso:
@waghanza we should probably delete whatever was on master and just merge this new project

waghanza:
@josephmancuso  When requesting `/`, I have
```python
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/site-packages/masonite/foundation/Application.py", line 49, in __call__
    return self.response_handler(*args, **kwargs)
  File "/usr/local/lib/python3.9/site-packages/masonite/foundation/response_handler.py", line 29, in response_handler
    application.make("exception_handler").handle(e)
  File "/usr/local/lib/python3.9/site-packages/masonite/exceptions/ExceptionHandler.py", line 39, in handle
    self.application.make("event").fire(
  File "/usr/local/lib/python3.9/site-packages/masonite/container/container.py", line 105, in make
    raise MissingContainerBindingNotFound(
masonite.exceptions.exceptions.MissingContainerBindingNotFound: event key was not found in the container
```

python/masonite/Kernel.py
```diff
@@ -0,0 +1,109 @@
...
+from masonite.auth import Sign
+from masonite.environment import LoadEnvironment
+from masonite.utils.structures import load
+from masonite.utils.location import base_path
+from masonite.middleware import (
+    SessionMiddleware,
+    EncryptCookies,
+    LoadUserMiddleware,
+    MaintenanceModeMiddleware,
+    VerifyCsrfToken,
```
waghanza, line 12: initially the version inside this implementation was used, why not using the one provided by `masonite`

waghanza (changes requested):

python/masonite/Kernel.py
josephmancuso: It is basically. This class is in new projects and inherits the Masonite class basically just for the purpose of being able to exclude routes from csrf

josephmancuso:
I don't get that issue when I run the server locally .. what are you doing?

The issue here is when a new exception is thrown it fires an event ... but I removed the event provider because there shouldn't be an exception thrown anyway. 

Even if I add it back it will return an exception but idk why it's throwing exceptions on that route ...

waghanza:
I run `gunicorn wsgi`, how can I run `masonite` otherwise ?

1. I run into the container with `bundle exec rake config && docker build -f .Dockerfile  -t test . && docker run -p 3000:3000 -ti test bash`
2. I start `gunicorn` with `gunicorn --bind 0.0.0.0:3000 wsgi`
3. I send an http request to `/` with `curl http://127.0.0.1:3000 -v` (since my container IP is bound to 127.0.0.1)

waghanza (approved):

PR #4965, opened 2021-12-28
[Rust] Add Axum

SabrinaJewson:
> `axum` is a web application framework that focuses on ergonomics and modularity.

cc @davidpdrsn @jplatte

### DIFF
--- a/rust/axum/Cargo.toml
+++ b/rust/axum/Cargo.toml
@@ -1,0 +1,15 @@
+[package]
+name = "server"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = "0.4.3"
+tokio = { version = "1.15.0", features = ["macros", "rt-multi-thread"] }
+
+[profile.release]
+codegen-units = 1
+opt-level = 3
+lto = true
+incremental = false
--- a/rust/axum/config.yaml
+++ b/rust/axum/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: tokio-rs/axum
+  version: 0.4
--- a/rust/axum/src/main.rs
+++ b/rust/axum/src/main.rs
@@ -1,0 +1,17 @@
+use axum::{extract::Path, routing, Router};
+
+#[tokio::main]
+async fn main() {
+    let router = Router::new()
+        .route("/", routing::get(|| async {}))
+        .route(
+            "/user/:id",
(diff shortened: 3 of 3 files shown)


rust/axum/Cargo.toml
```diff
@@ -0,0 +1,15 @@
+[package]
+name = "server"
+version = "0.1.0"
+edition = "2021"
+publish = false
+
+[dependencies]
+axum = "0.4.3"
```
waghanza, line 8: seems too strict for me, I think 
```suggestion
axum = "~0.4"
```

should be better, to make sure we are up-to-date but using 0.4.x (their is a bot taking care of updates here)

waghanza (changes requested):

waghanza:
Hi @SabrinaJewson,

Thanks for your contribution. 

I'm waiting for the explicit consent of any codeowner (of `axum`) before merging it.

I'll be glad if I can ping someone on each minor update

rust/axum/Cargo.toml
jplatte: Cargo treats `0.4.3` as `^0.4.3`, not `=0.4.3`. `~0.4` would additionally allow *older* versions than 0.4.3, newer ones would be chosen with either version bound (when there is no lockfile or when `cargo update` is run).

davidpdrsn:
I'm the creator of axum 😊 This looks good to me!

rust/axum/Cargo.toml
waghanza: I assume then this will keep into 0.4.x, that the he only thing to notice here.

I think better to stick to `rust` best practices, I let then you decide guys (I'm more a `ruby` dev and we do not have `^` constraints)

jplatte: Yeah, `^0.x.y` for some x > 0 means `>=0.x.y && <0.(x+1).0` in Rust & NodeJS land.

waghanza (approved):

waghanza:
@SabrinaJewson `axum` is in https://web-frameworks-benchmark.netlify.app/result

@davidpdrsn @jplatte May I notify one of you (or both) or minor updates ? It could help keeping a quality project

davidpdrsn:
> May I notify one of you (or both) or minor updates ? It could help keeping a quality project

Yep just ping me!

PR #4981, opened 2022-01-02
[Python] always patch meinheld

waghanza:


waghanza:
I've added a `version` in json data @SuspiciousLookingOwl

PR #4982, opened 2022-01-02
Update wsgi.py

josephmancuso:
I think this is missing from the previous integrations of Masonite. I am wondering if this is the culprit


waghanza (approved):
There is a change, but insignificant.

Still reach the 3K rps

PR #5005, opened 2022-01-10
Add Hexagon Toolkit (Kotlin) implementation

jaguililla:


kotlin/hexagon/build.gradle
```diff
@@ -0,0 +1,22 @@
+
+plugins {
+    id "org.jetbrains.kotlin.jvm" version "1.6.10"
```
waghanza, line 3:
```suggestion
    id "org.jetbrains.kotlin.jvm" version "1.6.+"
```

kotlin/hexagon/build.gradle
```diff
@@ -0,0 +1,22 @@
...
+}
+
+ext {
+    gradleScripts = "https://raw.githubusercontent.com/hexagonkt/hexagon/2.0.0-B3/gradle"
```
waghanza, line 7: what are the rationals behind this ?

kotlin/hexagon/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: hexagonkt.com
+  version: 2.0.0-B3
```
waghanza, line 3:
```suggestion
  version: 2.0
```

kotlin/hexagon/src/main/kotlin/Benchmark.kt
```diff
@@ -0,0 +1,30 @@
...
+import com.hexagonkt.http.model.ContentType
+import com.hexagonkt.http.server.HttpServerSettings
+import com.hexagonkt.http.server.jetty.JettyServletAdapter
+import com.hexagonkt.http.server.serve
+import java.net.InetAddress
+
+fun main() {
+    val textPlain = ContentType(TextMedia.PLAIN)
+    val adapter = JettyServletAdapter()
+    val settings = HttpServerSettings(InetAddress.getByName("0.0.0"), 3000)
```
waghanza, line 14:
```suggestion
    val settings = HttpServerSettings(InetAddress.getByName("0.0.0.0"), 3000)
```

waghanza (changes requested):
thanks for your contribution 🎉 

May you add a java implementation

jaguililla:
Hello and thanks for reviewing the PR so quickly :) Unfortunately, I cannot provide a Java implementation as the toolkit is designed to work only with Kotlin (using it in Java maybe harder than it seems). Can I submit only the Kotlin version?

kotlin/hexagon/build.gradle
jaguililla: Along with the toolkit, some Gradle scripts are also released to ease the projects configuration. I.e: https://raw.githubusercontent.com/hexagonkt/hexagon/2.0.0-B3/gradle/application.gradle would add utilities to create a single JAR with all the dependencies.

This variable is provided to be able to switch from one release to another one changing only the tag in the URL.

waghanza: I understand the idea (even I'm far away from `gradle` philosophy). will it not be clean (e.g -> more trust for developer) to use a `gradle` _plugin_ to define that ?

waghanza:
> Unfortunately, I cannot provide a Java implementation as the toolkit is designed to work only with Kotlin

Ah ok, I though `hexagon` was for both languages, I think I have confused with another framework

kotlin/hexagon/src/main/kotlin/Benchmark.kt
```diff
@@ -0,0 +1,30 @@
...
+    val adapter = JettyServletAdapter()
+    val settings = HttpServerSettings(InetAddress.getByName("0.0.0.0"), 3000)
+
+    serve(adapter, settings) {
+        get {
+            ok(contentType = textPlain)
+        }
+
+        path("/user") {
+            get {
```
waghanza, line 22: must be post no ?

waghanza (changes requested):

waghanza (changes requested):
final jar name must be `server`
```java
shadowJar {
    mergeServiceFiles()
    baseName = "server"
    classifier = null
    version = null
}
```

PS : feel free to adapt this part for other framework that use `gradle`

jaguililla:
> final jar name must be `server`
> 
> ```java
> shadowJar {
>     mergeServiceFiles()
>     baseName = "server"
>     classifier = null
>     version = null
> }
> ```
> 
> PS : feel free to adapt this part for other framework that use `gradle`

Done, Not using shadowJar, but I checked that `build/libs/server.jar` is generated and run properly with `java -jar build/libs/server.jar`

kotlin/hexagon/src/main/kotlin/Benchmark.kt
jaguililla: My bad, fixed in the latest update.

waghanza (approved):

PR #5025, opened 2022-01-13
scala: upgrade of base image

whiplash:
------
Closes #5009

### DIFF
--- a/scala/Dockerfile
+++ b/scala/Dockerfile
@@ -1,4 +1,4 @@
-FROM hseeberger/scala-sbt:11.0.12_1.5.5_2.13.6 AS build
+FROM hseeberger/scala-sbt:11.0.13_1.6.1_2.13.8 AS build
 
 WORKDIR /usr/src/app
 


waghanza:
You think 3.1 is not ready ?

waghanza (changes requested):
We do not need to change `scalaVersion` in sbt files ?

whiplash:
no no, this is `sbt` version. not `scala`

whiplash:
> You think 3.1 is not ready ?

Unfortunately Scala 3 migration is still an open point for many libraries/frameworks.

waghanza:
> > You think 3.1 is not ready ?
> 
> Unfortunately Scala 3 migration is still an open point for many libraries/frameworks.

Yeah, I saw, only `http4s` is ready as far as I understand

> We do not need to change scalaVersion in sbt files ?

This image does not update scala from 2.13.6 to 2.18.8 ?

whiplash:
You are right: 2.13.8 is the latest and I think all frameworks are already using it.

waghanza (approved):
scala version is already 👍🏻 
https://github.com/the-benchmarker/web-frameworks/blob/0c358cc58f39d76d09351dd56529e42ea6d6b427/scala/http4s/build.sbt#L4

PR #5066, opened 2022-01-26
Update to latest Hexagon release

jaguililla:


waghanza (commented):
could not this script be provided by package use in gradle build ?

jaguililla:
> could not this script be provided by package use in gradle build ?

I'm planning to move those scripts to a proper Gradle plugin, unfortunately, it is way down on the backlog (still a lot of functionality to code). Could it be approved that way until build scripts are moved to Gradle plugins? Thanks in advance!

waghanza (approved):

PR #5104, opened 2022-02-11
[Csharp] add AspNet MVC framework

dj-nitehawk:


csharp/aspnet-mvc/web.csproj
```diff
@@ -0,0 +1,9 @@
+<Project Sdk="Microsoft.NET.Sdk.Web">
+
+    <PropertyGroup>
+        <TargetFramework>net6.0</TargetFramework>
+        <Nullable>enable</Nullable>
```
waghanza, line 5: is it specific option for raw asp net ? I mean should it be replicated to other C# / F# frameworks ?

The only concern I have is about this being a tweak, and then having a messy comparison of C# / F# frameworks

dj-nitehawk: not sure if you're aware but asp.net 6 (foremaly known as `asp.net core`) has multiple options/patterns for doing rest apis. what you have in the `aspnetcore` folder is a pattern called `endpoints` or `minimal apis`. what i submitted here is the `mvc` method, which is the most widely used. so people might want to know how the mvc vs. minimal/endpoints are performing.

so i wouldn't consider this PR a tweak. it should be considered a framework on it's own which should be pitted against others. even techempower has separate benchmarks for mvc.

screen

waghanza: no I think all stuff should be included. I just meant we have to get some structure before.

for the moment we can merge as long as it will be approved, but in the future we need to add this approach level thing. I mean, the fwk is this asp.net, but as you say mvc, fastendpoint is different approches

dj-nitehawk: yes should look into supporting multiple endpoint frameworks for any given base framework. afaik in .net world there are at least 10 I can name. 

also pls look into a test scenario that will work the json serializer of all frameworks. at least at a basic level.

keep up the good work 👍

waghanza (approved):

dj-nitehawk:
yep. mvc is much heavier than minimal api/endpoints. have a look at my own [benchmark results](https://fast-endpoints.com/wiki/Benchmarks.html). same with [techempower benchmarks](https://www.techempower.com/benchmarks/#section=test&runid=fa199ea7-b3db-4cd5-bab8-85ff67217db0&hw=ph&test=json&l=zik0zh-sf&c=8&a=2). do note that minimal/endpoint approach is still missing from techempower.

dj-nitehawk:
also it might be logical to rename `aspnetcore` to `aspnet-minimalapi` to give a clear indication for people without requiring them to look at the source code. also after .net 5, the `core` name is gone.

waghanza:
Seems fair @dj-nitehawk. Feel free to open a PR for this

PR #5126, opened 2022-02-16
Kotlin Hexagon: use new server adapter (Netty)

jaguililla:


waghanza (commented):
Maybe worth to create `hexagon-netty` and `hexagon-jetty` no ?

Those implementation could be collapsed with https://github.com/the-benchmarker/web-frameworks/pull/3958, but their is no ready UI actually

waghanza:
Could you use `-jetty` and `-netty`. Will be easier for me ...

PS : CI will fail in this scenario

waghanza (approved):

PR #5235, opened 2022-03-24
fix(js-nanoexpress): make sure load correct path

dalisoft:


waghanza:
@dalisoft `Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: Package subpath './cjs' is not defined by "exports" in /usr/src/app/node_modules/nanoexpress/package.json`

dalisoft:
@waghanza Oops, it worked locally yesterday. Today or tomorrow i will try fix that

PR #5286, opened 2022-04-23
Update restana doc

waghanza:
This `PR` update [restana](https://github.com/BackendStack21/restana) website

@jkyberneees is it accurate to point on https://restana.21no.de instead of github ?


jkyberneees:
Hi @waghanza, thanks for asking. Yes, that is correct!

PR #2641, opened 2020-05-04
Rust.warp

whiplash:
Closes #2637 #751

PR #5307, opened 2022-05-02
Added New Framework WARP

krishnaTORQUE:


waghanza:
Hi @krishnaTORQUE,

We discuss this on https://github.com/the-benchmarker/web-frameworks/pull/2641 with @seanmonstar.

Personally, I'm not against including `warp` in `here`. That said, I prefer to include in here maintained implementations, and I'm not willing to force anyone to do so.

If you want to maintain this implementation and if @seanmonstar agrees with this (I mean he is ok to include this with this condition), I'm glad to include it here.

Regards,

krishnaTORQUE:
Hi
I can maintain the version issue.
@seanmonstar with your approval. 
Really like the framework & other projects too.

waghanza:
I'll merge this week end if their is no blocker

rust/warp/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: github.com/seanmonstar/warp
+  version: 0.3.2
```
waghanza, line 3:
```suggestion
  version: 0.3
```

rust/warp/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: github.com/seanmonstar/warp
```
waghanza, line 2:
```suggestion
  github: seanmonstar/warp
```

waghanza (changes requested):
last changes before merge

krishnaTORQUE:
I updated the PR. Please review it.

rust/warp/src/main.rs
```diff
@@ -0,0 +1,15 @@
...
+async fn main() {
+    let index = warp::path::end().and(warp::get()).map(|| "");
+
+    let user = warp::path!("user" / u32).and(warp::get()).map(|id: u32| id.to_string());
+
+    let user_post = warp::path("user").and(warp::post()).map(|| "");
+
+    let routes = index.or(user).or(user_post);
+
+    warp::serve(routes).run(([0, 0, 0, 0], 3000)).await;
```
seanmonstar, line 14: The `unstable_pipeline` option is specifically exposed for this benchmark: https://docs.rs/warp/latest/src/warp/server.rs.html#390

waghanza: Is it safe to use this in production ? I mean the idea (of this projet) is to keep as-close-as we can to **production**-ready app

seanmonstar: As long as there is a pipeline benchmark, the option should be enabled.

krishnaTORQUE: Hi @seanmonstar 
Sorry I did not understand exactly where I need to change, it changes require, can you please mark it with the correct one?

seanmonstar, line 14:
```suggestion
    warp::serve(routes)
        .unstable_pipeline()
        .run(([0, 0, 0, 0], 3000))
        .await;
```

seanmonstar: The option should be enabled. All it does is tell hyper's read/write loop to change its flushing behavior from the default (which is to flush after every response is ready) to flush once all pending responses are ready. This behavior is the same that is used in many of the benchmarks in this repo.

waghanza: Understood @seanmonstar. However, I'm in favor of disabling this for all framework (I was not aware of that so thanks  🙏 for it). But maybe this can be done it another PR

seanmonstar: Do you mean that request pipelining should be removed from the plain text benchmarks in general? Or just from some frameworks?

waghanza: Misunderstood, this is all about http pipelining. HTTP pipeling is OK imho, and could be enabled in **production**

But I am not sure how to use the following

> It's "unstable" because the overwhelming majority of production servers don't use pipelined requests

seanmonstar: HTTP/1 pipelining in general is deprecated, all browsers have disabled it. It's not a typical thing used in production servers, in any language. But I imagine removing it from these benchmarks entirely would be contentious.

If they stay, there is no reason to disable the option in hyper or warp or any other framework depending on hyper. The feature works the same as any other implementation does. It would be unfair to remove it selectively, even if I personally encourage people to not think about HTTP pipelining, since most people just look at the benchmark numbers.

waghanza: > It would be unfair to remove it selectively

Aligned

> to not think about HTTP pipelining

Aligned

In my opinion, we should remove it. The global idea of this benchmark is not to have the highest figures as you can, but an idea of what you **COULD** have in production

krishnaTORQUE:
The result is freaking awesome. :D

PR #5466, opened 2022-07-09
drop-chubbyjs

dominikzogg:


dominikzogg:
Hello @waghanza 

Not sure if this is fixable by me.
The reason for the drop its, that the chubbyts should get the successor of chubbyjs.

PR #5479, opened 2022-07-14
feat: added support for go-stgin

AminMal:


go/stgin/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: github.com/AminMal/stgin
+  version: 1.1.0
```
waghanza, line 3:
```suggestion
  version: 1.1
```

go/stgin/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: github.com/AminMal/stgin
```
waghanza, line 2:
```suggestion
  github: AminMal/stgin
```

go/stgin/go.mod
```diff
@@ -0,0 +1,7 @@
+module main
+
+go 1.18
```
waghanza, line 3:
```suggestion
```

go/stgin/go.mod
```diff
@@ -0,0 +1,7 @@
...
+
+require github.com/AminMal/stgin v1.1.0
+
+require github.com/AminMal/slogger v1.0.0 // indirect
```
waghanza, line 7: why both ? is there no a require only module that require both ?

waghanza (changes requested):

go/stgin/go.mod
AminMal: Unfortunately not for now. That's the logging lib. But I'll make sure to take a look on how to do it in future stgin releases.

waghanza (approved):

PR #5537, opened 2022-08-09
update spliffy to 1.0.0

snowbldr:


waghanza (changes requested):
Could you remove `package.lock` file ?

PS : No need to take care about update (unless their is a API change), `renovatebot` make it for us

PR #5582, opened 2022-08-30
Update

waghanza:
------
### Removed (from results) framework

------

- vibed
- jennet
- one
- slim-swoole
- chubbyphp-swoole
- basolato
- nawa

ℹ️  For now, I'll remove the above frameworks from results. Need to find a better solution 

The ideal solution for `one` and `chubbyphp` will be to use `swoole` 4.8, but the toolset is not made for that now

cc @dominikzogg @lizhichao


sy-records (dismissed):
LGTM. One and chubbyphp need to support swoole v5.

waghanza:
OK. I'll pin `swoole` for `one` and `chubby`

PR #5583, opened 2022-08-30
[PHP] Update swoole to v5.0

waghanza:


dominikzogg:
@waghanza https://github.com/chubbyphp/chubbyphp-swoole-request-handler 1.3 is not using swoole 5

PR #5601, opened 2022-09-05
Add Athena Routing

Blacksmoke16:
This adds a new framework that is using https://athenaframework.org/Routing/ by itself, which provides a way to have a super simple framework that essentially just adds routing to the stdlib's `HTTP::Server`.


crystal/athena-routing/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  github: athena-framework/routing
+  version: 0.1.3
```
waghanza, line 3:
```suggestion
  version: 0.1
```

waghanza (changes requested):

waghanza (approved):

PR #5640, opened 2022-09-15
update express app  and add cluster

alirezanqp:


waghanza (commented):
Is this really necessary ? I mean all nodes implementation uses `pm2`

PR #5641, opened 2022-09-15
update fastify and add cluster

alirezanqp:


waghanza (commented):
Is this really necessary ? I mean all nodes implementation uses `pm2`

alirezanqp:
Oh, if you use pm2, no, it's not necessary

PR #5644, opened 2022-09-15
[Java] Upgrade to v17

### LINKED ISSUE #5642: [Java] upgrade to JDK 17
Could we try and upgrade JDK to the 17? It's still LTS and shouldn't be breaking. I was about to add https://helidon.io which requires Java 17.

waghanza:
Not sure all frameworks are ready @whiplash

Let see ...

----

Close #5642

### DIFF
--- a/java/Dockerfile
+++ b/java/Dockerfile
@@ -3,7 +3,7 @@
 {{/image}}
 
 {{^image}}
-  FROM maven:3-jdk-11-slim AS build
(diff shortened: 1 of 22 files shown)


waghanza:
Worked like a charm @whiplash except for `act`

cc @greenlaw110

whiplash (approved):

whiplash (approved):

PR #5704, opened 2022-10-04
Version Updated

krishnaTORQUE:


rust/warp/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "server"
-version = "0.1.0"
+version = "0.0.0"
```
waghanza, line 3: not sure to understand this change

waghanza (changes requested):
missing `config.yaml`

rust/warp/Cargo.toml
krishnaTORQUE: because this does not make sense for this web application, so making `0.0.0` better I think.

waghanza: agree, this information is not relevant. could not simply remove it ?

krishnaTORQUE: I removed but build failed. So it looks like rust require version to be exists in order to build/compile.

waghanza (approved):

PR #5719, opened 2022-10-08
update results

waghanza:
###### Failing frameworks

+ Vide.D
+ ArcHTTP
+ Jennet
+ Cyclone
+ Ambiorix
+ Basolato
+ Nawa

###### Hidden frameworks

+ Rack App
+ Hanami
+ Camping

@sy-records `pecl install swoole` setup swoole 4.x instead of 5.x (last major). is this intended ? or a bug in `alpine`'s `pecl` **package** ?


jodosha:
@waghanza Why Hanami and Hanami-API are hidden?

waghanza:
my mistake for `hanami-api`, I'll re-update result, as soon as `hanami` is fixed (and add `hanami-api`)

PR #5793, opened 2022-11-02
Update Grip from 1.0.0 to 2.0.0

grkek:


crystal/grip/shard.yml
```diff
@@ -11,6 +11,6 @@ targets:
 dependencies:
   grip:
     github: grip-framework/grip
-    version: ~> 1.0.0
+    version: 2.0.0
```
waghanza, line 14: in order to use last patch version, should we not do ?

```suggestion
    version: ~> 2.0.0
```

waghanza (changes requested):

waghanza (approved):

PR #5862, opened 2022-11-16
Added tide framework

krishnaTORQUE:


waghanza:
When running inside the container, I have
```
/usr/src/app/target/release/server: /lib/x86_64-linux-gnu/libm.so.6: version `GLIBC_2.29' not found (required by /usr/src/app/target/release/server)
```

@krishnaTORQUE 

cc @msrd0

krishnaTORQUE:
@waghanza 
I think you need `build-base` package.
If you are using alpine then here is `apk add build-base`

waghanza:
you mean https://packages.debian.org/fr/buster/build-essential, I think

krishnaTORQUE:
> you mean https://packages.debian.org/fr/buster/build-essential, I think

Yes maybe for debian. What I just see in dockerfile.
But it suppose to be there on build time & not runtime.
2nd is rust image should have this as well. (Not sure)

waghanza:
seems that rust base image does not have those libs installed. the dockerfile is
```dockerfile
FROM rust:1.65-slim AS build
WORKDIR /usr/src/app
COPY . ./
RUN cargo build --release && strip target/release/server
FROM debian:buster-slim
COPY --from=build /usr/src/app/target/release/server /usr/src/app/target/release/server
CMD /usr/src/app/target/release/server
```

krishnaTORQUE:
@waghanza 
Try this & let me know
```Dockerfile
FROM rust:1.65-slim AS build
WORKDIR /usr/src/app
COPY . ./
RUN apt install libc6-dev build-essential # <-- Here Try it & let me know
RUN cargo build --release && strip target/release/server
FROM debian:buster-slim
COPY --from=build /usr/src/app/target/release/server /usr/src/app/target/release/server
CMD /usr/src/app/target/release/server
```

waghanza:
same with
```dockerfile
FROM rust:1.65-slim AS build

WORKDIR /usr/src/app

RUN apt-get -qy update

  RUN apt-get -qy install libc6-dev
  RUN apt-get -qy install build-essential

COPY . ./

RUN cargo build --release && strip target/release/server

FROM debian:buster-slim

COPY --from=build /usr/src/app/target/release/server /usr/src/app/target/release/server

CMD /usr/src/app/target/release/server
```

msrd0:
This is probably due to the debian version mismatch. You're building with the default debian-based rust docker image which should be based on debian bullseye but run it in debian buster. Try `FROM debian:slim` for the 2nd stage to use the latest debian version, or pin the debian version in the first stage.

waghanza:
@jbr @Fishrock123  @yoshuawuyts @skade do you consent to include `tide` here ? beside, if any one of you allow me to notify him/her on each (minor) update, it will be 👍🏻

jbr:
In the past tide's preference was to not participate in benchmarks when there was an option, which is part of why the TechEmpower framework benchmark entry was written by a third party and is unofficial. Tide is [currently not actively being developed](https://github.com/http-rs/tide/discussions/888), but I'm not sure if that is reason to exclude it from your benchmark.

waghanza:
As I understand above thread @jbr, `tide` is :
- still maintained (at least not because of feature addition, but because somehow some folks could patch it in case)
- used in **production**

for this reason, I think it could be included here (I'll be glad if some folks help me on maintaining the implementation here). also it could be a great way to detect failures (compilation issue with the last rust version ....)

but if any main contributor does not want it to be included here, I'll respect this choice ❤️

krishnaTORQUE:
> As I understand above thread @jbr, `tide` is :
> 
>     * still maintained (at least not because of feature addition, but because somehow some folks could patch it in case)
> 
>     * used in **production**
> 
> 
> for this reason, I think it could be included here (I'll be glad if some folks help me on maintaining the implementation here). also it could be a great way to detect failures (compilation issue with the last rust version ....)
> 
> but if any main contributor does not want it to be included here, I'll respect this choice heart

I can maintain tide as I am also maintaining the warp. I have no issue.
But yes, if author does not want it, then we should respect their choice.

krishnaTORQUE:
@waghanza 
I am going to close this PR for now. & Raise another PR after I test & fix the issue with docker in my system.
Is it fine?

waghanza:
I prefer to keep it open for a while (some days probably), just to make sure we gather consent (or non-opposition) of author(s)

krishnaTORQUE:
> I prefer to keep it open for a while (some days probably), just to make sure we gather consent (or non-opposition) of author(s)

Yes sure,
That make sense

msrd0:
Regardless of tide getting added to the benchmark or not, you should keep the changes you made to the Dockerfile.

waghanza:
@msrd0 changes are already in **master**. after digging into docker hub, we need to use `debian:stable-slim` instead of `debian:slim`, and better to have this for all `rust` **frameworks**

waghanza (approved):

PR #5902, opened 2022-11-25
[Nim] Use whip 0.2

### LINKED ISSUE #4930: [Nim] Use jester 0.2

waghanza:
cc @mattaylor 

Closes #4930


waghanza:
Not sure why but it does not use `httpbeast` **0.4.1** (the latest version)

```nim
Error: unhandled exception: Address family not supported by protocol [OSError]
```

PR #5910, opened 2022-11-28
Upgrade RESTHeart to v7

ujibang:
This PR upgrades RESTHeart to v7


java/restheart/pom.xml
```diff
@@ -9,7 +9,7 @@
   <properties>
     <maven.compiler.source>17</maven.compiler.source>
     <maven.compiler.target>17</maven.compiler.target>
-    <restheart.version>[6.6,6.7)</restheart.version>
+    <restheart.version>[7.0,7.100)</restheart.version>
```
waghanza, line 12: just curious : what is this kind of range ?

ujibang: A range that will select all `7.x` versions and will not select `8.0.0-SNAPSHOT`

I found that a snapshot release like `8.0.0-SNAPSHOT` is taken by [7.0,80)

waghanza: I mean that `[7.0,7.1)` feels more natural (at least for me). What does `.100 means` ?

waghanza: Yes, please. I prefer to stick to minor. Thus in `java` world some package are release with `beta` status **and** in this range, we need to make sure that results published uses **only** production-ready frameworks (so no `beta`)

waghanza (approved):

PR #5939, opened 2022-12-08
Update

waghanza:
###### jennet

As `ice` I could not find `libexecinfo` on the last alpine.

Any clue @Theodus ?

###### ambiorix

I can not start `ambiorix`, I have

```r
Error in library(ambiorix) : there is no package called ‘ambiorix’
Calls: source -> withVisible -> eval -> eval -> library
Execution halted
```

cc @JohnCoene

###### comet

I can start `comet`, I have

```php
Fatal error: Type of Comet\Factory\CometPsr17Factory::$responseFactoryClass must be string (as in class Slim\Factory\Psr17\Psr17Factory) in /usr/src/app/vendor/gotzmann/comet/src/Factory/CometPsr17Factory.php on line 6
```

cc @gotzmann 

###### ice

I can not setup `ice` on **alpine**
```
 apk info --who-owns execinfo.h
ERROR: execinfo.h: Could not find owner package
```

How can I do so @mruz ?

###### nawa

Not sure why but at some point the server does not respond

cc @jatofg

PR #5942, opened 2022-12-09
fix: ambiorix install from remote

JohnCoene:
I believe this should fix ambiorix #5939


waghanza:
Which `r` instruction do you need to use to install from `DESCRIPTION` ? @JohnCoene

waghanza:
Manage to do it ... but not very satisfied with this kind of solution

waghanza (approved):

PR #5981, opened 2022-12-20
[PHP] Added Lemon Microframework

TENMAJKL:
This PR adds Lemon Microframework for PHP.

### DIFF
--- a/php/lemon/composer.json
+++ b/php/lemon/composer.json
@@ -1,0 +1,5 @@
+{
+    "require": {
+        "lemon_framework/lemon": "^3.10"
+    }
+}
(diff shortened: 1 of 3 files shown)


php/lemon/composer.json
```diff
@@ -0,0 +1,5 @@
+{
+    "require": {
+        "lemon_framework/lemon": "^3.10"
```
waghanza, line 3:
```suggestion
        "lemon_framework/lemon": "~3.10.0"
```

php/lemon/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: Lemon-Framework/Lemon
+  version: 3.10.1
```
waghanza, line 3: we need to force as string, cause `3.10` will show `3.1` in results

```suggestion
  version: '3.10'
```

waghanza (approved):

PR #6039, opened 2023-01-04
[FSharp][Giraffe / Giraffe-endpoints] Update giraffe to 6.0

SIRHAMY:
Giraffe has been on its 6.0 release since mid 2022, so updating it here as well.

Giraffe repo: https://github.com/giraffe-fsharp/Giraffe 

# Testing

Ran both locally with `dotnet run`


waghanza (approved):
Thanks 🎉

PR #6053, opened 2023-01-09
[Javascript] Upgrade uwebsocket

waghanza:


waghanza:
@jkyberneees @schamberg97 perhaps a good idea to track `uWebSocket` update through https://github.com/marketplace/renovate

PR #5929, opened 2022-12-06
Fix fsharp giraffe benchmarks

SIRHAMY:
This commit aims to rectify oddities in the fsharp giraffe benchmarks (giraffe, giraffe-endpoints) that likely led to its unusually poor results.

The fix to each is to clear the default logging providers - presumably this removes a good amount of processing / IO which leads to better request / response performance.

## Why is this okay? Is this cheating?

I believe this is okay because this seems to be the standard for dotnet benchmarks (fsharp and csharp) - both in this repo and in others (like tech empower). 

It is a little sus (idk if other languages are doing this) but it will at least get these benchmarks in parity with others on the same tech stack.

fsharp:

- Falco - clears - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/falco/Program.fs#L11
- Saturn - clears - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/saturn/Program.fs#L19
- frank - clears - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/frank/Program.fs#L31
- suave - not sure - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/suave/Program.fs
- websharper - clears - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/websharper/Main.fs#L44

csharp:

aspnet-minimal-api - clears - https://github.com/the-benchmarker/web-frameworks/blob/master/fsharp/websharper/Main.fs#L44 

tech empower

- giraffe in tech empower also clears - https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/FSharp/giraffe/src/App/Program.fs#L140

References
- @panesofglass benchmark perf questions - https://github.com/giraffe-fsharp/Giraffe/issues/529

PR #6089, opened 2023-01-25
[fsharp][suave] Remove logger IO in Suave to improve performance

SIRHAMY:
# Overview

I'm removing the default logger from Suave by using its built-in [Logger interface](https://github.com/SuaveIO/suave/blob/57d69e39e67a5b1a1c60b7c015c698a1901a9420/src/Suave/Utils/Logging.fs#L178) (see [docs](https://github.com/SuaveIO/suave/blob/57d69e39e67a5b1a1c60b7c015c698a1901a9420/docs/api.md))

I expect this to improve Suave's performance considerably and make this a better comparison against other fsharp frameworks.

# Context

* Problem: [I noticed that Suave was performing terribly](https://hamy.xyz/labs/2023-01-top-fsharp-backend-frameworks) - a negative outlier 
* It seems that Suave has default logging enabled when using the defaultConfig
* Solution: Remove logging

Note: I made [a similar PR to the Giraffe benchmark](https://github.com/the-benchmarker/web-frameworks/pull/5929) a few months ago and it seemed to work resulting in [6.5x improvement in benchmark](https://hamy.xyz/labs/2022-12-improving-fsharp-giraffe-web-benchmarks).

# Testing

* Ran Suave locally with dotnet run -> confirmed server runs as expected

### DIFF
--- a/fsharp/suave/Program.fs
+++ b/fsharp/suave/Program.fs
@@ -1,5 +1,6 @@
 open Suave
 open Suave.Filters
+open Suave.Logging
 open Suave.Operators
 open System.Net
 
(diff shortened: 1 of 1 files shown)


waghanza:
Thanks @SIRHAMY. As you may know I'm probably not the best guy to discuss about `F#`, but it this is considered as a `best-practices` let's go (I mean if you safely can run this in production)

SIRHAMY:
It should be safe to run a server this way and it runs / behaves as expected with `dotnet run`.

The caveat is I wouldn't really suggest anyone actually run an app in production with logging turned off. But since benchmarks tend to try and limit uncontrolled variables (in this case the implementations of default logging / IO) I think it makes sense for benchmarks.

Note that all the other fsharp benchmarks also have logging IO turned off so this should make this benchmark more consistent with the other ones.

waghanza (approved):

PR #6099, opened 2023-01-30
Latest version of unic framework

rajkumardusad:
Added latest version of unic framework and used swoole server.

### DIFF
--- a/php/unic/config.yaml
+++ b/php/unic/config.yaml
@@ -5,4 +5,4 @@ framework:
 php_ext:
   - swoole
 
-command: php index.php
+command: php public/index.php
--- a/php/unic/config.yaml
+++ b/php/unic/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: unicframework.github.io/docs
-  version: 1.0
+  version: 1.0.1
 
 php_ext:
   - swoole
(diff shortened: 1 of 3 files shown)


php/unic/config.yaml
```diff
@@ -1,3 +1,8 @@
 framework:
   website: unicframework.github.io/docs
-  version: 1.0
+  version: 1.0.1
```
waghanza, line 3:
```suggestion
  version: 1.0
```

waghanza (changes requested):

php/unic/public/index.php
```diff
@@ -2,8 +2,10 @@
 require(__DIR__.'/../vendor/autoload.php');
 
 use Unic\App;
+use Swoole\Http\Server;
 
 $app = new App();
+$server = new Server('0.0.0.0', 3000);
```
waghanza, line 8: You let swoole use it's default values. Maybe you should dig into https://github.com/the-benchmarker/web-frameworks/blob/771608e476428245d58d0d9ef1c36583450c801f/php/swoole/start.php#L8

The most important is to disable log output

waghanza (approved):

PR #6151, opened 2023-02-18
Add JS Framework : Rexuws

Aderinom:
Hello Community! 

I found a express adapter for uWebsockets.js, Rexuws.
https://github.com/rexuws/rexuws

I want to see how well this does in the benchmarks,to see if I should start putting some work into it. 

With this PR I want to add it to the JS benchmarks.

### DIFF
--- a/javascript/rexuws/app.js
+++ b/javascript/rexuws/app.js
@@ -27,6 +27,4 @@ app.post("/user", function (req, res) {
   res.send("");
 });
 
-app.listen(3000, function () {
-  console.log("listening on port 3000");
-});
+app.listen(3000);
--- a/javascript/rexuws/app.js
+++ b/javascript/rexuws/app.js
@@ -1,0 +1,32 @@
+/**
+ * 
+ * Rexuws (Replace Express with uWebsocket) is a framework based on uWebsocket.js
+ * It aims to be a drop-in, high speed replacement for express.
+ * 
+ * With uWebsockets.js clustering only properly works on Linux (not on WSL!).
+ * (presumably dependend the OS's of SO_REUSEPORT support)
+ * 
+ */
+
+
+const rex = require('rexuws').default;
+
+const app = rex({
+  useDefaultParser:true,
+})
+
+app.get("/", (req, res) => {
+  res.send("");
+})
+
+app.get("/user/:id", function (req, res) {
+  res.send(req.params.id);
+});
+
+app.post("/user", function (req, res) {
+  res.send("");
+});
+
+app.listen(3000, function () {
+  console.log("listening on port 3000");
+});
--- a/javascript/rexuws/config.yaml
+++ b/javascript/rexuws/config.yaml
@@ -1,0 +1,14 @@
+framework:
+  website: https://github.com/rexuws/rexuws
+  github: rexuws/rexuws
+  version: 1.1.4
+
+build_deps:
+  - git 
+  - python3
+  - make
+  - g++
+  - libmagic  # required by uWebSockets.js
+
+bin_deps:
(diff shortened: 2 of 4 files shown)


javascript/rexuws/config.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+  website: https://github.com/rexuws/rexuws
```
waghanza, line 2: not useful (website will target github due to next config)

javascript/rexuws/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  github: rexuws/rexuws
+  version: 1.1.4
```
waghanza, line 4:
```suggestion
  version: 1.1
```

javascript/rexuws/package.json
```diff
@@ -0,0 +1,5 @@
+{
+    "dependencies": {
+        "rexuws": "https://github.com/Aderinom/rexuws-update-deps"
```
waghanza, line 3: a fork ?

waghanza (changes requested):

waghanza:
I only can accept this `PR` after https://github.com/rexuws/rexuws/pull/27 is released @Aderinom 

However, I appreciate your participation, and feel free to contribute for anything else (there is some issue listing javascript frameworks, like `Adonis`)

javascript/rexuws/package.json
Aderinom: Yes. Updated the version of uws to allow newer node version to run it. 
Didn't really think about the ergononics at that point though.

Already read your comment. Makes sense to block it until then.

Aderinom:
Pushed a possible workaround for the incompability of the older rexuws version.
Added the newer UWS version as a override. Allows the server to run under Node 18.14.

If that's too hacky I understand. 

Also changed the config.yaml as you requested.

javascript/rexuws/app.js
```diff
@@ -0,0 +1,32 @@
...
+app.get("/user/:id", function (req, res) {
+  res.send(req.params.id);
+});
+
+app.post("/user", function (req, res) {
+  res.send("");
+});
+
+app.listen(3000, function () {
+  console.log("listening on port 3000");
```
waghanza, line 31: I think this could mess-up the results a little bit

javascript/rexuws/package.json
```diff
@@ -0,0 +1,14 @@
+{
+    "scripts": {
+        "run": "node app.js"
+    },
+    "dependencies": {
+        "rexuws": "^1.1.4",
```
waghanza, line 6:
```suggestion
        "rexuws": "~1.1.4",
```

the idea is to have the same operation type in all package.json for javascript (not sure which is more standard)

waghanza (changes requested):
let's wait a little bit to merge this (maybe in the next round), just to make sure the author accept https://github.com/rexuws/rexuws/pull/27

javascript/rexuws/package.json
Aderinom: Sure, makes sense to limit it to only minor version changes.

javascript/rexuws/app.js
Aderinom: This only runs once after the socket has successfully been bound, not on every request.

So this should not cause any difference in req/s.

Aderinom: Buuut... since it's not required I guess I'll remove it

waghanza:
I'll close this PR, and rise it again when `rexuws` 2.x will be out

Aderinom:
Yes, sorry. I completely forgot about this!

PR #6205, opened 2023-03-09
[Javascript] Update uWS to 20.20

waghanza:
This `PR` upgrade `uwebsockets` in here

The following frameworks could be affected :

+ routejs-uwebsocket
+ sifrr
+ hyper-express
+ fyrejet-uwebsockets
+ spliffy
+ low-http-server
+ naturaljs-router
+ chubbyts-uwebsockets

Maybe it could be valuable to add some bots update on your side @dominikzogg ?

Regards,


dalisoft (approved):
As soon as all apps still running fine, updating recommended by me for dependency uWebSockets.js
LGTM

dominikzogg:
@waghanza i do not  like those bots, cause i want to support a wide range of dependencies, but i've updated the related library

waghanza:
I understand @dominikzogg, but how do you manage when uws will drop support of some version of node used by some folks using `chubbyts` ?

I admit those bots could create some noise however ...

jkyberneees (approved):
LGTM

PR #6233, opened 2023-03-21
[PHP] Added kuiper framework

wenbinye:
Kuiper is a simple php framework, document by https://kuiper-framework.github.io/en/

### DIFF
--- a/php/kuiper/composer.json
+++ b/php/kuiper/composer.json
@@ -1,0 +1,26 @@
+{
+  "require": {
+    "php": "^8.1",
+    "ext-json": "*",
+    "kuiper/web": "^0.8"
+  },
+  "require-dev": {
+    "kuiper/component-installer": "^0.1"
+  },
+  "autoload": {
+    "psr-4": {
+      "app\\": "src/"
+    }
+  },
+  "scripts": {
+    "container-config": "kuiper\\component\\ComponentInstaller::generate"
+  },
+  "extra": {
+    "kuiper": {
+      "config-file": "src/container.php",
+      "whitelist": [
+        "kuiper/*"
+      ]
+    }
+  }
+}
--- a/php/kuiper/config.yaml
+++ b/php/kuiper/config.yaml
@@ -1,0 +1,8 @@
+framework:
+  website: kuiper-framework.github.io
+  version: 0.8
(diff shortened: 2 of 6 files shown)


php/kuiper/composer.json
```diff
@@ -0,0 +1,26 @@
+{
+  "require": {
+    "php": "^8.1",
+    "ext-json": "*",
+    "kuiper/web": "^0.8"
```
waghanza, line 5: to align constraint on php based frameworks

```suggestion
    "kuiper/web": "~0.8"
```

php/kuiper/composer.json
```diff
@@ -0,0 +1,26 @@
...
+  },
+  "require-dev": {
+    "kuiper/component-installer": "^0.1"
+  },
```
waghanza, line 9: do we need this ?

php/kuiper/config.yaml
waghanza, line 1: don't kuiper require pcntl ?

php/kuiper/composer.json
wenbinye: kuiper framework is not released currently. ^0.8 equal to >=0.8 and < 0.9, ~0.8 equal to >= 0.8 and < 1.0，previous one is prefered

php/kuiper/composer.json
wenbinye: not require at runtime, only for development

waghanza: ok, maybe use `@stable` then :-)

waghanza (commented):
I have a notice 
```
[2023-03-21 09:31:29 *17.2]	NOTICE	ProcessFactory::end() (ERRNO 1005): session#42 does not exists
```
when running instead the container

php/kuiper/composer.json
waghanza: you are right, `composer` act differently for versions behind 1.0. maybe using `~0.8.0` will be clearer then

php/kuiper/config.yaml
wenbinye: I think it is not necessary. Swoole can handle signal well.

waghanza (approved):

PR #6295, opened 2023-04-20
restheart, Use configuration override file in place of env var

ujibang:
as discussed I switched form env var to configuration file


java/restheart/config.yaml
```diff
@@ -5,17 +5,8 @@ framework:
...
-    /core/name->\"restheart-benchmarker\";
-    /core/worker-threads->0;
-    /core/requests-limit->100000;
-    /fullAuthorizer/enabled->true;
-    /logging/log-level->\"OFF\";
-    /ping/enabled->false;
-
 binaries:
   - target/benchmark.jar
+  - etc/conf-overrides.yml
```
waghanza, line 10: sorry, this not very obvious, but this should be upload in all containers

```suggestion
  - etc/conf-overrides.yml
  
files:
  - etc/conf-overrides.yml
```

java/restheart/etc/conf-overrides.yml
waghanza, line 1: I wonder why not using `properties` files with thing like `server.port=3000` ? or a more standard yaml file ?

waghanza (commented):
thanks @ujibang. a few questions however

java/restheart/etc/conf-overrides.yml
waghanza: cc @ujibang

ujibang: This is a configuration override files that modifies the default configuration.

For more info see https://restheart.org/docs/configuration

PR #6303, opened 2023-04-23
Added same release profile to all rust frameworks & updated all dependencies to same as well.

krishnaTORQUE:


waghanza (commented):
is there a way to mutualize config ? I mean either to use compilations, or to have a configuration file ?

krishnaTORQUE:
> is there a way to mutualize config ? I mean either to use compilations, or to have a configuration file ?

As far I know, there is no way.

waghanza (approved):

Issue #6317, opened 2023-04-28
Practical proposal for a minimal I/O benchmark


mindplay-dk:
Currently, this is a benchmark of raw HTTP round-trip overhead and basic routing.

This is meaningful, but doesn't tell you much about real-world scenarios, where actual body content needs to be sent: there is no real-world application in which every route returns an empty body.

I've read the other issues regarding this issue, including #110, #3803 and #1728.

Other proposals have suggested making database connections, which doesn't strictly say much about web frameworks, and starts to move into the territory of full-stack frameworks, which most of these frameworks aren't. You would need to pick database integrations for each of the implementations, which moves away from testing raw framework overhead. It's also considerably more work to implement. Things like JSON encoding gets you into similar situations: now you're benchmarking the standard language run-time (or a library) rather than the framework.

**I'd like to propose something simpler:**

* Add 3 files to a folder in the root of the repository, e.g. `10kb.txt`, `1mb.txt`, `100mb.txt`.
* All implementations must support requests to e.g. `files/10kb.txt` etc. and return the file.
* `10kb.txt` *should* be cached in-memory: this will indicate raw output performance. (REST APIs, cached static assets, etc.)
* The other two files *must not* be cached: this will indicate I/O overhead in general. (static assets, media, etc.)

This should be fairly easy to add to almost every implementation, and will make the results a lot more realistic. People looking to use a framework for a REST API might be more interested in the `10kb.txt` results, while someone looking to serve images or videos might look at the `1mb.txt` or `100mb.txt` results, and so on.

Of course, this won't tell you anything about JSON, XML or GraphQL parsing or encoding overhead - but this is currently a benchmark of raw server overhead, and this proposal keeps it that way. It's still useful information for someone building from scratch, or someone choosing a web server for building a library - it can inform you about baseline server performance, all application logic aside.

And it's probably a more reasonable and realistic next step, as there are *a lot* of benchmarks in *many* different tech stacks that need to be updated.

Add would suggesting adding something like `features: [ "io" ]` to `config.yaml` to indicate this benchmark feature is implemented, so that benchmarks can progressively move forward - and eventually adding an option to the website to only display frameworks that implement this feature.

cyrusmsk:
Any news about improvements?

waghanza:
No, but it should be the next move for this project.

Not sure about an incremental update (hard to deal with with front end ui).

'll work on something that reflect that reflect real world workload + easy to implement

[...]

Issue #1064, opened 2019-03-10
Add uWebSockets.js (Node)


ewwink:
It claim faster than fasthttp (go) and maybe can beat japronto for requests per seconds :wink: 

+ Language : node
+ Repo : https://github.com/uNetworking/uWebSockets.js
+ Author : @uNetworking

![benchmark](https://raw.githubusercontent.com/uNetworking/uWebSockets/master/misc/bigshot_lineup.png)

waghanza:
I think this could be categorized as node, since app are written in javascript

waghanza:
can we do a non  HTTP1/1 app with **uWebSocket** => without **SSL** ?

ewwink:
I don't know, I can't find about that in the [source code](https://github.com/uNetworking/uWebSockets/blob/master/src/HttpResponse.h#L175) and [docs](https://htmlpreview.github.io/?https://raw.githubusercontent.com/uNetworking/uWebSockets.js/master/docs/generated/index.html)

waghanza:
I'm not against adding this here, but flagged as experimental since not "native"

mahdisml:
> I'm not against adding this here, but flagged as experimental since not "native"

its stable now .
it have two version :
cpp : https://github.com/uNetworking/uWebSockets
javascript (NodeJS) : https://github.com/uNetworking/uWebSockets.js

i tested nodejs version , it have awesome performance ! it handled more requests than Go-FastHttp and Rust-Actix with less memory usage .
i'm really surprised ! 😨

waghanza:
@mahdisml what I meant bu saying `not native`, is that this implementation is not like native one (`node` or `c++`), not related to stability

thus, I'm happy to read any `PR` :heart: 

/cc @aichholzer

waghanza:
@alexhultman Did you consent of adding  `uWebSocket` here ?

alexhultman:
You are making mistakes in your benchmarking,

node (12.7) | sifrr (0.0) | 176261.33 | 155.18 MB
c (11) | agoo-c (0.5) | 175676.67 | 101.52 MB

176261 is not "better" than 175676. They are the same. They are insignificantly different.

You are not reporting CPU-time usage, so it is impossible to debug why they are the same, but I bet you are making the same exact mistake TechEmpower make;

You're not keeping track of CPU-time usage so you cannot keep track of when other parts bottleneck (such as your networking set-up).

[...]

PR #6318, opened 2023-04-28
Improved benchmarking of uWebSockets.js

mindplay-dk:
* Add a dedicated benchmark for `uWebSockets.js`, so we can compare against frameworks that use it.
* Fix a performance issue with existing `routejs-uwebsocket` benchmark.
* Upgrade to latest version of `uWebSockets.js` in relevant benchmarks, for realistic comparison among them.

### DIFF
--- a/javascript/routejs-uwebsocket/app.js
+++ b/javascript/routejs-uwebsocket/app.js
@@ -12,10 +12,11 @@ app
   .post("/user", (req, res) => res.end(""))
   .use((req, res) => res.writeStatus("404").end(""));
 
+const handler = app.handler();
+
 uWS
   .App()
   .any("/*", (res, req) => {
-    const handler = app.handler();
     handler(req, res);
   })
   .listen(3000, () => {
--- a/javascript/routejs-uwebsocket/package.json
+++ b/javascript/routejs-uwebsocket/package.json
@@ -1,7 +1,7 @@
 {
   "dependencies": {
     "@routejs/router": "~2.1.0",
-    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.20.0"
+    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.23.0"
   },
   "type": "module"
 }
--- a/javascript/sifrr/package.json
+++ b/javascript/sifrr/package.json
@@ -2,6 +2,6 @@
   "dependencies": {
     "@sifrr/server": "~0.0.6",
     "graphql": "*",
-    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.20.0"
+    "uWebSockets.js": "github:uNetworking/uWebSockets.js#v20.23.0"
   }
 }
--- a/javascript/uwebsockets/app.js
+++ b/javascript/uwebsockets/app.js
@@ -1,0 +1,19 @@
+import uWS from "uWebSockets.js";
+
+uWS
+  .App()
+  .get("/", (res, req) => {
+    res.end();
+  })
+  .get("/user/:id", (res, req) => {
+    res.end(req.getParameter(0));
+  })
+  .post("/user", (res, req) => {
+    res.end()
+  })
+  .any("/*", (res, req) => {
+    res.writeStatus("404").end();
+  })
+  .listen(3000, () => {
+    console.log("Server started on port 3000");
+  });
--- a/javascript/uwebsockets/config.yaml
+++ b/javascript/uwebsockets/config.yaml
@@ -1,0 +1,9 @@
+framework:
+  website: github.com/uNetworking/uWebSockets.js
+  version: 20.23.0
+
+build_deps:
+  - git # required by uWebSockets.js
+
+bin_deps:
+  - gcompat # required by uWebSockets.js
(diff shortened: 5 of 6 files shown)


waghanza (changes requested):
Thanks for making some update, but unfortunately I'm not willing to accept `uWebSockets.js` here, because the author do not want to https://github.com/the-benchmarker/web-frameworks/issues/1064

waghanza:
I understand you frustration @mindplay-dk.

You are right when you are saying that `uWebSocket` is tested here (in fact more that twice), but this is behind the scene, to have a reference implementation, we must have @uNetworkingAB non opposition.

I can also understand that this project is young, and need improvements, but not sure using code with explicit opposition of core contributors will be for the best of this project.

-------

You have updated uws for `siffr` and `routejs`, maybe you can split this PR to have this part approved

mindplay-dk:
This PR only exists because I wanted to know how much overhead routejs (et al) add to uWS base overhead.

Is there anything I can do to help improve the situation to please that maintainer?

Anything with uWS in it already the fastest thing on that list, so I don't know what he's so unhappy about. 🤷‍♂️

mindplay-dk:
Would [this](https://github.com/the-benchmarker/web-frameworks/issues/6317) help?

waghanza:
I think so @mindplay-dk. I am actually working on https://github.com/the-benchmarker/web-frameworks/pull/3958, so I can afford time for this.

-----------

I suggest to :
- remove raw `uws` from here, so I can approve
- work on what you propose, either you or me, but it requires a long run (and I can not do it rn), I mean, it requires to change one implementation, as for review ... and change the ~250 here
- re-suggest to its author to submit `uws` after changing the base implementation

kartikk221:
@mindplay-dk The `sifrr` benchmark pretty much represents `uWebsockets.js` performance because it directly exports uWebsockets.js and provides a functional approach for doing things like writing headers which is pretty much zero cost.

mindplay-dk:
> @mindplay-dk The `sifrr` benchmark pretty much represents `uWebsockets.js` performance because it directly exports uWebsockets.js and provides a functional approach for doing things like writing headers which is pretty much zero cost.

That was my point - since we're *already* benchmarking `uWebSockets.js` indirectly, it would be nice to have the direct benchmark for comparison, precisely so the `siffr` benchmark can be seen in perspective to the "base load". :-)

waghanza:
Agreed with you @mindplay-dk, but using a piece of code (even OSS) with opposition of its author, is not ... what I expect to do.

Can you update you PR and remove direct usage of `uws` ? and probably open an other PR for that and ping its author

kartikk221:
@mindplay-dk It is certainly a bit misleading to have `sifrr` and `routejs-websocket` up there because their slight performance comes from not caching request metadata like headers, query etc etc. If there was an async handler scenario or something similar of that sort, you would see a performance decrease. This is why while `hyper-express` is not the top in the Javascript ecosystem, it is the top in the ability to provide an Express-like experience with no questions asked or catches to the request life cycle. You don't have to worry about reading headers before doing any async operation because `hyper-express` caches the metadata for you from uws even if it costs performance.

waghanza:
@kartikk221 I agree with you when you are saying that we need a more complex scenario (in order to have like a standard workload test), but if some framework want to cache it their choice, a design choice that should not be used to rank some frameworks

kartikk221:
> @kartikk221 I agree with you when you are saying that we need a more complex scenario (in order to have like a standard workload test), but if some framework want to cache it their choice, a design choice that should not be used to rank some frameworks 

I see what you are saying and would agree to an extent but it is also important to consider that the caching being done in this case from the uWS request metadata is pretty much necessary If you are to do any asynchronous work in your routes which almost all Node.js developers will be doing. Essentially, uWS requires you to read any metadata about the request in the initial synchronous call from its route handler. This means If you don't know what you want ahead of time, you won't be able to read metadata. This can start to become a big problem when you begin working with middle-wares and async handlers and you will be accessing various metadata like headers, query, path etc etc throughout your code.

If the performance of some web servers drops the moment you test them with something like reading headers or doing any sort of async work, which is something almost 100% of Node.js developers will be doing, then did those webservers truly represent their performance against others well? That's the question we want to consider.

waghanza:
I think we are on the same line @kartikk221. Some frameworks caches, some frameworks don't. It their choice. To have the fairest benchmark we can, we should stick to have some realistic scenarios (async part of them imho), so if some frameworks are not able to have good figures for async, not sure we should worry here ... but again having standard workload scenarios is the biggest discussion we need to have before considering this project as **mature enough**

kartikk221:
Agreed, adding more workloads like async, echoing body, headers etc etc shouldn't be unfair but rather improve the accuracy of the results.

waghanza:
Yes, I'm ok to start a discussion based on this. There is a lot of ideas here, but separated in may issues

sginiatulin:
> Yes, I'm ok to start a discussion based on this. There is a lot of ideas here, but separated in may issues

I believe this PR could be closed, as it is already merged. Also will be interesting to see your update on roadmap of the repo. Some plans and upcoming features, etc. :)

cyrusmsk:
> Yes, I'm ok to start a discussion based on this. There is a lot of ideas here, but separated in may issues

WebSockets already merged, but could you elaborate about plans for this project? About new improvements for the test?

waghanza:
Maybe a more complete scenario which mimic database calls for example, some random number generation .... @cyrusmsk

PR #6328, opened 2023-05-02
Update server.nimble

bung87:


nim/scorper/server.nimble
```diff
@@ -8,4 +8,4 @@ license = "MIT"
 # Dependencies
 
 requires "nim >= 1.3.1"
-requires "scorper >= 1.0 & < 1.1"
+requires "scorper >= 1.0 & <= 1.1"
```
waghanza, line 11:
```suggestion
requires "scorper >= 1.1 & < 1.2"
```

waghanza (changes requested):
missing `config.yaml` (to update version number that will be displayed)

waghanza (approved):

PR #6372, opened 2023-05-20
update happyx version and happyx server 👀

Ethosa:
`happyx` version changed to `0.20.0`

### DIFF
--- a/nim/happyx/config.yaml
+++ b/nim/happyx/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: HapticX/happyx
-  version: 0.20
+  version: '0.20'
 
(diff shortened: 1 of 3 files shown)


nim/happyx/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: HapticX/happyx
-  version: 0.16
+  version: 0.20
```
waghanza, line 3:
```suggestion
  version: '0.20'
```
because of yaml parsing in ruby :-)

waghanza (changes requested):

waghanza (approved):

PR #6380, opened 2023-05-23
[Crystal] Update spider-gazelle to v6

waghanza:
cc @stakach


stakach:
I'll write a pull request for you, we removed magic methods

stakach:
use #6381

PR #6391, opened 2023-05-24
Update DESCRIPTION

JohnCoene:
So very sorry, I just realised it's because it came off CRAN, this should fix it.

Fixes #6300

### DIFF
--- a/r/ambiorix/DESCRIPTION
+++ b/r/ambiorix/DESCRIPTION
@@ -1,4 +1,5 @@
 Depends:
   ambiorix (>= 2.1.0)
 Remotes: devOpifex/ambiorix
-Imports: ambiorix
+Remotes: 
+  devOpifex/ambiorix


r/ambiorix/DESCRIPTION
```diff
@@ -1,4 +1,5 @@
 Depends:
   ambiorix (>= 2.1.0)
 Remotes: devOpifex/ambiorix
-Imports: ambiorix
+Remotes:
```
waghanza, line 4: seems  a duplicate with above line @JohnCoene

waghanza:
Seems that we use head => https://github.com/the-benchmarker/web-frameworks/actions/runs/5072401229/jobs/9110074411?pr=6391#step:7:107

Maybe we should use a tag on something like that

waghanza (approved):

PR #6421, opened 2023-06-08
Add Bandit adapter to Elixir frameworks

Sgiath:
https://github.com/mtrudel/bandit


waghanza:
feel free to add it @Sgiath, and maybe ping @mtrudel to see if he is not opposed to :-)

waghanza:
as I understand @Sgiath `bandit` is more like the runtime, I suggest that we try phoenix on default `cowboy` and `bandit`

mtrudel:
> Do you mind we introduce Bandit here @mtrudel ?

Not at all! This looks great!

mtrudel:
> as I understand @Sgiath bandit is more like the runtime, I suggest that we try phoenix on default cowboy and bandit

Yep, the tests here look solid:

* Cowboy using its native request handlers (there is no Bandit equivalent here; Plug *is* Bandit's native request handler)
* Cowboy using Plug
* Bandit using Plug
* Cowboy using Phoenix
* Bandit using Phoenix

elixir/cowboy/mix.exs
```diff
@@ -3,33 +3,14 @@ defmodule Server.MixProject do
 
   def project do
     [
-      # App config
       app: :server,
       version: "0.1.0",
-
-      # Elixir config
-      elixir: "~> 1.10",
+      elixir: "~> 1.14",
```
waghanza, line 8:
```suggestion
      elixir: "~> 1.15",
```

waghanza:
I see ... Understood

+ cowboy
+ cowboy_stream
+ plug-cowboy
+ plug-bandit
+ phoenix-cowboy
+ phoenix-bandit

Since what we want to test is framework (plug, phoenix ...). A future evolution of this project will merge for example  phoenix-cowboy and phoenix-bandit

waghanza (changes requested):

elixir/config.yaml
```diff
@@ -1,8 +1,10 @@
 language:
   version: 1.15
   files:
-    - "**/*.ex"
-    - "**/*.exs"
+    - "mix.exs"
```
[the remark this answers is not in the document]
Sgiath: I feel like just adding `mix.exs` specifically is more explicit and clear. We don't need any potential other `*.exs` files.

elixir/cowboy/mix.exs
Sgiath: Elixir 1.15 is still in release candidate state so it wouldn't work. Once it is released it can be updated, but it is not hugely important since this just indicates MINIMUM Elixir version required to run the project.

waghanza: then we should probably not use 1.15 in here
https://github.com/Sgiath/web-frameworks/blob/1b77fa8b00039bbc00351b996502448bc744ee53/elixir/Dockerfile#L1

Sgiath: It was updated here https://github.com/the-benchmarker/web-frameworks/pull/6389

The 1.15 official release is matter of days now so I wouldn't stress too much about it.

waghanza: ok, maybe the above PR was fired too early then. I do not know how elixir is officially released :-)

waghanza (approved):

PR #6425, opened 2023-06-10
chore: fix fastify deprecation

Eomm:
Hello,
just updated the Fastify bench to fix the deprecation (and get 1 req/sec more 😄)


waghanza (approved):
Thanks

PR #6429, opened 2023-06-12
Add Fast-Deno.

jim-king-2000:
In one benchmark it is in the second place and beats Hono. But in another test, it is far less then Hono. We'd like to see how it performs.


jim-king-2000:
I saw this error message:
`curl: (7) Failed to connect to 172.17.0.2 port 3000 after 3080 ms: No route to host`

What is the root cause?

waghanza:
the server has not started into container
```javascript
  if (isJSON(res) || Array.isArray(res)) return Response.json(res, { status });
                                                         ~~~~
    at https://deno.land/x/fast@6.0.0-alpha.1/decode.ts:14:58

TS2345 [ERROR]: Argument of type '{ port: number; hostname: string; cert: string; key: string; transport: "tcp"; }' is not assignable to parameter of type 'ListenTlsOptions'.
  Object literal may only specify known properties, and 'cert' does not exist in type 'ListenTlsOptions'.
    cert,
    ~~~~
    at https://deno.land/std@0.163.0/http/server.ts:721:5

Found 2 errors.
```
You can check it by login onto it
```
docker build -f .Dockerfile.deno -t test . && docker run -ti -p 3000:3000 test ash
```
and start the api with `deno run --allow-net --allow-read --allow-env app.t`

jim-king-2000:
> You can check it by login onto it

How to login? Where could I run the `docker` command?

jim-king-2000:
I guess the version of demo is the key point. May I update the version number of deno?

waghanza:
Not sure why but `deno` was still in **1.16**

Try to upgrade `javascript/deno.Dockerfile` with `denoland/deno:alpine-1.34.2` instead of `denoland/deno:alpine-1.16.1`

Normally `renovatebot` update such definition in dockerfiles

jim-king-2000:
Sure. Let me try it.

jim-king-2000:
Weird. Case `low-http-server` failed. I didn't touch this case. Is it a false alarm?

waghanza:
CI is not perfectly stable, this one is a false positive

jim-king-2000:
Glad to hear that. So, I'm waiting for the merge.

jim-king-2000:
By the way, I'd like to upgrade node.js to v20.3.0. Is that good for you?

waghanza:
no @jim-king-2000. I prefer we stick to LTS version

javascript/fast/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  website: deno.land/x/fast@6.0.0-alpha.1
+  version: 6.0.0-alpha.1
```
waghanza, line 3:
```suggestion
  version: 6.0
```

jim-king-2000:
OK. Let's wait until v20 becomes LTS.

waghanza (approved):

PR #6488, opened 2023-07-04
D frameworks

cyrusmsk:
Added 3 more frameworks:
* arsd:cgi
* handy-https
* lighttp
Also small changes to archttp and Docker file (for latest version of LDC).

Please review and test to approve the PR into master branch.


waghanza:
@adamdruppe @andrewlalis @Kripth we are going to add those framework in here. is it ok for you ?

d/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM dlang2/ldc-ubuntu AS build
```
waghanza, line 1: why this change ? what is the common usage in dlang community ?

d/Dockerfile
```diff
@@ -7,7 +7,7 @@ WORKDIR /usr/src/app
 {{/files}}
 
 ENV DC=ldc2
-RUN dub build -b release --compiler=ldc2 --verbose
+RUN dub build -b release-nobounds --compiler=ldc2 --verbose
```
waghanza, line 10: why ?

d/archttp/dub.json
```diff
@@ -1,7 +1,9 @@
 {
   "name": "server",
   "dependencies": {
-    "archttp": "~>1.1.1"
+    "archttp": "~>1.1.1",
+    "mir-cpuid": "~>1.2.10"
```
waghanza, line 5: is it a common usage ? I mean physical resources handling should not be a part of the framework / language ?

d/cgi/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: adamdruppe/arsd
+  version: 10.9.10
```
waghanza, line 3:
```suggestion
  version: 10.9
```

d/config.yaml
```diff
@@ -1,5 +1,5 @@
 language:
-  version: 2.096
+  version: 1.32.2
```
waghanza, line 2: ?

ℹ️ Not sure how to determine language version

d/handy/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: andrewlalis/handy-httpd
+  version: 7.1.0
```
waghanza, line 3:
```suggestion
  version: 7.1
```

d/lighttp/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: Kripth/lighttp
+  version: 0.5.4
```
waghanza, line 3:
```suggestion
  version: 0.5
```

waghanza (changes requested):

adamdruppe:
yeah i think benchmarks are generally silly and useless (they are so far away from any real world concerns) but i don't mind them being there.

d/archttp/dub.json
cyrusmsk: Probably it will work without it. I've just took it from other repo: https://github.com/tchaloupka/httpbench/tree/master/dlang/archttp

d/Dockerfile
cyrusmsk: I've checked other sources - for example Nim - and they disable bounds check as well. It could help with performance. Also for better performance I need to add -flto-full but it needs be tested before adding

cyrusmsk:
> @adamdruppe @andrewlalis @Kripth we are going to add those framework in here. is it ok for you ?

The approve from Andrew is here: https://github.com/andrewlalis/handy-httpd/discussions/20

d/config.yaml
cyrusmsk: D has several compilers.
DMD has versions 2.XXX - now it is 2.104 https://dlang.org/download.html#dmd
LDC (based on LLVM) has different versions. The latest is 1.32.2 https://github.com/ldc-developers/ldc/releases

d/Dockerfile
waghanza: Do whatever is recommended / best-practice for this language

d/archttp/dub.json
waghanza: seems like a hack for me, I prefer no to, but let's check with  @zoujiaqing

d/Dockerfile
adamdruppe: I recommend never using even release since it can remove useful assertions, and release-nobounds is just plain irresponsible for a real world deployment.

Only benchmarks and fools use these things.

waghanza: so we could probably also remove it for `nim` @cyrusmsk (probably with an other PR btw)

d/archttp/dub.json
cyrusmsk: I propose to try with this one. In case it doesn't give any value - we can remove it.

d/Dockerfile
cyrusmsk: Ok, I removed nobounds flag in last commit

d/Dockerfile
cyrusmsk: It seems this updated docker image needs some extra packages to be installed to run applications. I am revert to the previous one and run CI tests in my local branch. It works for everything except handy-http - probably because the compiler version too old and doesn't support some changes that were made in newer releases. Let me try to fix it somehow.

cyrusmsk:
@waghanza do you know the error with wrong URL?
docker run -p 3000:3000 -td d.archttp.default > d/archttp/cid-default.txt
docker inspect `cat d/archttp/cid-default.txt` | jq -r '.[0].NetworkSettings.IPAddress' > d/archttp/ip-default.txt
curl --retry 5 --retry-delay 5 --retry-max-time 180 --retry-connrefused http://`cat d/archttp/ip-default.txt`:3000 -v
* Closing connection -1
curl: (3) URL using bad/illegal format or missing URL

Don't know how to fix it.

waghanza:
Maybe if server is not listening to 0.0.0.0 inside the container, and on port 3000 @cyrusmsk

cyrusmsk:
It seems problem in Docker configuration. If I will use old ubuntu-based two solutions passed tests successfully. But now I'm trying to use Alpine image and faced with this problem.

cyrusmsk:
> Maybe if server is not listening to 0.0.0.0 inside the container, and on port 3000 @cyrusmsk

is it possible to echo/cat the file 'd/archttp/ip-default.txt' somehow with Docker command?
I can't understand where is everything is running.
Who is giving CMD command?

cyrusmsk:
Ok, with last commit everything passed tests: https://github.com/cyrusmsk/web-frameworks/actions/runs/5592528052/jobs/10225103834
Please review now @waghanza

d/config.yaml
waghanza: If using `ldc` package of debian, how can we be sure of version number (of the language) ?

cyrusmsk: I just saw that C uses Debian as well. So you are right - we can rely only on version from the package. Could we use it as temporary decision? While the updated versions of official (or at least community-driven) docker images are under preparation?

waghanza (changes requested):
Are you sure that `ldc` provided by debian side use d lang version 1.32 @cyrusmsk ?
https://packages.debian.org/unstable/ldc

cyrusmsk:
> Are you sure that `ldc` provided by debian side use d lang version 1.32 @cyrusmsk ?
> https://packages.debian.org/unstable/ldc

Fixed to 1.30.0. And also I saw that you are planning to move other Debian based images to Bookworm, so I've change the container to Bookworm as well. All tests are passed https://github.com/cyrusmsk/web-frameworks/actions/runs/5602125606 in local repo.

waghanza:
Could you use multi-layer container @cyrusmsk ?

cyrusmsk:
> Could you use multi-layer container @cyrusmsk ?

Updated to multi-stage container. But because D need runtime - it still required ldc package to run. Probably it is possible to build static executables, but I haven't tested them for HTTPs frameworks. Test locally run here https://github.com/cyrusmsk/web-frameworks/actions/runs/5603841227

waghanza (approved):

cyrusmsk:
Hi @waghanza
Anything else should be added to be merged? Please let me know

PR #6535, opened 2023-07-21
Update web.csproj

ShreyasJejurkar:
Enable PGO

### DIFF
--- a/csharp/aspnet-minimal-api/web.csproj
+++ b/csharp/aspnet-minimal-api/web.csproj
@@ -3,6 +3,7 @@
     <PropertyGroup>
         <TargetFramework>net7.0</TargetFramework>
         <ImplicitUsings>enable</ImplicitUsings>
+        <TieredPGO>true</TieredPGO>
     </PropertyGroup>
 
 </Project>
--- a/csharp/aspnet-mvc/web.csproj
+++ b/csharp/aspnet-mvc/web.csproj
@@ -4,6 +4,7 @@
         <TargetFramework>net7.0</TargetFramework>
         <Nullable>enable</Nullable>
         <ImplicitUsings>enable</ImplicitUsings>
+        <TieredPGO>true</TieredPGO>
     </PropertyGroup>
 
 </Project>


waghanza (commented):
what is PGO ?

ShreyasJejurkar:
It stands for Profile guided optimization! In case if you are not aware -> https://en.wikipedia.org/wiki/Profile-guided_optimization

waghanza (commented):
seems interesting. 

is it something that is **production-ready** / recommended for **production** ?

is it something that could be applied on other csharp frameworks ?

waghanza (commented):
then please replicate this option on all frameworks :-)

could it also be used on `F#` ? or not at all ?

ShreyasJejurkar:
It should work on F# as well, but I really never tried. I guess we can get this merge to C# first and then we can try F# after that!

ShreyasJejurkar:
> then please replicate this option on all frameworks :-)

Let the authors of the FW take care of them, I just made it on official aspnetcore framework!

cyrusmsk:
This is what I asked before @waghanza 
Do all kind of optimizations (like LTO, PGO, switched off checks) are allowed?

waghanza:
Not sure then why this reply was made https://github.com/the-benchmarker/web-frameworks/pull/6488#discussion_r1253091679.

I'm honestly not the best person to say if this is legitimate or not

wdyt @dom96 @panesofglass @adamdruppe @itsumura-h ?

adamdruppe:
PGO is ok. My comment was specifically about D's horrible `-release` switch - which is independent of the optimization switches, etc - which strips out array bound checking (which is important for preventing a common source of security holes, technically you are never supposed to go out of bounds anyway, but nobody *intentionally* writes security holes and you want defense in depth) and disables asserts in a way that can lead to undefined behavior.

But PGO, LTO, -O2, stuff like that, are entirely separate (in fact, D's broken -release switch doesn't even actually enable any of those things!) and those are fine to use.

ShreyasJejurkar:
I understand what you guys are suspecting here like hacky optimization and things. But let me clear one thing enabling PGO is not a hacky optimization technique, it's the compiler-level static optimization so nothing is odd there! Moreover, Dynamic PGO is by default going forward with .NET 8 (https://github.com/dotnet/runtime/pull/86225), so nothing is hacky here!

waghanza:
Ok, then.  This was specific to `D`

waghanza:
> et the authors of the FW take care of them, I just made it on official aspnetcore framework!

Not sure we should. My point is that we can merge this, if it's supported by `C#` compiler + it is a best practice, let's see it this affect negatively performances

ShreyasJejurkar:
I don't think it will degrade perf, moreover, it will improve it by far! Just look at the linked issues here at the end of this https://github.com/dotnet/runtime/pull/86225 😅😅

ShreyasJejurkar:
I guess we are ready to merge this unless someone has doubts about this!

waghanza:
I have not doubt about this configuration, but I would prefer to replicate it on each frameworks

Not sure we will use `.NET` **8**, I prefer to keep with **LTS**

waghanza (approved):

PR #6540, opened 2023-07-22
Update `nim/happyx` version

Ethosa:
`v1.7.0.` -> `v1.10.0`

### DIFF
--- a/nim/happyx/config.yaml
+++ b/nim/happyx/config.yaml
@@ -1,6 +1,6 @@
 framework:
   github: HapticX/happyx
-  version: 1.7
+  version: 1.10
 
 build_opts: "-d:beast --threads:on"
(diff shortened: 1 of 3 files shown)


nim/happyx/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: HapticX/happyx
-  version: 1.7
+  version: 1.10
```
waghanza, line 3:
```suggestion
  version: '1.10'
```
otherwise it will show `1.1`

waghanza (changes requested):

nim/happyx/config.yaml
Ethosa: @waghanza please, see updates

waghanza (approved):

PR #6577, opened 2023-08-05
v | added vex

koplenov:
Easy-to-use, modular web framework for [V](https://vlang.io/).

site: https://nedpals.github.io/vex/
source: https://github.com/nedpals/vex


waghanza:
@nedpals Is it ok for you if we can add `vex` in here ? Maybe also a quick review cool be cool

waghanza (approved):

PR #6586, opened 2023-08-08
Update config.yaml

ShreyasJejurkar:
We don't need this as we are on .NET 7.


waghanza (commented):
Is it the same for F# ?

ShreyasJejurkar:
Ideally yes, we have not touched F# yet for the PGO changes, we can add this in that corresponding issue! 

I am just checking the throughput behavior on the C# side with these PGO changes, hopefully, the next run will be sooner once this gets merged!

waghanza (approved):

PR #6606, opened 2023-08-09
Update Stric to 4.0.9

aquapi:
If the base URL does not match you can change it to the URL that you are benchmarking


aquapi:
image
I got this code run correctly and somehow this is failing @waghanza

waghanza:
Yes, CI is not so stable. I need to work on it.

waghanza (approved):

Issue #5329, opened 2022-05-17
Disable HTTP pipelininng ?


waghanza:
> It would be unfair to remove it selectively

Aligned

> to not think about HTTP pipelining

Aligned

In my opinion, we should remove it. The global idea of this benchmark is not to have the highest figures as you can, but an idea of what you **COULD** have in production

_Originally posted by @waghanza in https://github.com/the-benchmarker/web-frameworks/pull/5307#discussion_r874986694_

PR #6623, opened 2023-08-14
[D] Fix socket read error

### LINKED ISSUE #6609: [D] Frameworks issues
Hi @waghanza 

I've tried to run servers handy and serverino locally with the command:
`wrk -H 'Connection: keep-alive' -d 5s -c 512 --timeout 8 -t 8 http://127.0.0.1:3000`

They showed some values of requests, and still servers continue to work without any errors.
But I didn't use Dockers. Just run servers and wrk locally.
Can you get the error logs from your run or give any recommendations to test them locally?

waghanza:
Hi @cyrusmsk,

I open this PR in order to solve #6609

-----
Closes #6609

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,7 +53,6 @@ jobs:
           make "build"  -f "${DIRECTORY}/.Makefile"
         env:
           DIRECTORY: ${{ matrix.directory }}
-          ENGINE: ${{ matrix.engine }}
 
       - name: Wait
         run: sleep 1
@@ -63,4 +62,9 @@ jobs:
         env:
           LANGUAGE: ${{ matrix.language }}
           FRAMEWORK: ${{ matrix.framework }}
+
+      - name: Siege
+        run: wrk http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
+        env:
+          DIRECTORY: ${{ matrix.directory }}
           ENGINE: ${{ matrix.engine }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,7 +32,7 @@ jobs:
     runs-on: ubuntu-latest
 
     strategy:
-      fail-fast: false
+      fail-fast: true
       matrix: ${{fromJson(needs.matrix.outputs.matrix)}}
       max-parallel: 4
 
@@ -66,8 +66,8 @@ jobs:
 
       - name: Siege
         run: |
-          apt-get -y update 
-          apt-get -y install wrk
+          sudo apt-get -y update 
+          sudo apt-get -y install wrk
           wrk http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
         env:
           DIRECTORY: ${{ matrix.directory }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,7 +67,7 @@ jobs:
         run: |
           sudo apt-get -y update 
           sudo apt-get -y install wrk
-          wrk -H 'Connection: keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
+          wrk -H 'Connection:keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
         env:
           DIRECTORY: ${{ matrix.directory }}
           ENGINE: ${{ matrix.engine }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -28,7 +28,6 @@ jobs:
 
   test:
     needs: matrix
-
     runs-on: ubuntu-latest
 
     strategy:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -68,7 +68,7 @@ jobs:
         run: |
           sudo apt-get -y update 
           sudo apt-get -y install wrk
-          wrk http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
+          wrk -H 'Connection: keep-alive' --connections 64 --threads 8 --duration 15 --timeout 1 http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
         env:
           DIRECTORY: ${{ matrix.directory }}
           ENGINE: ${{ matrix.engine }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   - pull_request
 jobs:
   matrix:
-    runs-on: ubuntu-latest
+    runs-on: ubuntu-22.04
 
     steps:
       - uses: actions/checkout@v3
@@ -62,9 +62,13 @@ jobs:
         env:
           LANGUAGE: ${{ matrix.language }}
           FRAMEWORK: ${{ matrix.framework }}
+          ENGINE: ${{ matrix.engine }}
 
       - name: Siege
-        run: wrk http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
+        run: |
+          apt-get -y update 
+          apt-get -y install wrk
+          wrk http://`cat ${DIRECTORY}/ip-${ENGINE}.txt`:3000
         env:
           DIRECTORY: ${{ matrix.directory }}
           ENGINE: ${{ matrix.engine }}
--- a/.tasks/collect.rake
+++ b/.tasks/collect.rake
@@ -34,8 +34,8 @@ task :collect do
   hostname = ENV.fetch('HOSTNAME')
   engine = ENV.fetch('ENGINE')
 
-  `wrk -H 'Connection: keep-alive' -d 5s -c 8 --timeout 8 -t #{threads} http://#{hostname}:3000`
-  `wrk -H 'Connection: keep-alive' -d #{duration}s -c 256 --timeout 8 -t #{threads} http://#{hostname}:3000`
+  `wrk -H 'Connection:keep-alive' -d 5s -c 8 --timeout 8 -t #{threads} http://#{hostname}:3000`
+  `wrk -H 'Connection:keep-alive' -d #{duration}s -c 256 --timeout 8 -t #{threads} http://#{hostname}:3000`
 
   db = PG.connect(database)
 
@@ -67,7 +67,7 @@ task :collect do
       concurrency_level_id = res.first['id']
 
       command = format(
-        "wrk -H 'Connection: keep-alive' --connections %<concurrency>s --threads %<threads>s --duration %<duration>s --timeout 1 --script %<pipeline>s http://%<hostname>s:3000#{uri}", concurrency: concurrency, threads: threads, duration: duration, pipeline: PIPELINE[method.to_sym], hostname: hostname
+        "wrk -H 'Connection:keep-alive' --connections %<concurrency>s --threads %<threads>s --duration %<duration>s --timeout 1 --script %<pipeline>s http://%<hostname>s:3000#{uri}", concurrency: concurrency, threads: threads, duration: duration, pipeline: PIPELINE[method.to_sym], hostname: hostname
       )
 
       Open3.popen3(command) do |_, stdout, stderr|
--- a/d/Dockerfile
+++ b/d/Dockerfile
@@ -16,7 +16,7 @@ FROM debian:bookworm
 
 WORKDIR /opt/web
 
-RUN apt-get -qq update && apt-get -qy install openssl wget ldc nginx
+RUN apt-get -qq update && apt-get -qy install openssl wget ldc nginx supervisor
 
 COPY --from=build /usr/src/app/server /opt/web/server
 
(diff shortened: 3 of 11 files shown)


cyrusmsk:
Did it solved also handy server issues?

waghanza:
Nop, I opened this PR in order to find a solution, but I didn't find yet

waghanza:
@cyrusmsk I managed to run also on CI and I have the same kind of results
https://github.com/the-benchmarker/web-frameworks/actions/runs/5866851930/job/15906476528?pr=6623#step:10:119

This is however an ubuntu 22.04 not like my dist (fedora 38).

Not sure to understand how you can have different results using docker layer.

waghanza:
May you have a clue @zoujiaqing

I have the issue for `handy`, `serverino` and `vibe.d` only. The other are OK

- https://github.com/the-benchmarker/web-frameworks/actions/runs/5866851930/job/15906475910?pr=6623#step:10:119
- https://github.com/the-benchmarker/web-frameworks/actions/runs/5866851930/job/15906476068?pr=6623#step:10:119
- https://github.com/the-benchmarker/web-frameworks/actions/runs/5866851930/job/15906476371?pr=6623#step:10:119

cyrusmsk:
> Not sure to understand how you can have different results using docker layer.
@waghanza I'm running docker on M1 Mac.
This is what I got for handy:
wrk -H 'Connection: keep-alive' -d 5s -c 64 --timeout 8 -t 8 http://127.0.0.1:3000
Running 5s test @ http://127.0.0.1:3000
  8 threads and 64 connections
    Latency    21.98ms   67.59ms 428.76ms   93.65%
    Req/Sec     1.25k   411.63     1.65k    87.50%
  16393 requests in 5.04s, 1.24MB read
  Socket errors: connect 0, read 16393, write 0, timeout 0
Requests/sec:   3254.27
Transfer/sec:    251.06KB

For serverino almost the same result:
 wrk -H 'Connection: keep-alive' -d 5s -c 64 --timeout 8 -t 8 http://127.0.0.1:3000
Running 5s test @ http://127.0.0.1:3000
  8 threads and 64 connections
    Latency    25.50ms   77.43ms 473.15ms   93.27%
    Req/Sec     1.20k   463.77     1.63k    82.35%
  16389 requests in 5.03s, 0.00B read
  Socket errors: connect 0, read 16389, write 0, timeout 0
Requests/sec:   3255.35
Transfer/sec:       0.00B

waghanza:
So all requests has a socket read error like me.

cyrusmsk:
> So all requests has a socket read error like me.

Yes. But author of handy used Apache JMeter and the server for his projects and he said there were no any errors or problems. So he doesn’t know why wrk thinks there is read errors.
So maybe we could then just exclude those frameworks from the list.

waghanza:
For the moment they are hidden (I don't mind doing it manually), but I'm very curious about the reason(s) behind. Let's keep them for now, at least before I automate the results publication

waghanza:
Probably, Apache communications are made differently. For example keeping an opened socket

waghanza:
It could be interesting though to replicate the handy author configuration (having an Apache in front of handy) inside a docker and see what is going on.

Maybe handy handle well Unix socket communication but not http based communication @cyrusmsk

cyrusmsk:
Maybe we can ping @tchaloupka if he has any similar issues in his benchmark. He used serverino and wrk as well.

Hi Tomáš. Have you faced with any socket errors in your run with wrk?

cyrusmsk:
> It could be interesting though to replicate the handy author configuration (having an Apache in front of handy) inside a docker and see what is going on.
> 
> Maybe handy handle well Unix socket communication but not http based communication @cyrusmsk

sounds too advanced for me :)
I've prepared a Dockerfile update to switch on LTO (send it soon). But I'm not sure how to deal with low-level errors of sockets.

waghanza:
cc @zoujiaqing

andrewlalis:
@waghanza yes, I use nginx as a reverse proxy for handy-http so that I get HTTP 2 and other benefits, but I suspect that there's an implementation detail in `wrk` that's causing it to report socket read errors where there are none. I've ran tests with Apache JMeter and it does not report errors when load-testing handy-http, and various browsers have no problem interacting with it.

waghanza:
@trikko should the server close connection with keep alive header ?

trikko:
> @trikko should the server close connection with keep alive header ?

From RFC:

«Servers will usually have some time-out value beyond which they will no longer maintain an inactive connection. Proxy servers might make this a higher value since it is likely that the client will be making more connections through the same server. The use of persistent connections places no requirements on the length (or existence) of this time-out for either the client or the server.

When a client or server wishes to time-out it SHOULD issue a graceful close on the transport connection. Clients and servers SHOULD both constantly watch for the other side of the transport close, and respond to it as appropriate. If a client or server does not detect the other side's close promptly it could cause unnecessary resource drain on the network.»

adamdruppe:
a server closing the connection gracefully returns 0 to that read function but wrk only reports as an error if it is -1.

But yeah, i don't remember exactly what I did to change this in mine but it seems to work ok now.

trikko:
shutdown() vs close() ?

waghanza:
or both

https://dlang.org/library/std/socket/socket.close.html

waghanza:
or both
https://dlang.org/library/std/socket/socket.close.html

trikko:
Yes but I think you *should* wait for shutdown message to be received. But, as said, too bad if they didn't already received it: I close connection anyway. It doesn't really change anything in this practical case.

trikko:
Finally I have access to my laptop. Something strange is related to ```-H 'Connection: keep-alive'```. 
Without that, wrk does not report any error.

Http 1.1 has keep-alive by default, so that header is not required. Still trying to understand if it's a client-side or server-side problem. (and without that header it gives better results!)

trikko:
Sorry I deleted the last comment by mistake.
Today I have access to my laptop. I'm doing some tests. I see you use ```-H "connection:keep-alive"``` but http 1.1 use keep-alive by default.

Removing that argument from command line, it works fine. No errors are reported and serverino has better performance.

I'm trying to understand if it is a server-side or a client-side problem.

trikko:
Hmm I wonder if it is a bug in wrk.  Using ```-H "Connection:keep-alive"``` works fine.

Probably they don't check for spaces after ":"? 
Their header parser is a quite complex state machine :)
https://github.com/wg/wrk/blob/a211dd5a7050b1f9e8a9870b95513060e72ac4a0/src/http_parser.c#L1414

cyrusmsk:
@trikko  When I tried -H "Connection:keep-alive" I've got segfaults.. Can you share whole wrk command please?
I've used this one
`wrk -H 'Connection:keep-alive' -d 5s -c 64 --timeout 8 -t 8 http://127.0.0.1:3000`

On Docker it was just some logs: https://paste.myst.rs/9um95tts

When I tried to run locally (macOS) I've got also extra rows:

★ [051763] [l] 2023-08-17 15:53 [connectionhandler.d:0315] Socket error on read. Resource temporarily unavailable
core.exception.ArrayIndexError@../../../../../.dub/packages/serverino-0.4.5/serverino/source/serverino/daemon.d(437): index [119] is out of bounds for array of length 119
----------------
??:? onArrayIndexError [0x104d8e36f]
??:? _d_arraybounds_index [0x104d8e8ff]
../../../../../.dub/packages/serverino-0.4.5/serverino/source/serverino/daemon.d:437 void serverino.daemon.Daemon.wake!(app).wake(std.typecons.Typedef!(serverino.config.DaemonConfig*, null, null).Typedef) [0x104ba74fb]
../../../../../.dub/packages/serverino-0.4.5/serverino/source/serverino/main.d:175 int serverino.main.wakeServerino!(app).wakeServerino(ref serverino.config.ServerinoConfig) [0x104ba3997]
../../../../../.dub/packages/serverino-0.4.5/serverino/source/serverino/main.d:147 int app.__mixin8.__mixin2.mainServerinoLoop(immutable(char)[][]) [0x104ba3813]
../../../../../.dub/packages/serverino-0.4.5/serverino/source/serverino/main.d:109 _Dmain [0x104ba39db]
■ [051766] [l] 2023-08-17 15:53 [worker.d:0241] Killing worker. [REASON: socket closed?]

trikko:
Oh, yes. That's a regression i fixed some days ago. Just update to last
serverino.

waghanza:
Same for me @trikko 
https://github.com/the-benchmarker/web-frameworks/actions/runs/5891481803/job/15978641101?pr=6623#step:10:125

cyrusmsk:
For me 0.4.6 also hasn't solved the issue with socket read error(

trikko:
Also removing the header at all?

cyrusmsk:
> Also removing the whole connection header?

Yes. I’ve tried all with space and without space and without this -H option at all

trikko:
On linux:
```andrea /tmp/serv > wrk -H 'connection: keep-alive' -d 5s -c 512 --timeout 8 -t 4 http://127.0.0.1:3000
Running 5s test @ http://127.0.0.1:3000
  4 threads and 512 connections
    Latency    45.89ms  264.41ms   3.36s    97.32%
    Req/Sec     8.78k     1.25k   11.57k    66.00%
  174673 requests in 5.05s, 20.16MB read
  Socket errors: connect 0, read 174670, write 0, timeout 0
Requests/sec:  34597.58
Transfer/sec:      3.99MB
andrea /tmp/serv > wrk -H 'connection:keep-alive' -d 5s -c 512 --timeout 8 -t 4 http://127.0.0.1:3000
Running 5s test @ http://127.0.0.1:3000
  4 threads and 512 connections
    Latency   159.11ms  683.46ms   5.01s    94.42%
    Req/Sec    19.46k     7.52k   29.90k    39.61%
  298293 requests in 5.06s, 34.42MB read
Requests/sec:  58951.33
Transfer/sec:      6.80MB
```

Just to be sure, add this to your serverino config:

```
.setHttpTimeout(10.seconds)
.enableKeepAlive(180.seconds);
```

cyrusmsk:
@waghanza I've tested different tool instead of wrk (https://github.com/codesenberg/bombardier)
And it seems it doesn't have issues for serverino.

Did you test proposed changes on linux btw? Because on macOS even with 'connection:keep-alive' wrk is still showing soccer read errors.

waghanza:
Yes, both macros and Linux remain in error.

Will give another try will timeout set on the servernio side.

waghanza:
Maybe using https://github.com/lnx-search/rewrk could be a workaround

waghanza:
Wdyt about this issue @OvermindDL1 @Ethosa @tbrand @jeremyevans @ahopkins @krishnaTORQUE @ioquatix ?

ioquatix:
Try my fork of wrk it fixed some bugs.

waghanza:
Mmh, seems interesting https://github.com/wg/wrk/compare/master...ioquatix:wrk:master

krishnaTORQUE:
@waghanza  Definitely there are some bugs in header that I can see.
Also from `ubuntu-latest` >> `ubuntu-22` make sense to me.

trikko:
rewrk sounds like another app anyway, not a fork.

waghanza:
I think we should disable http pipelining. As I know, this feature is enabled by some browsers by default. But what you check here is like an API, not a website so http pipelining could be inaccurate here

waghanza:
Related to https://github.com/the-benchmarker/web-frameworks/issues/5329

trikko:
Which browsers? I don't think it is enabled by anyone. Curl completly removed pipeline support in 2019:

https://daniel.haxx.se/blog/2019/04/06/curl-says-bye-bye-to-pipelining/

ioquatix:
FWIW, I did not observe `wrk` doing pipelining, but maybe I did not look close enough.

trikko:
If only wrk could give better error reports...

adamdruppe:
I wonder what was so hard about the http pipelining. I guess I never tested it specifically so maybe I did it wrong too, but it seems like it would be trivial to implement since the tcp stream just works that way anyway.

trikko:
It is implemented indeed in Serverino (almost, there's a bug fixed here locally). It adds extra work: for example you need to keep a queue of requests (or simply a buffer) since you must not run them in parallel.

waghanza:
> > FWIW, I did not observe `wrk` doing pipelining, but maybe I did not look close enough.
> 
> We use header in https://github.com/the-benchmarker/web-frameworks/blob/master/.tasks/collect.rake#L70
> 
> PS : I confess that the codebase is not very clean, I should clean it before considering this project stable
> 
> @waghanza did you mean to edit my comment with your reply? :)

Sorry @ioquatix. I badly used GitHub Mobile app

waghanza:
I think I am confused about pipelining, and keep-live. I'll test @ioquatix fork this Saturday, to display results on Monday (if accurate for all languages)

trikko:
In the meanwhile I've released a new version of serverino.

Please notice that wrk use http/1.1 and ```keep-alive``` is enable by default.
It sounds like ```-H "connection: keep-alive"``` tricks wrk because of that space!

You should remove that header or use ```-H "connection:keep-alive"``` (but it is implicit)

ahopkins:
Sorry I'm late to the conversation. What specifically do you want me to weigh in upon? The explicit nature of `keep-alive` seems desired, even if enabled by default.

cyrusmsk:
@waghanza just to pay your attention in this PR, that handy version bumped to 7.10.4, serverino to 0.4.7 and cgi to 11.0.1

PR #6632, opened 2023-08-16
Update `happyx` to `2.2.4`

Ethosa:


nim/happyx/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: HapticX/happyx
-  version: "1.10"
+  version: "2.2"
 
-build_opts: "-d:beast --threads:on"
+build_opts: "-d:beast --threads:on -d:numThreads:4"
```
waghanza, line 5: why 4 ? is there not a way to "guess" thread number

Ethosa: hm..

PR #6687, opened 2023-09-03
D frameworks 2

cyrusmsk:
Pump version for serverino (now it should show better performance)
Also I've added LTO flag to compiler.


cyrusmsk:
@waghanza if it is possible, could you merge this PR before updating new results on Monday please

d/Dockerfile
```diff
@@ -3,33 +3,23 @@ FROM debian:bookworm AS build
...
-RUN apt-get -qy install build-essential wget openssl ldc dub libssl-dev zlib1g-dev
+RUN apt-get -qy install build-essential openssl ldc dub libssl-dev zlib1g-dev llvm-dev
 
 {{#files}}
   COPY '{{source}}' '{{target}}'
 {{/files}}
 
 ENV DC=ldc2
-RUN dub build -b release --compiler=ldc2 --verbose
+RUN dub build -b release-nobounds --compiler=ldc2
```
waghanza, line 13: I though this not recommended
https://github.com/the-benchmarker/web-frameworks/pull/6488#discussion_r1253091679

cyrusmsk: I think it depends on type of the system. If it is big web application with a lot of logic - probably no. If it is simple micro-service where owner would like to see max performance - probably fine.
And everyone else are still using it.

waghanza: In this comment, it was specific about D, seems fine in `nim` for example.

Also, the logic will grow later

d/cgi/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   github: adamdruppe/arsd
-  version: 10.9
+  version: "11.0.1"
```
waghanza, line 3:
```suggestion
  version: "11.0"
```

d/handy/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   github: andrewlalis/handy-httpd
-  version: 7.1
+  version: "7.10.4"
```
waghanza, line 3:
```suggestion
  version: "7.10"
```

d/handy/source/app.d
```diff
@@ -3,11 +3,11 @@ import std.algorithm: startsWith;
 
 void main() {
     ServerConfig cfg = ServerConfig.defaultValues();
-    cfg.workerPoolSize = 5;
+    cfg.workerPoolSize = 16;
```
waghanza, line 6: why 16 ?

d/serverino/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   github: trikko/serverino
-  version: 0.4
+  version: "0.4.7"
```
waghanza, line 3:
```suggestion
  version: 0.4
```

waghanza (changes requested):
Apart the `release-nobounds`  flag that is awkward for me, seems good

d/handy/source/app.d
cyrusmsk: I just found it is good to have around 2xCores. But I don’t know what is suggested value for this number

waghanza: then it might be interesting to use `totalCPU`, the results are made on a 8 core CPU, but it might change when running on cloud

cyrusmsk: Agree. Update handy code with totalCPUs in last commit

cyrusmsk:
Regarding 'release-no bounds' the idea of the comment related to all languages, not only to D. I believe @adamdruppe meant that usually it is not recommended, especially for large codebase.
But for benchmarks I think it is fine to enable 'release-nobounds' to see the potential speed without those checks.
But if you want we can remain on 'release'. Just results of D couldn't be compared with Nim/Rust, which are disable bounds checking.

d/Dockerfile
waghanza: Please remove this flag,  and I will update results with this PR

adamdruppe:
We should write a benchmark test that just pwns those servers.

d/Dockerfile
cyrusmsk: In last PR move back to 'release'

waghanza:
Not sure I get you @adamdruppe.

Not a compilation language dev, but seems not recommended in D
https://stackoverflow.com/questions/38147988/what-are-the-optimization-levels-on-d

@cyrusmsk The idea could be to compile as for **production**, probably the best idea is to remove those flags, not depending on language

adamdruppe:
I'm against turning off the safety checks in production code because it is an unnecessary security risk. I'd rather lose a benchmark than lose control of a live server or to lose private user data.

I'd be amusing to illustrate that dramatically by crafting a request in the "benchmark" that exploits a security hole in the servers. Bounds checks cost you some percent of performance, but do a good job plugging one of the most common types of vulnerabilities in this level of code, even if the programmer made mistakes that slipped past review.

Of course, I don't think it is realistic to do such a dramatic presentation :)

waghanza:
I understand. I'm quite understand your concern @adamdruppe and I have the same. I pursue the goal of having a tool (this project), that help decision when choosing a new tech.

This goal could not be reached if we do not stick with (vm options, compilation options....) that are not **production-ready**

waghanza (approved):

PR #6701, opened 2023-09-06
jooby: netty implementation

jknack:
- move from undertow to netty web server

### DIFF
--- a/kotlin/kooby/config.yaml
+++ b/kotlin/kooby/config.yaml
@@ -1,3 +1,11 @@
 framework:
   website: jooby.io
   version: 3.0
+
+  command: >
+    /usr/bin/java -server
+    -Xms2g -Xmx2g -XX:+UseNUMA -XX:+UseParallelGC
+    -Dio.netty.disableHttpHeadersValidation=true
+    -Dio.netty.buffer.checkBounds=false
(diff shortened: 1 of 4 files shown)


waghanza (dismissed):
LGTM but you know you can have two folders (jooby-undertow and jooby-netty) if you want.

And you can also do the same for kotlin

@jknack

jknack:
@waghanza didn't know about multiple folder/server, thank you. For now let's keep just one and let me update the kotlin version

kotlin/kooby/config.yaml
```diff
@@ -1,3 +1,11 @@
 framework:
   website: jooby.io
   version: 3.0
+
+  command: >
```
jknack, line 5: @waghanza added these lines for kotlin implementation

waghanza (approved):

waghanza:
is there a reason that jooby is named `kooby` in `kotlin` @jknack ?

waghanza:
I think it could be a good thing to rename it, but if you agree I'll do it (in another PR when I craft another UI)

Ofc if you are ok with that @jknack

jknack:
@waghanza sure. Thank you.

waghanza (approved):

Issue #6489, opened 2023-07-05
[Javascript] Add nitro


waghanza:
Hi @jamaluddinrumi,

Thanks for you proposal in https://github.com/the-benchmarker/website/discussions/43.

Feel free to open a PR ❤️ 

@danielroe Are you OK if we introduce `nitro` here ?

danielroe:
That sounds great to me! ❤️ Might also be worth adding `h3` (https://github.com/unjs/h3), which is the lower-level framework (akin to `express`) that nitro uses.

cc: @pi0

pi0:
Thanks, @waghanza for the idea. Yes, adding `h3` (with+without router) would probably make more sense here 👍🏼 Please let me know if needed any help for impl.

waghanza:
@jamaluddinrumi let me know if any trouble creating a PR

jamaluddinrumi:
@pi0 @danielroe could you takeover from here please? 🙏

PR #6703, opened 2023-09-07
javascript: add h3 / nitro

qin-guan:
Linked issue: https://github.com/the-benchmarker/web-frameworks/issues/6489

TODO:

- [ ] Add `h3` without router
- [x] Fix generated Dockerfile


javascript/nitro/tsconfig.json
```diff
@@ -0,0 +1,3 @@
+{
+  "extends": "./.nitro/types/tsconfig.json"
```
waghanza, line 2: where is this file?

waghanza:
cc @pi0 @danielroe

javascript/nitro/tsconfig.json
qin-guan: Hey, this file is generated during `npm install` as postinstall script

javascript/nitro/nitro.config.ts
```diff
@@ -0,0 +1,10 @@
+//https://nitro.unjs.io/config
+export default defineNitroConfig({
+    // File with [] in name can't be copied properly
```
qin-guan, line 3: Btw @waghanza, I am not very familiar with how Ruby works, would you happen to be able to know a solution for this issue?

Essentially, the original file is `routes/user/[id].get.ts`, however, when the Dockerfiles are generated, I am unable to add proper escaping to the file.

I believe the correct Dockerfile COPY command that should be used is:

`COPY 'routes/user/[[]id].get.ts' 'routes/user/[[]id].get.ts'`

But the generated Dockerfile is:

`COPY 'routes/user/[id].get.ts' 'routes/user/[id].get.ts'`

Thank you!! 🙏🏻 

Docker docs ref: https://docs.docker.com/engine/reference/builder/#copy

pi0:
(Nitro maintainer here) thanks for PR dear @qin-guan ❤️ Considering other frameworks i can see in benchmarks, i think it is right to only include `h3` (it is directly used by Nitro as well)

waghanza:
Agreed with @pi0. As my understanding `h3` is more accurate to be in here, since it a piece of software that use `nitro` for the http part.

-------

`Nitro` seems to me more a generic wrapper around node built-in capabilities.

-----

The main drawback is to have something clear to understand. The idea of this project is to factually compare frameworks, or pieces of software that can be used to scratch a project, no matter how low-level/high-level it is. But comparing something that could be focused on http... (`h3`) and something more generic (`nitro`) is not very relevant imho.

-------

We have something that is quite the same in `php`, `swoole` vs `basicphp`.

However, `swoole` is more focused on http... than a generic library 

There is also the cost of producing results (actually it is approximately 0, since it is on my machine) and the time consumed for maintenance...

-----

Regarding, nitro it could be good to have h3 also, just to see how it is written and how it could

javascript/nitro/nitro.config.ts
waghanza: Honestly, I could find a solution for this, but it seems very awkward for me to have some square brackets in filenames. 

It tend to keep using [a-zA-Z0-9] for filenames

qin-guan:
Thanks for the feedback! 

Let me come up with an `h3` implementation for review... 🚀

javascript/nitro/nitro.config.ts
qin-guan: I think we can resolve this if we only intend to keep the `h3` benchmarks

qin-guan:
Hey all, I've added the `h3` setup using `createRouter`, but could not figure out how to use the dynamic `/user/:id` paths in a non-router setup. 

@pi0 could you provide some guidance whether this is possible and point me in the right direction? Thank you!! 🙏🏻

javascript/h3/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: https://github.com/unjs/h3
```
waghanza, line 2:
```suggestion
  github.com: unjs/h3
```

javascript/h3/package.json
```diff
@@ -0,0 +1,5 @@
+{
+  "dependencies": {
+    "h3": "^1.8.1"
```
waghanza, line 3:
```suggestion
    "h3": "~1.8.1"
```

waghanza (commented):
Can you remove nitro folder ?

qin-guan:
@waghanza Done!

waghanza (approved):

javascript/h3/app.js
```diff
@@ -0,0 +1,26 @@
+const { createServer } = require("node:http")
+const { createApp, eventHandler, toNodeListener, getRouterParams, createRouter} = require("h3")
```
pi0, line 2: H3 is dual bundled with esm and CJS. But since you are using modern versions of node, it is possible (as also advised in docs, to use ESM imports synax. You can rename to `app.mjs` to make it happen)

Hebilicious:
Was literally about to add h3 here 😄

waghanza (approved):

qin-guan:
Thanks @waghanza for merging, so happy to finally get `h3` in the benchmarks. 

Thanks everyone for your feedback and guidance throughout the PR, really appreciate it ❤️

waghanza:
I've  merged this  PR to include `h3` in next run (monday), feel free to adjust

waghanza:
Results updated https://web-frameworks-benchmark.netlify.app/result?f=h3&l=javascript

pi0:
Thanks for the updates @waghanza! If you plan to make a public tweet about h3, maybe worth to wait for next minor version i am working on some performance tweaks (there seems to be some regressions on 1.8)

qin-guan:
Would love to work on this as well! (: I've done some SSR benchmarking previously between popular frameworks, it would be interesting to see the results in a more real world scenario.

waghanza:
> Have you considered adding basic server-side rendering scenarios? It would emulate a pretty much closer to real-world I guess. (would be happy to chat sometime about this if you like :)

No yet @pi0, but would be more than happy if you can start a discussion (in above tab) regarding to this, not only SSR, but something  more server-side that we can use in all implementations here

qin-guan:
I've started a discussion [here](https://github.com/the-benchmarker/web-frameworks/discussions/6726), so we can continue there. Have also penned down some of my initial thoughts in the discussion!

Issue #6760, opened 2023-09-25
Add "without keep alive" metric


KostyaTretyak:
Single Page Applications (which are very popular now) often use two web servers: one for static files, one for dynamic requests. Benchmarks are usually run for frameworks with dynamic requests, and  probably in 90% of cases, such a feature as "keep alive" is not used (in particular because of use a single GraphQL query).

If `wrk` is run with `-H 'Connection: close'` option it changes the results significantly. Single Page Applications work exactly in this mode, so I suggest adding the "work without keep alive" metric to your benchmarks.

PR #6761, opened 2023-09-28
refactor(ditsmod): simplifying routing and add prettier.

KostyaTretyak:


javascript/ditsmod/.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
waghanza, line 1: It could be cool to have this file in the above folder (I mean in the javascript folder) so as we can have a common style on all js / ts / json files

And even a GitHub action that lint files directly

KostyaTretyak: Done!

javascript/ditsmod/src/main.ts
```diff
@@ -1,8 +1,8 @@
 import { Application } from '@ditsmod/core';
-import { ServerOptions } from 'node:http';
+import type { ServerOptions } from 'node:http';
 
 import { AppModule } from './app/app.module.js';
 
 const serverOptions: ServerOptions = { keepAlive: true, keepAliveTimeout: 0 };
```
waghanza, line 6: I think you should disable keep alive @KostyaTretyak (server keep alive is disabled on other implementation it could mess the results, and it is totally removed in http2)

KostyaTretyak: > it could mess the results

The results are already incorrect before this Pull Request. Because the default keep alive timeout in NodeJS is 5 seconds, and for example, in fastify and NodeJS+fastify, this option is set to 75 seconds, and this has a very significant effect on the result.

If you want to disable this option, all other frameworks, at least based on NodeJS, should do so.

waghanza: not sure we are talking about the same thing

are we speaking of https://en.wikipedia.org/wiki/HTTP_persistent_connection ?

KostyaTretyak: Yes, it is.

waghanza: where is it set to 75 seconds is NestJS+fastify ?I think we should leave the default node value

KostyaTretyak: Actually - 72 seconds, but it doesn't change anything: https://github.com/fastify/fastify/blob/9a019624d4ca408ed58b00c5642a05d405b7d67c/lib/configValidator.js#L19

Alternatively, you can run the fastify app and make a request like this:

```bash
curl -i localhost:3000
```

You will see the header `Keep-Alive: timeout=72`.

KostyaTretyak: > we should leave framework default

This will not be fair, because if we leave 72 seconds of keep alive in fastify, then all frameworks in which this option is smaller will show a much worse result.

KostyaTretyak: The simplest solution would be this:

```bash
wrk -H 'Connection: close' # othe options
```

See also https://github.com/the-benchmarker/web-frameworks/issues/6760

PR #6762, opened 2023-09-29
Avoid unnecessary copy in actix implementation

dgel:


dgel:
Well I'm not a seasoned Rust programmer either, and not familiar with all these web frameworks. However, the actix framework clearly already made a copy of the string data before passing it into the get_user function (otherwise it would be a `&str`). 

I feel this is comparable to other frameworks which extract the same string using a getter function before returning it as-is, such as viz. Additionally, Poem just [moves the received string as well](https://github.com/the-benchmarker/web-frameworks/blob/master/rust/poem/src/main.rs) and rocket doesn't even create a String object in the user code, just [receiving and returning a str slice](https://github.com/the-benchmarker/web-frameworks/blob/master/rust/rocket/src/main.rs)

I'm not familiar at all with nim, but [happyx](https://github.com/the-benchmarker/web-frameworks/blob/master/nim/happyx/server.nim) for example doesn't seem to make an extra copy either.

I do agree this feels like the program doesn't do much, but there is already variation in what different frameworks do: some parse the id into an int whereas others just output the received string. Perhaps it would be more fair to standardize on what programs have to do with the received id beyond just outputting it, such as demanding that the id is parsed into an int and formatted again. Then again this would make it more of an int parsing / formatting benchmark than a web framework benchmark

waghanza:
Could you give a hand @msrd0 @krishnaTORQUE @sunli829 @chrislearn  ?

The argument seems reasonable here

msrd0 (approved):
Looking at the code for [gotham](https://github.com/the-benchmarker/web-frameworks/blob/master/rust/gotham/src/main.rs) ... we don't parse the input to an int or anything, and we don't run it through Rust's formatting code (because `to_string()` uses `Display` under the hood, and while the formatting code is pretty performant, it is still slower when formatting just one object (format-string `{}`), especially when we start with a string to begin with). So I'd say it is only fair to give the same treatment to actix - no int parsing and no formatting.

waghanza (approved):

PR #6763, opened 2023-10-01
fix `Bun.serve` command for Hono

yusukebe:
Hi.

In this PR, I've changed the entry point of Hono app. Without the `reusePort: true` option, we get a duplicate port error when starting with pm2.

The errors:

Screenshot 2023-10-01 at 22 26 50

When I run the `pm2 list` command, a process other than one fails:

Screenshot 2023-10-01 at 22 27 11

I haven't tried this to see how much it affects performance, but it would fix the problem of forking. Thanks!

### DIFF
--- a/javascript/hono/app.ts
+++ b/javascript/hono/app.ts
@@ -5,4 +5,8 @@ app.get("/", (c) => c.text(""));
 app.get("/user/:id", (c) => c.text(c.req.param("id")));
 app.post("/user", (c) => c.text(""));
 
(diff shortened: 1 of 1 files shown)


waghanza:
Thanks for this `PR` @yusukebe, seems this is the same for`stricjs` @aquapi

waghanza (approved):

yusukebe:
@waghanza Thanks!

PR #6764, opened 2023-10-02
php/coresky dir added

energy-coresky:
Please add PHP Coresky framework for benchmarking.
(fullstack)


php/coresky/composer.json
```diff
@@ -0,0 +1,6 @@
+{
+  "require": {
+    "php": ">7.0",
+    "energy/air": "dev-master"
```
marwan-rabbaa, line 4:
```suggestion
    "energy/air": "~0.501"
```

php/coresky/config.yaml
```diff
@@ -0,0 +1,14 @@
+framework:
+  website: coresky.net
+  version: 0.508
```
marwan-rabbaa, line 3:
```suggestion
  version: 0.501
```

:information_source:  There will be an update PR on each `coresky` update (made by a bot)

php/coresky/config.yaml
```diff
@@ -0,0 +1,14 @@
...
+  files:
+    - hole.base
+
+  bootstrap:
+    - mkdir -p var/cache && chown www-data var/cache
+    - mkdir -p var/gate && chown www-data var/gate
+    - mkdir -p var/jet && chown www-data var/jet
+    - mkdir -p var/mem && chown www-data var/mem
+    - chown www-data hole.base
+    - php main/warmcache.php
```
marwan-rabbaa, line 14: there is a warm-up step before stress test, is this file require in such a case ?

php/coresky/hole.base
marwan-rabbaa, line 1: what is  this fie ?

energy-coresky: PHP-SQlite3 database.
If the extension is not loaded, it is not a problem, it will not be used if there are no errors. I am almost sure that there should be no errors, the tests should pass.

energy-coresky: PHP-SQlite3 database.
If the extension is not loaded, it is not a problem, it will not be used if there are no errors. I am almost sure that there should be no errors, the tests should pass.

php/coresky/config.yaml
energy-coresky: php main/warmcache.php - required (prepare cached version before start).
All other don't required if tests passed without errors.

php/coresky/composer.json
energy-coresky: Wait a little.. I will prepare fixed release in the packagist repository.
Let it be version: 
"energy/air": "~0.508"

energy-coresky: Done. Please use:
"energy/air": "0.508"

php/coresky/config.yaml
energy-coresky: Please use 0.508

waghanza (commented):
do you listen to 0.0.0.0:3000 ?

energy-coresky:
> do you listen to 0.0.0.0:3000 ?

no, port 80 only

php/coresky/main/gate.php
```diff
@@ -0,0 +1,146 @@
+<?php
```
waghanza, line 1: Why not having a bootstrap command that generates this file ? Instead of having one to upload onto the container?

php/coresky/composer.json
```diff
@@ -0,0 +1,6 @@
+{
+  "require": {
+    "php": ">7.0",
+    "energy/air": "0.508"
```
waghanza, line 4:
```suggestion
    "energy/air": "~0.508"
```

php/coresky/main/rewrite.php
```diff
@@ -0,0 +1,32 @@
+<?php
```
waghanza, line 1: Can this file be generated during the bootstrap step ?

php/coresky/main/gate.php
energy-coresky: Both of these files are generated **only at the development stage**. The development tool with web-interface is used. The information obtained using the web interface is stored in the PHP array and is part of the application code. Classic routing has been replaced in my framework with SkyGate. Try using a translator:
https://github.com/energy-coresky/air/wiki/%D0%9A%D0%BE%D0%BD%D1%82%D1%80%D0%BE%D0%BB%D0%BB%D0%B5%D1%80%D1%8B-%D0%B8-SkyGate

php/coresky/main/rewrite.php
energy-coresky: The same answer like for file: **php/coresky/main/gate.php**. Develoment tool generate this files (from web-interface data) on the develoment stage only.

energy-coresky: > Can this file be generated during the bootstrap step ?

No, this is an application code.

waghanza (approved):

energy-coresky:
Hi! Coresky has overtaken phalcon. If I screw the SWOLE extension (asynchronous processing using 0.0.0.0:3000) for PHP, I may take the first place in the PHP family. Need some advice: how much will I be able to attract attention to my work in this case? It's important to me because there are other things to do. Read what is below only if you have the time and desire to do it. I write code differently than everyone else, I violate the basic principle of PSR (PHP Standard Recommendations). Basically, I don't use namespaces. Even cartoon frameworks in your rating do not allow themselves to do this. But I can explain why sometimes it is possible and necessary to do so. Problems arising from the lack of namespace are solvable. And the emerging positive aspect is very significant and productivity is only a small part. I have a lot of original ideas, like SkyGate. I have my own coding style.  Best regards,Sergey Smolyachenko От: Marwan RabbâaОтправлено: вторник, 3 октября 2023 г. в 08:48Кому: the-benchmarker/web-frameworksКопия: Energy; AuthorТема: Re: [the-benchmarker/web-frameworks] php/coresky dir added (PR ***@***.*** approved this pull request.—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you authored the thread.Message ID: ***@***.***>

waghanza:
I think performance is great, but not the only this that developers are excited about.

Surely, performance is one interesting metric, but there is also :
- the ease of writing
- a custom style (I mean some documentation about it)
- the number of problem solved (can I connect to a DB, make a SPA on top, searching for some example)


PS : There is swoole for async php, but there is also reactphp, workerman, perhaps making a core, some adapters for reactphp and swoole and workerman is a way to attract more people

PR #6790, opened 2023-10-08
[PHP SWOOLE]Set the number of processes to be equal to the number of CPU cores

NathanFreeman:


waghanza:
Hi @NathanFreeman ,

Thanks for your interest in this project. I'm not sure which practices is recommended for `swoole` in term of worker number 
cc @sy-records 

However, enabing reuse port seems ok to me.

Could you replicate those settings on all swoole based framework ? (Search engine markup for swoole)

PR #6799, opened 2023-10-08
fix (java): use default options (no explicit)

ShreyasJejurkar:


waghanza:
cc @whiplash

whiplash:
It's possible these configuration (especially `-XX:+UseNUMA -XX:+UseParallelGC`) were copied/pasted form some tutorial on a specific Java framework. we could try and remove them. I don't foresee any issues or any special performance loss.

waghanza (approved):

PR #6800, opened 2023-10-08
fix (C#) : use defaults 

ShreyasJejurkar:


waghanza:
Why removing tiered pgo, it is the default?

ShreyasJejurkar:
It is not now, but as we decided not to use any custom options in framework so I am removing this.

And also in .NET 8 it will be ON by default at runtime level itself, so anyway we don't need this.

waghanza:
Ok. We will have this option by default in mid-November then


Let's avoid it for the couple of updates until then

waghanza:
Thanks @ShreyasJejurkar. I'm aligned with this. Maybe we should / could have the same operation for all F# frameworks

Wdyt @panesofglass @jchannon?

waghanza (approved):

ShreyasJejurkar:
@waghanza I am addressing that next. Could you please create an issue for all of these settings, so that people will update their FW implementation to respect the same.

PR #6801, opened 2023-10-08
[NEW] add `javascript/happyx` (NodeJS)

Ethosa:
This will works only on Linux.

I tested it on Ubuntu `22.04`

```shell
npm install happyx
npm i
node app.js
```

### DIFF
--- a/javascript/happyx/app.js
+++ b/javascript/happyx/app.js
@@ -1,0 +1,22 @@
+import { Server } from "happyx";
+
+
+// Create a new HappyX server
+const app = new Server("0.0.0.0", 3000);
+
+
+app.get("/", (req) => {
+  return "";
+});
+
+app.get("/user/{id}", (req) => {
+  return req.params.id;
+});
+
+app.post("/user", (req) => {
+  return "";
+});
+
+
+// start application
+app.start();
--- a/javascript/happyx/config.yaml
+++ b/javascript/happyx/config.yaml
@@ -1,13 +1,4 @@
 framework:
   github: Hapticx/happyx
   website: hapticx.github.io/happyx
-  version: 1.1.1
-
-  bootstrap:
-    - npm install happyx
-    - npm install
-
-  command: node app.js
-
-  engines:
-    - node
+  version: 1.1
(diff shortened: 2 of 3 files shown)


javascript/happyx/config.yaml
```diff
@@ -0,0 +1,13 @@
...
+  version: 1.1.1
+
+  bootstrap:
+    - npm install happyx
+    - npm install
+
+  command: node app.js
+
+  engines:
+    - node
```
waghanza, line 13: no need `node` is the default engine here

javascript/happyx/config.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+  github: Hapticx/happyx
+  website: hapticx.github.io/happyx
+  version: 1.1.1
+
+  bootstrap:
+    - npm install happyx
+    - npm install
+
+  command: node app.js
```
waghanza, line 10: we use pm2 for node based, you might remove it

javascript/happyx/config.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+  github: Hapticx/happyx
+  website: hapticx.github.io/happyx
+  version: 1.1.1
+
+  bootstrap:
+    - npm install happyx
+    - npm install
```
waghanza, line 8: no need, this is kind a standard and will be done (there is twice this command in generated dockerfile)
see `bundle exec rake config`

javascript/happyx/config.yaml
```diff
@@ -0,0 +1,13 @@
+framework:
+  github: Hapticx/happyx
+  website: hapticx.github.io/happyx
+  version: 1.1.1
```
waghanza, line 4:
```suggestion
  version: 1.1
```

javascript/happyx/app.js
```diff
@@ -0,0 +1,22 @@
...
+  return req.params.id;
+});
+
+app.post("/user", (req) => {
+  return "";
+});
+
+
+// start application
+app.start();
```
waghanza, line 22: do you listen on `0.0.0.0:3000` ?

waghanza:
Seems that there is no timeout (an open connection but no return) onto the container

Ethosa:
@waghanza have you some idea how to fix it? 🤔

waghanza:
Will try something later today

PR #6803, opened 2023-10-08
Simplify-dotnet-projects

ShreyasJejurkar:


csharp/Dockerfile
```diff
@@ -4,6 +4,7 @@ WORKDIR /usr/src/app
 
 # copy csproj and restore as distinct layers
 COPY *.csproj .
+COPY Directory.Build.props .
```
waghanza, line 7: not sure it will work, docker root directoy is `csharp/carter` not `csharp`

ShreyasJejurkar: why its set to csharp/carter? hmm, how do I achieve this now?

ShreyasJejurkar: I did ../Directory.build.props but still does not work. any idea?

waghanza: the easyiest way to do the trick @ShreyasJejurkar is, imho, to use symlinks (create main file is `csharp`, and create symlinks in all frameworks directory)

PR #6837, opened 2023-10-17
Update D versions and fix Fomo controller

cyrusmsk:
Fix Fomo's controller (based on documentation)
Update versions for D libs: cgi and handy_httpd


cyrusmsk:
Idk what is wrong with the Fomo build. Only Controller code was updated, without touching anything else

waghanza:
CI is sometimes wrong. I need some times to figure it out, as a workaround, I test it locally

waghanza (approved):

waghanza:
@amirfaramarzi @cyrusmsk this implementation does not work for `fomo`

waghanza:
I restored old version then

cyrusmsk:
I propose to exclude it until it will be fixed

cyrusmsk:
Hardcoded HTTP response is no way "production" solution

waghanza:
> I propose to exclude it until it will be fixed

A little bit rushy for me. We should keep it and exclude it after 1 or 2 rounds, in order to let @amirfaramarzi fix it

PR #6850, opened 2023-10-23
update fomo

amirfaramarzi:
Hello

I myself knew that the documentation was written in a different way, but by using this method I tried to show the best performance of fomo.

I apologize if your policies are not like this

fomo was updated to its latest version and in the form of documentation

Thanks, Amir


php/fomo/composer.json
```diff
@@ -16,7 +16,7 @@
   ],
   "require": {
     "php": ">=8.1",
-    "fomo/framework": "^2.0"
+    "fomo/framework": "^2.2"
```
waghanza, line 19:
```suggestion
    "fomo/framework": "~2.2.0"
```

php/fomo/composer.json
```diff
@@ -16,7 +16,7 @@
   ],
   "require": {
     "php": ">=8.1",
```
waghanza, line 18:
```suggestion
```

not needed

PS : the idea is to keep the minimum amount of code for a working implementation (here this line is more an indication, we are on php8.2)

waghanza (changes requested):

php/fomo/composer.json
amirfaramarzi: There will be no problem with version 8.2

waghanza (approved):

PR #6878, opened 2023-11-02
chore: update hertz config

Skyenought:
I found the configuration of the current hertz code too cumbersome compared to fiber, so I chose to update hertz to mimic the fiber code.

### DIFF
--- a/go/hertz/main.go
+++ b/go/hertz/main.go
@@ -8,20 +8,16 @@ import (
 
 func main() {
 
-	r := server.New(server.WithHostPorts(":3000"))
+	r := server.New(
+		server.WithHostPorts(":3000"),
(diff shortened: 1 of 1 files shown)


li-jin-gou:
> Thanks for the catch-up @Skyenought.
> 
> As I understand, this PR is about enabling header normalization check and not having debug for routes print.
> 
> Those are enabled by default @Dusila ?

https://github.com/the-benchmarker/web-frameworks/blob/master/go/fiber/main.go#L21
image
Configuration is disabled by default and to align some of fiber's configurations and ensure fair testing.

waghanza:
Maybe the best is the terse of this PR then.

It should be disabled by default imho 
cc @ReneWerner87 @Fenny

waghanza (approved):

PR #6887, opened 2023-11-04
[Javascript] HappyX binding

waghanza:


waghanza:
@Ethosa seems that the problem is on **alpine** using **-slin** variant is ok on my side.
maybe I could update results for monday with happyx :-)

waghanza:
however, I use pm2 on node to use all cores, except for happyx, does the underlying server take care of using all cores ?

Ethosa:
@waghanza yeap, happyx nodejs bindings compiled with httpx that should to support multi-core 🤔

waghanza:
When doing 
```
wrk -H 'Connection: keep-alive' --connections 512 --threads 8 --duration 15 --timeout 1 --script /home/waghanza/workspace/benchmark/web/pipeline.lua http://172.17.0.2:3000/
```
I have

```
Running 15s test @ http://172.17.0.2:3000/
  8 threads and 512 connectionn
    Latency   427.03ms   29.22ms 850.81ms   78.06%
    Req/Sec   253.35    231.23     0.93k    71.04%
  27384 requests in 15.06s, 4.58MB read
  Socket errors: connect 0, read 0, write 0, timeout 66
  Non-2xx or 3xx responses: 13692\n
Requests/sec:   1818.65\n
Transfer/sec:    311.69KB\n
```

cc @Ethosa

PR #6920, opened 2023-11-12
Spring WebFlux Benchmark

dmartinsid:
- Added Spring WebFlux for the-benchmark
- Fixed original Spring implementation that was using Spring WebFlux as dependency and Spring WebMvc as implementation. Now for spring folder we use only use Spring Web MVC (Thread per Request Concurrency Model)


java/spring-webflux/pom.xml
```diff
@@ -0,0 +1,40 @@
...
+		<version>3.1.5</version>
+		<relativePath/>
+	</parent>
+	<groupId>framework</groupId>
+	<artifactId>benchmark-webflux</artifactId>
+	<version>1.0.0</version>
+	<name>benchmark-webflux</name>
+	<description>Benchmark for Spring WebFlux</description>
+	<properties>
+		<java.version>17</java.version>
```
waghanza, line 17: spring is not yet compatible with 21 ?

java/spring/pom.xml
```diff
@@ -6,14 +6,14 @@
 	<parent>
 		<groupId>org.springframework.boot</groupId>
 		<artifactId>spring-boot-starter-parent</artifactId>
-		<version>[3.1,3.2)</version>
+		<version>3.1.5</version>
```
waghanza, line 9: why did you change this line ?

java/spring/pom.xml
```diff
@@ -6,14 +6,14 @@
...
 	</parent>
 	<groupId>benchmark.spring-boot</groupId>
 	<artifactId>benchmark</artifactId>
 	<version>1.0.0</version>
 	<name>benchmark</name>
-	<description>Demo project for Spring Boot</description>
+	<description>Benchmark for Spring Web Mvc (thread per request model)</description>
 
 	<properties>
 		<java.version>17</java.version>
```
waghanza, line 19: java 21 ?

waghanza (changes requested):

java/spring-webflux/pom.xml
dmartinsid: Spring Boot version 3.2, scheduled for release on November 23, will be fully compatible with Java 21. This update includes support for Virtual Threads, an exciting feature that enhances performance and efficiency, particularly in benchmark scenarios.  I think it is better to change it to Java 21 after November 23 and we also enable Virtual Threads feature.

java/spring/pom.xml
dmartinsid: Fixed versions offer stability and protect against unexpected changes, but now I see that in the whole project we used range. I will change it back to range

java/spring/pom.xml
dmartinsid: With Java 21, we can expect significant performance enhancements for this profile with Virtual Threads

java/spring-webflux/pom.xml
waghanza: could you create an issue for that ? or wait until the bot make an update PR ?

java/spring/pom.xml
waghanza: Yes, I prefer using range to avoid too maintainance. The idea is to use the latest stable version of each framework

I still have issue for clojure (fixed version) r some jvm based framework that I need to exclude milestone (range in maven could takes milestones in addition of stable)

waghanza (approved):

PR #6940, opened 2023-11-15
[Kotlin] Update ktor to 2.3

waghanza:


waghanza:
need to output gradle build in a specific file for ktor, equivalent of `jarAll` for kotlin. could you give me a hand @jaguililla ?

jaguililla:
Sure... But I'll need a couple of days, I need some time to recover from minor surgery... I'm writing this single handed   😅

waghanza:
sure ... health first, but in the mean time I was too curious and fixed it (there is just a configuration issue on port now)

PR #6942, opened 2023-11-16
Aligning hertz configuration items with gofiber

Skyenought:
### Description

In `hertz` 0.7.2, the `disableDate` and `disableContentType` configuration options have been added.

This commit aims to bring hertz configuration closer to gofiber.

### DIFF
--- a/go/hertz/main.go
+++ b/go/hertz/main.go
@@ -12,6 +12,8 @@ func main() {
 		server.WithHostPorts(":3000"),
 		server.WithDisableHeaderNamesNormalizing(true),
 		server.WithDisablePrintRoute(true),
+		server.WithDisableDefaultDate(true),
+		server.WithDisableDefaultContentType(true),
 	)
 
 	r.GET("/", func(_ context.Context, ctx *app.RequestContext) {})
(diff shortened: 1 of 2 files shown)


go/hertz/main.go
```diff
@@ -12,6 +12,8 @@ func main() {
 		server.WithHostPorts(":3000"),
 		server.WithDisableHeaderNamesNormalizing(true),
 		server.WithDisablePrintRoute(true),
+		server.WithDisableDefaultDate(true),
+		server.WithDisableDefaultContentType(true),
```
waghanza, line 16: Not sure this one should be here (maybe it should be removed from fiber too)

waghanza (changes requested):
Thanks for your contribution :tada:

go/hertz/main.go
Skyenought: GET. 
If it is determined that alignment is not needed, I turn off PR.

waghanza: Feel free to remove this kind of option in all frameworks (including `fiber`

waghanza (dismissed):

waghanza:
weird commpilation fail on `fiberè

waghanza (approved):

PR #6976, opened 2023-11-24
helidon-mp should be working

whiplash:
helidon-mp should be working


waghanza:
could you merge master ? it will decrease PR size

whiplash:
sure. let me try. I usually make a mess when merging. But I can try.

PR #6981, opened 2023-11-26
Update of 27th of November 2023

waghanza:
###### Failiing frameworks

- happyx (python) how can I enable debug mode to find the issue @Ethosa ?
- one cc @lizhichao swoole 4 is not compilable on `php:8.2-fpm` ?
- low-http-server cc @jkyberneees @schamberg97 uws version is quite old https://github.com/jkyberneees/low-http-server/blob/5f07246c2f9b6ee48404042ac65d615af1233b3c/package.json#L38


Ethosa:
@waghanza I think that issue of `python/happyx` on side of `nim/happyx`

PR #6983, opened 2023-11-27
enable virtual thread mode.

sasaki-kouhei:
## summary

The coolest feature of Java21 and SpringBoot 3.2 is Virutal Thread, so I want to include it!


waghanza:
Is it also enabled in webflux ?

waghanza:
Is it also enabled in webflux ?

waghanza (approved):

PR #6990, opened 2023-11-28
[JavaScript] Hono runs on Node.js

yusukebe:
[Hono](https://hono.dev/) runs on several JavaScript runtimes. Until now, in this benchmark, it has been running on Bun, but the performance hasn't been very impressive.

We've recently improved the performance of Hono's Node.js Adapter. Therefore, this PR proposes using the Node.js Adapter for Hono and running it on Node.js instead of Bun. This change is likely to result in faster performance than on Bun.


waghanza (approved):
Awesome 🎉

Should I understand that Hono runs also on nodejs ?

Maybe the best solution, is to have both 
```yaml
engines:
- nodejs
- bun
```

In config

Actually, it will do nothing, but I plan to have results comparing different engines

yusukebe:
Thanks.

> Should I understand that Hono runs also on nodejs ?

Yes.

> Actually, it will do nothing, but I plan to have results comparing different engines

Ah, great!

PR #7033, opened 2023-12-09
[D] Version update

cyrusmsk:
Update some frameworks version and version of compiler

### DIFF
--- a/d/Dockerfile
+++ b/d/Dockerfile
@@ -3,7 +3,7 @@ FROM debian:bookworm AS build
 WORKDIR /usr/src/app
 
 RUN apt-get -qq update
-RUN apt-get -qy install build-essential openssl libssl-dev zlib1g-dev llvm-dev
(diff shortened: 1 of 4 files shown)


waghanza (approved):

d/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM debian:bookworm AS build
+FROM debian:trixie AS build
```
waghanza, line 1: mmh, not sure we should use trixie here @cyrusmsk 

using a stable version is closest to real world usage for me

cyrusmsk: Other langs are using their own Docker images (which not always could based in stable). But we can revert to bookworm or I can try to use D-specific image (if I find one that works)

waghanza: Yes, I understand that every image could not be based on debian stable, but it's always a distribution considered as stable.

I would prefer to use the last stable, or better an official D docker image, if it exists

PR #7085, opened 2023-12-26
Use `UNARY` type for Armeria `HttpService`

ikhoon:
The default `ExchangeType` for `HttpService` is `BIDI_STREAMING` which is not optimized for non-streaming / simple requests. `UNARY` type is suitable for the current benchmark.

In fact, `ExchangeType` is automatically inferred for annotated service, gRPC, and Thrift services so normal users don't need to set the type explicitly.

### DIFF
--- a/java/armeria/src/main/java/benchmark/armeria/BenchmarkApplication.java
+++ b/java/armeria/src/main/java/benchmark/armeria/BenchmarkApplication.java
@@ -1,24 +1,37 @@
 package benchmark.armeria;
 
+import com.linecorp.armeria.common.ExchangeType;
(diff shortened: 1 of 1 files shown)


waghanza:
Thanks for your explanation @ikhoon .

Maybe we should then switch to annotations then ? I don't know how armeria is used (I mean how developers write app with it).

I just know that annotations are used by a wide majority of spring users.

It's kind of weird for me to override the framework's internal.

Regards,

ikhoon:
> Maybe we should then switch to annotations then ?

It makes sense. Many users would prefer annotations.

waghanza (approved):

PR #7114, opened 2024-01-09
Add workerman to cubex

MrEssex:


waghanza:
Not actually @MrEssex. The correct form is to have a folder named `cubex-workerman`.

This will change in the future, but for now, it is better to have separated folders (even if it creates duplicated code)

php/cubex/.workerman/public/index.php
```diff
@@ -0,0 +1,19 @@
...
+use Cubex\Workerman\CubexWorker;
+use App\Application;
+
+$loader = require_once('vendor/autoload.php');
+
+$worker = CubexWorker::create(
+  __DIR__,
+  $loader,
+  function (Cubex $cubex) { return new Application($cubex); },
+  'http://0.0.0.0:' . '3719',
```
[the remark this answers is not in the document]
MrEssex: This is no fixed

MrEssex:
Thanks for the reply @waghanza, I've updated so it is now in a separate folder. 
Cheers!

php/cubex-workerman/config.yaml
```diff
@@ -0,0 +1,7 @@
+framework:
+  github: cubex/framework
+  website: cubex.io
+  version: "4.20"
+
+  engines:
+    - php-fpm
```
waghanza, line 7:
```suggestion
    - workerman
```

php/cubex-workerman/composer.json
```diff
@@ -0,0 +1,20 @@
+{
+  "require": {
+    "cubex/framework": "~4.20.0",
+    "cubex/workerman": "*",
```
waghanza, line 4:
```suggestion
    "cubex/workerman": "@stable",
```

php/cubex-workerman/public/index.php
```diff
@@ -0,0 +1,19 @@
...
+use App\Application;
+
+$loader = require_once(dirname(__DIR__) . '/vendor/autoload.php');
+
+$worker = CubexWorker::create(
+  __DIR__,
+  $loader,
+  function (Cubex $cubex) { return new Application($cubex); },
+  'http://0.0.0.0:3000',
+)->setCount(4);
```
waghanza, line 14: why 4?

waghanza (changes requested):
feel free to add other engines if cubex supports it (road runner for example)

php/cubex-workerman/config.yaml
MrEssex: Updated the config.

php/cubex-workerman/composer.json
MrEssex: Change made

php/cubex-workerman/public/index.php
MrEssex: Good question, updated to get the nproc

php/cubex-workerman/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  website: cubex.io
+  version: "4.20"
+
+  engines:
+    - workerman
+
+language:
+  engines:
+    workerman:
+      command: php bin/index start
```
waghanza, line 12:
```suggestion
```

php/cubex-workerman/bin/index
waghanza, line 1: could you move this to `public/index.php` ?

waghanza (changes requested):
`php-fpm` => 9231.28 rps
`workerman` => 34799.69 rps

🎉 

PS : On my personal laptop (those figures will vary on next run)

php/cubex-workerman/public/index
waghanza, line 1: should be `index.php`

waghanza (changes requested):
Working (after last comment), but any wrong call leads to weird things
+ GET /ping => 200
+ POST /ping => no reply from server and an error server-side
```php
RuntimeException: unable to handle your request in /usr/src/app/vendor/cubex/framework/src/Routing/RouteProcessor.php:138
Stack trace:
#0 /usr/src/app/vendor/cubex/framework/src/Routing/RouteProcessor.php(43): Cubex\Routing\RouteProcessor->_processRoute(Object(Cubex\Context\Context), 'home', 'home')
#1 /usr/src/app/vendor/cubex/framework/src/Controller/Controller.php(57): Cubex\Routing\RouteProcessor->handle(Object(Cubex\Context\Context))
#2 /usr/src/app/vendor/cubex/framework/src/Cubex.php(334): Cubex\Controller\Controller->handle(Object(Cubex\Context\Context))
#3 /usr/src/app/vendor/cubex/workerman/src/CubexWorker.php(76): Cubex\Cubex->handle(Object(App\Application), false)
#4 /usr/src/app/vendor/workerman/workerman/Connection/TcpConnection.php(646): Cubex\Workerman\CubexWorker->onMessage(Object(Workerman\Connection\TcpConnection), Object(Workerman\Protocols\Http\Request))
#5 [internal function]: Workerman\Connection\TcpConnection->baseRead(Resource id #34, 2, Resource id #34)
#6 /usr/src/app/vendor/workerman/workerman/Events/Event.php(193): EventBase->loop()
#7 /usr/src/app/vendor/workerman/workerman/Worker.php(1635): Workerman\Events\Event->loop()
#8 /usr/src/app/vendor/workerman/workerman/Worker.php(1426): Workerman\Worker::forkOneWorkerForLinux(Object(Cubex\Workerman\CubexWorker))
#9 /usr/src/app/vendor/workerman/workerman/Worker.php(1400): Workerman\Worker::forkWorkersForLinux()
#10 /usr/src/app/vendor/workerman/workerman/Worker.php(560): Workerman\Worker::forkWorkers()
#11 /usr/src/app/public/index.php(22): Workerman\Worker::runAll()
#12 {main}
worker[none:18] exit with status 64000
```

php/cubex-workerman/public/index
MrEssex: Updated and added catch-all route

waghanza (changes requested):
curl http://127.0.0.1:3000/user -v

return 0 instead of a non 2xx http code

MrEssex:
> curl http://127.0.0.1:3000/user -v
> 

> return 0 instead of a non 2xx http code

I'm sorry I don't understand. This is the response I'm getting running locally 200 - (int) 0 response

Screenshot 2024-01-10 at 10 58 59

php/cubex-workerman/composer.json
```diff
@@ -0,0 +1,20 @@
...
+    "cubex/workerman": "@stable",
+    "packaged/polyfill-ctype": "@stable",
+    "packaged/polyfill-mbstring": "@stable",
+    "packaged/polyfill-intl": "@stable",
+    "packaged/polyfill-php71": "@stable",
+    "packaged/polyfill-php72": "@stable",
+    "packaged/polyfill-php73": "@stable",
+    "packaged/polyfill-php74": "@stable",
+    "packaged/polyfill-php80": "@stable",
+    "packaged/polyfill-php81": "@stable"
```
waghanza, line 13: if I understand each php version requires a polyfier, thus I think we need `"packaged/polyfill-php83` only

waghanza:
> > curl http://127.0.0.1:3000/user -v
> 
> > return 0 instead of a non 2xx http code
> 
> I'm sorry I don't understand. This is the response I'm getting running locally 200 - (int) 0 response
> 
> Screenshot 2024-01-10 at 10 58 59

this should not return anything, and output an error (404,422,400,, your choice)

php/cubex-workerman/composer.json
MrEssex: polyfill-php83 currently doesn't exist. 

The existence of these polyfills is simply to prevent symfony bringing in their polyfills (which are unnecessary) as we currently use their http-foundation. (Hopefully to be removed in the next major version 

Cubex can be run without any polyfill library.

waghanza: sane for 82 version ?

waghanza (approved):
some implementation seems already broken (POST /user/1 should not be a successful call for example). 

I need to add more test on the test suite to make sure the routes are correctly defined for all impl. 

The problem is not specific to `cubex` so I'm OK with this PR

PR #7137, opened 2024-01-14
V: fix dockerfile

enghitalo:


v/Dockerfile
```diff
@@ -12,7 +13,7 @@ WORKDIR /app
 
 RUN v . -prod -o server
 
-FROM alpine
+FROM alpine:latest
```
waghanza, line 16: not better to be `debian` ?

v/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM thevlang/vlang:debian as build
 
 WORKDIR /app
+RUN v up
```
waghanza, line 4: If I understand correctly, `v up` fetch the latest version. Then the version displayed on `config.yaml` will be wrong

waghanza (changes requested):

v/Dockerfile
enghitalo: v still on V 0.4.x

waghanza: for noz, but in the future `v up` could lead to 0.5 right ?

is there any param like `v up -v 0.4` ?

v/Dockerfile
```diff
@@ -1,6 +1,7 @@
 FROM thevlang/vlang:debian as build
```
waghanza, line 1: weird, this lead to alpine instead of debian
cc @koplenov

v/Dockerfile
waghanza: also should not be `v up -prod` ?

waghanza (changes requested):

v/Dockerfile
enghitalo: > for noz, but in the future `v up` could lead to 0.5 right ?
> 
> is there any param like `v up -v 0.4` ?

I can change to `latest`. What do you think?

enghitalo: > also should not be `v up -prod` ?

No

v/Dockerfile
enghitalo: I think not

waghanza (changes requested):
I have 
```v
V panic: [VEX] Failed to listen to port 3000
v hash: ca3da8b
backtrace_symbols is missing => printing backtraces is not available.
Some libc implementations like musl simply do not provide it.
```
when running the produced binary

enghitalo:
> I have
> 
> ```v
> V panic: [VEX] Failed to listen to port 3000
> v hash: ca3da8b
> backtrace_symbols is missing => printing backtraces is not available.
> Some libc implementations like musl simply do not provide it.
> ```
> 
> when running the produced binary

fixed on last commit

enghitalo:
@waghanza CI was stopped?

waghanza:
I h

> @waghanza Was CI stopped?

No. I have to approve since your are a first time contributor

cyrusmsk:
> fixed on last commit

There is also a tricky part with vweb (or with x.vweb I don't remember exactly) - that it has nice performance only if it is built with "-gc none".. It is leaking a bit but it is fine)
But other frameworks don't need it..

enghitalo:
@waghanza Can this workflow be rerun?

waghanza:
time to continue @enghitalo @cyrusmsk @koplenov ?

enghitalo:
> time to continue @enghitalo @cyrusmsk @koplenov ?

I will try finish it in next 15 days

v/Dockerfile
```diff
@@ -10,9 +15,9 @@ WORKDIR /app
   RUN {{{.}}}
 {{/bootstrap}}
 
-RUN v . -prod -o server
+RUN v . {{#build_flags}} {{{.}}} {{/build_flags}} -o server
 
-FROM alpine
+FROM thevlang/vlang:debian-dev
```
waghanza, line 20:
```suggestion
FROM debian
```

v/vex/config.yaml
```diff
@@ -1,6 +1,10 @@
 framework:
   github: nedpals/vex
-  version: 0.3.5
+  version: 0.3.6
```
waghanza, line 3:
```suggestion
  version: 0.3
```

waghanza (changes requested):

waghanza:
what is the release cycle of `v` @koplenov @enghitalo 

any reason forusing a 2023 version for pico and 2024 for vweb?

enghitalo:
> what is the release cycle of `v` @koplenov @enghitalo

There is no standard but, at most, every month.

waghanza (approved):

PR #7146, opened 2024-01-18
feat: add spry framework

medz:
Hello, I added spry framework, It's a Dart web framework.

framework docs: https://spry.fun


dart/spry/config.yaml
```diff
@@ -0,0 +1,5 @@
+framework:
+  name: Spry
```
waghanza, line 2: not used

dart/spry/pubspec.lock
waghanza, line 1: could you remove this file ? we do not keep lockfiles here, to make sure the last patched version of framework is tested

dart/spry/pubspec.yaml
```diff
@@ -0,0 +1,7 @@
+name: spry_app
+
+environment:
+  sdk: '>=3.2.0 <3.3.0'
+
+dependencies:
+  spry: ^3.2.0
```
waghanza, line 7: not remember exactly, but does it guarantee to be on 3.2.x?

waghanza (commented):
Thanks for your contribution 🎉

dart/spry/config.yaml
medz: i delete it

dart/spry/pubspec.lock
medz: i delete it

dart/spry/pubspec.yaml
medz: Yes, it is guaranteed to always work on 3.2.x

waghanza (approved):

PR #7147, opened 2024-01-18
refactor(dart): Runtime using AOT

medz:
I find the previous way of building Dockerfiles unfair compared to other frameworks.

> [!TIP]
> The final executable program of Dart server should be the AOT compiled binary file.

The previous Dockerfile used the Dart VM to run the web server, which was usually only used during development in Dart development. Formal deployment is the result of using AOT compilation.

Since VM will affect some performance, the test results of Dart-related frameworks are inaccurate.


medz:
@waghanza  Hi, I have refactored it to ensure that the Dart server-side app compiles through AOT and then runs the server.

> The previous Dart Dockerfile used Dart VM. Normally, Dart developers only use Dart VM during the development stage, and formal deployment uses AOT-compiled binaries.

medz (commented):
This confirms the universality of the code

waghanza (commented):
Is it useful then to compare AOT vs VM ?

medz:
> Is it useful then to compare AOT vs VM ?

It is useful to compare AOT and VM performance, but I personally think it is unnecessary. In the case of formal deployment in Dart, few people will use VM to run Dart apps. VMs cause a lot of unnecessary performance penalties.

In fact, when I submitted this PR, I wanted to let Dart have both AOT data and VM data, but I didn't know how to implement the dockerfile. Because there is no document found in the config.yaml file to configure the dockerfile for the engine.

medz:
I personally believe that performance testing, in a fair way, should show the best way to build the language in which the framework is built.

Dart is one of the few cross-platform languages. AOT compilation is used here just like no one would use `cargo run` to run a Rust app, but it is usually recommended to use VM during development.

dart/Dockerfile
```diff
@@ -1,15 +1,25 @@
-FROM dart:3.2
+FROM dart:3.2 AS build
 
+# Set the working directory
 WORKDIR /app
 
-ADD pubspec.yaml /app/
-
+# Add pubspec.yaml and get dependencies
+ADD pubspec.* ./
```
waghanza, line 7:
```suggestion
COPY pubspec.yaml pubspec.yaml
```

dart/Dockerfile
```diff
@@ -1,15 +1,25 @@
...
 RUN dart pub get --no-precompile
 
-ADD . /app/
-
+# Copy app source code and AOT compile it
+ADD . .
```
waghanza, line 11:
```suggestion
{{#files}}
  COPY '{{source}}' '{{target}}'
{{/files}}
```

this will copy all dart files (with the modification below on `dart/config.yaml`)

dart/config.yaml
```diff
@@ -1,12 +1,12 @@
...
-    default:
-      command: dart main.dart
+    aot:
+      command: /app/server
 
 framework:
   files:
-    - "**/*.dart"
+    - '**/*.dart'
     - pubspec.yaml
```
waghanza, line 10:
```suggestion
```

No need since it is hard-coded now on dockerfile

waghanza (changes requested):
Thanks for your explanation. If benchmarking VS vs AOT is not a requirement it is good for me, results take 24h to produce 😛 

I try to update them each monday

dart/Dockerfile
```diff
@@ -1,15 +1,25 @@
...
+# Build minimal serving image from AOT-compiled `server` executable
+FROM scratch
+COPY --from=build /runtime/ /
+COPY --from=build /app/server /app/server
+
+# Fix CI
+COPY --from=build /bin/sh /bin/sh
 
-CMD {{command}}
+CMD {{command}} {{options}}
```
waghanza, line 25: what is option `{{options}}` ?

medz: This is prepared for other frameworks in the future. Currently shelf and spry do not require it, but there are other frameworks in the dart ecosystem. As far as I know, the `serverpod` framework requires parameters.

medz: In addition, {{options}} I copied the Dockerfile from other language frameworks, because the config.yaml configuration document is missing, and I don't know where it comes from.

waghanza: fyi : {{x.y.z}} that you can use in dockerfiles in 
```yaml
x:
  y:
    z:
```
in `config.yaml`

dart/Dockerfile
```diff
@@ -1,15 +1,27 @@
...
+# AOT compile `server.dart` to server executable
+RUN dart compile exe server.dart -o server
+
+# Build minimal serving image from AOT-compiled `server` executable
+FROM scratch
+COPY --from=build /runtime/ /
+COPY --from=build /app/server /app/server
+
+# Fix CI
+COPY --from=build /bin/sh /bin/sh
```
waghanza, line 25: I do not understand this one

medz: This is in accordance with the Dart dockerize recommendation to use the smallest container for the AOT compiled program, which helps to reduce the image size, because using `scratch` (empty image) the final compiled image will be cropped to the smallest possible size (so that It contains only what is needed).

medz: In addition, I found that the ci test script of `the-benchmarker/web-frameworks` seems to inject `/bin/sh` in `CMD`. For AOT compiled programs, it only calls the executable file. If you feel there is no need to reduce the size of the image and instead include files such as deps that are no longer needed, I will update it after you let me know.

waghanza: I understand that aot could use the smallest image available, but not sure a shell is required.

At least in other environments, the shell is provided by the distribution (debian in most case), since we can not use scratch image.

If no need I prefer to remove, and even maybe use aot techniques for other languages if available

medz: /bin/sh is not required, I use `COPY --from=build /bin/sh /bin/sh` simply because {{CMD}} automatically injects calls to `/bin/sh`. I don’t know about other AOT programs. The Dart binary does not require it. In addition, I am using Swift’s vapor framework and Rust’s tokio. They do not require sh.

medz: https://github.com/the-benchmarker/web-frameworks/actions/runs/7569685394/job/20613537609#step:7:117

That's why I copied /bin/sh (shell)

image

medz:
@waghanza 

https://github.com/the-benchmarker/web-frameworks/blob/fc88364f0818b279ac2806fde97cdb772f51cdef/config.yaml#L38

I found it but I don't understand it, I don't know why this command requires the container to have shell software.

medz:
If you simply run `docker run -p 3000:3000 -td <name>`, there will be no problem. It seems that the `>` symbol is used to output the content to the file, thus causing the /bin/sh command in the image to be called. But the minimal image constructed through AOT compilation usually does not provide it.

waghanza:
> @waghanza 
> 
> https://github.com/the-benchmarker/web-frameworks/blob/fc88364f0818b279ac2806fde97cdb772f51cdef/config.yaml#L38
> 
> I found it but I don't understand it, I don't know why this command requires the container to have shell software.

This command only run the container, using its default CMD or ENTRYPOINT

medz:
Now we have two options:

- [ ] Abandon building minimal containers
- [ ] Fix it so that it no longer requires image to provide the `/bin/sh` command

waghanza:
> If you simply run `docker run -p 3000:3000 -td <name>`, there will be no problem. It seems that the `>` symbol is used to output the content to the file, thus causing the /bin/sh command in the image to be called. But the minimal image constructed through AOT compilation usually does not provide it.

No it is just to capture output of docker command. Docker run output the container id, which is used to know the IP address

dart/Dockerfile
```diff
@@ -1,15 +1,28 @@
...
 RUN dart pub get --offline --no-precompile
 
-WORKDIR /app
+# AOT compile `server.dart` to server executable
+RUN dart compile exe server.dart -o server
+
+# Build minimal serving image from AOT-compiled `server` executable
+# FROM scratch
+# COPY --from=build /runtime/ /
+# COPY --from=build /app/server /app/server
```
waghanza, line 22: L19-21 are useful

dart/Dockerfile
```diff
@@ -1,15 +1,28 @@
...
+
+# Fix CI, Figure out why and don't delete it when copying sh will succeed.
+# PS: It is not a requirement of AOT, but caused by `docker run {optoins} > {file}`.
+# COPY --from=build /bin/sh /bin/sh
```
waghanza, line 26: but L26 is not

waghanza (changes requested):

waghanza:
> I haven’t deeply understood the principle of docker run, and I don’t know why using `>` to output to a file requires image to provide sh

- `docker run` has an output (docker, not the container)
- this output its the container id
- this id is used to get the container ip 
- (the id is used on `docker inspect` command)

dart/Dockerfile
```diff
@@ -1,15 +1,24 @@
...
 RUN dart pub get --offline --no-precompile
 
-WORKDIR /app
+# AOT compile `server.dart` to server executable
+RUN dart compile exe server.dart -o server
+
+# Build minimal serving image from AOT-compiled `server` executable
+# FROM scratch
+# COPY --from=build /runtime/ /
+# COPY --from=build /app/server /app/server
```
waghanza, line 22:
```suggestion
FROM scratch
COPY --from=build /runtime/ /
COPY --from=build /app/server /app/server
```

medz:
Well, I gave up on building the minimal image in this test because I couldn't solve the problem. Of course, it goes beyond PR.

> I noticed that other AOT apps do not use the minimum image

medz:
Of course you asked me to go ahead and implement the minimal image, but I can't solve the above problem because I don't know why I get this error. And you can try to build minimal image and run it, it is successful. But in CI, he requested `/bin/sh`, which caused the failure.

waghanza:
> Well, I gave up on building the minimal image in this test because I couldn't solve the problem. Of course, it goes beyond PR.

yes, it is out of the original idea of this PR, but I'm sure it is for the best (minimal image could be a valuable argument when facing production usage)

medz:
image
I'm very puzzled. It still succeeds without copy shell, but I don't know why it fails in CI. I tried docker run > file locally and it still failed. Maybe it's a problem with docker itself?

waghanza:
I look into this

medz:
An issue should be launched for the minimum image construction plan. Because I think it is an excellent choice for other aot apps to implement the minimum image as well.

waghanza (approved):

PR #7172, opened 2024-01-26
Update mixphp

onanying:


php/mixphp/config.yaml
```diff
@@ -1,3 +1,6 @@
+language:
```
waghanza, line 1: Does not mix-php support last version of PHP ?

waghanza (changes requested):
You've fix version of PHP. Does not mix-php support 8.3 ?

PR #7189, opened 2024-01-31
[D] Update handy to v8.3

waghanza:
cc @cyrusmsk


cyrusmsk:
> cc @cyrusmsk

Actually I think we can remove handy from the benchmark. It is under active development and kinda slow (not optimized for heavy request flow).
So we can add it back after the solution will be improved.

waghanza:
Ok for me. Feel free to reintroduce when more stable 

cc @andrewlalis

PR #7190, opened 2024-02-01
chore: [Elysia] refactored to builder-pattern

yaikohi:
I refactored the elysia example app to reflect the intended way of writing elysia apps, which is by using a builder pattern.

See the [elysia docs](https://elysiajs.com/at-glance.html).

### DIFF
--- a/javascript/elysia/app.ts
+++ b/javascript/elysia/app.ts
@@ -1,15 +1,7 @@
 import { Elysia } from "elysia";
 
-const app = new Elysia();
-
-app.get("/", () => "");
-
-app.get("/user/:id", (req) => req.params.id);
-
-app.post("/user", () => "");
-
-Bun.serve({
-  fetch: app.fetch,
-  reusePort: true,
-  port: 3000
-})
+new Elysia()
+  .get("/", () => "")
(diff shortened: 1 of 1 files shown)


waghanza:
@yaikohi `node` based frameworks used `pm2` under the hood.

The primary intention of the previous code was to use reuse port, so that comparison with `node`  is not unfair.

Not sure if this implementation could fulfill this goal.

cc @SaltyAom

yaikohi:
I see. In that case the least amount of change would be something like

```typescript
import { Elysia } from "elysia";

const app = new Elysia()
  .get("/", () => "")
  .get("/user/:id", (req) => req.params.id)
  .post("/user", () => "");

Bun.serve({
  fetch: app.fetch,
  reusePort: true,
  port: 3000
})
```

But I am also curious if my initial commit would suffice.

waghanza:
To have a fair comparison with `pm2` we need to implement the cluster mode ourselves, https://github.com/oven-sh/bun/issues/2428  I think.

This implementation is not ideal, but it is ok for me.

Coyld you updatte your PR accordingly ?

waghanza:
Maybe #7192 could interest you @yaikohi

PR #7192, opened 2024-02-01
[Javascript] Clean-up clustering mode

waghanza:
This `PR` remove `pm2` usage.

To be fair with `deno`, `node` and `bun`, maybe we should remove `pm2` clustering.

What do you think  @pi0 @qin-guan @dalisoft @jkyberneees @schamberg97 @kartikk221 @mcollina ?

Regards,

### DIFF
--- a/.tasks/ci.rake
+++ b/.tasks/ci.rake
@@ -8,7 +8,7 @@ namespace :ci do
     files = Dir.glob('*/*/config.yaml') if files.include?('data.json')
 
     files += files
-             .find_all { |path| path.end_with?('Dockerfile') }
+             .find_all { |path| path.end_with?('Dockerfile') || (path.split('/').size == 2 && path.end_with?('config.yaml')) }
              .map { |path| path.split(File::SEPARATOR).shift }
              .flat_map { |language| Dir.glob(File.join(language, '*', 'config.yaml')) }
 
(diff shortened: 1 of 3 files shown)


kartikk221:
I am not sure if removing PM2 clustering usage to test the web-servers in a single threaded environment would be a step in the right direction.

Keep in mind, PM2 clustering simply spawns multiple child processes listening on the same port hence it isn't utilizing any special multi-threading technique but rather just horizontal scaling to maximize load on all available cores.

For web-servers which do not support this functionality, that should be visible to users as a deficiency in the web-server rather than limiting others to bring within a level scope.

waghanza:
Good point @kartikk221. 

I'm interested to know if `pm2` is still used in `js`/`ts` world ? and what is the _common_ practices in **production** ?

Since, for example, some  `js`/`ts` dev start using `bun`, but it does not support clustering right now

waghanza:
So using a workaround for `bun` like https://github.com/oven-sh/bun/issues/2428#issuecomment-1750544989 would be the way to go for you @kartikk221.

If there is somebody using `bun` in **production**, could you mind give us feedback? (If you could use this approach, or if it is valid for you)

waghanza:
@ragokan How can I use clustering mode for `bunicorn` ?

ragokan:
Sorry for replying late, looks like you did it correctly, thank you a lot!

> @ragokan How can I use clustering mode for `bunicorn` ?

Issue #7219, opened 2024-02-12
hono is only tested using node, should be tested using bun


bompus:
hono is native to bun, with an adapter to support node. It looks like your current implementation is only using the node engine with the node server adapter.

If tested using bun, without the node adapter, it should rank between stricjs and elysia, based on my own testing.

waghanza:
you are right @bompus , `hono` should be tested at least on
+ `node`
+ `bun`
+ `deno`

before, this is already possible, but the UI should first be adapted for that (planned but I don't have any ETA)

bompus:
@waghanza I wonder if it's possible without a UI update, to just label them with suffixes like hono (node), hono (bun), hono (deno). It makes the list longer, but wouldn't require a UI update to add an engine column/filter.

waghanza:
Sure @bompus, it could be easily done

Feel free to open a PR, I will follow 😉

bompus:
Thank you for tackling this!

I saw the latest results, and they don't line up with my own experience and tests. On Bun, hono and elysia should be right up there with stricjs, very close to it in terms of performance. However, both hono-bun and elysia are way down on the list, and I'm not sure why.

I compared the configs, cluster, app files between stricjs, hono-bun, and elysia, and they all look correct to me, but the performance just doesn't line up with what I see when I run bun cluster.ts on my own VM.

Perhaps this should be a new issue, but it's so recent I figure we could discuss here, just in case something was missed somewhere, or perhaps the results parser is parsing something incorrectly?

waghanza:
I am not sure why both frameworks should be close, but indeed stricjs seems far away

https://web-frameworks-benchmark.netlify.app/result?f=hono-bun,stricjs,elysia

Worth to investigate

A PR could be open with debug enabled to understand why this diff. Also there is the idea of timeout. The actual sieging command is removing all requests with a timeout greater than 1 sec.

Maybe we should try to get wrk raw output (stress tool direct output, without any calculation)

waghanza:
Also be aware that network in use is a loopback, not a real connection

PR #7224, opened 2024-02-14
[Javascript] Add bun / deno for hono

waghanza:
Hi @bompus,

We can add `bun` and `deno` for `hono` like this.

Regards,

----------------------------

Closes https://github.com/the-benchmarker/web-frameworks/issues/7219


waghanza:
Maybe I need a hand @yusukebe . I can not understand how to run `hono` on `bun`

yusukebe:
@waghanza 

Hmm. I don't know a good way to make Bun and Deno support multi-core. I tried pm2 before, but it did not work well.

Or we can use Web Worker?

Bun and Deno are faster in single threaded environments, but in multi-core environments, Node.js apps clustered with pm2 are faster.

SaltyAom:
AFAIK, on Bun you can use pm2 to spawn process equal to total thread on the server (which we already did by using `-i max`) and set `reusePort: true` which Bun will handle load-balancing automatically.

Here's from the definition of `Bun.serve`
```ts
    /**
     * If the `SO_REUSEPORT` flag should be set.
     *
     * This allows multiple processes to bind to the same port, which is useful for load balancing.
     *
     * @default false
     */
    reusePort?: boolean;
```

So instead of [maybe](https://hono.dev/getting-started/bun#change-port-number):
```ts
// from
export default app

// to
export default { 
  port: 3000,
  fetch: app.fetch,
  reusePort: true
}
```

Then just use PM2 to run Bun.

yusukebe:
@SaltyAom @littledivy 

Thank you both!

But maybe I'm doing it wrong, but with pm2 only one process responds.

https://github.com/the-benchmarker/web-frameworks/assets/10682/6cbdc23b-4d46-4a66-8e16-639b6ce32a24

Code:

```ts
import { Hono } from 'hono'
import { logger } from 'hono/logger'

const app = new Hono()

app.use(logger())
app.get('/', (c) => c.text(''))

export default {
  port: 3000,
  fetch: app.fetch,
  reusePort: true
}
```

Command:

```
nohup pm2-runtime start "bun app.ts" -i 8
```

waghanza:
> Then just use PM2 to run Bun.

But you need `node` to run `pm2`
```javascript
/usr/bin/env: 'node': No such file or directory
```

@SaltyAom

waghanza:
With `cluster.js`, I do not require `nodejs` since this is a native solution for `bun`.

This solution perform slightly better
+ with `pm2`
```
Running 15s test @ http://127.0.0.1:3000
  8 threads and 64 connections
    Latency   467.53us    1.29ms  58.40ms   97.39%
    Req/Sec    22.37k     6.23k   35.72k    73.41%
  2686326 requests in 15.10s, 192.14MB read
Requests/sec: 177893.60
Transfer/sec:     12.72MB
```
+ with `cluster.js`
```
Running 15s test @ http://127.0.0.1:3000
  8 threads and 64 connections
    Latency   458.26us    1.28ms  58.78ms   97.43%
    Req/Sec    22.66k     6.63k   35.80k    73.38%
  2719641 requests in 15.10s, 194.52MB read
Requests/sec: 180107.27
Transfer/sec:     12.88MB
```

waghanza:
I tried `pup` for `deno`, but
```
pup run --cmd "deno run --unstable-kv --allow-read=. --allow-net app.ts"
```
leads to
```
2024-02-15T20:28:21.938Z [LOG] [core:processes] Process 'task' loaded
Error while writing status to kv store: Deno.openKv is not a function
2024-02-15T20:28:21.940Z [LOG] [core:terminate] Termination requested
2024-02-15T20:28:21.940Z [LOG] [task:block] Process blocked, reason: terminating
Error while writing status to kv store: Deno.openKv is not a function
```

PR #7229, opened 2024-02-15
feat(adonis): add http server only

RomainLanz:
Hey there! 👋🏻 

This PR adds another benchmark to test only the HTTP server of AdonisJS 6.

### DIFF
--- a/javascript/adonisjs6-http/config.yaml
+++ b/javascript/adonisjs6-http/config.yaml
@@ -1,3 +1,3 @@
 framework:
   website: adonisjs.com
-  version: 6.2
+  version: 7.0
--- a/javascript/adonisjs6-http/config.yaml
+++ b/javascript/adonisjs6-http/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: adonisjs.com
+  version: 6.2
--- a/javascript/adonisjs6-http/package.json
+++ b/javascript/adonisjs6-http/package.json
@@ -1,0 +1,10 @@
+{
+  "type": "module",
+  "dependencies": {
+    "@adonisjs/application": "^8.0.4",
+    "@adonisjs/encryption": "^6.0.1",
+    "@adonisjs/events": "^9.0.1",
+    "@adonisjs/http-server": "^7.0.3",
+    "@adonisjs/logger": "^6.0.1"
+  }
+}
--- a/javascript/adonisjs6-http/app.js
+++ b/javascript/adonisjs6-http/app.js
@@ -24,7 +24,7 @@ server.getRouter().get('/user/:id', async (ctx) => {
   return ctx.response.ok(ctx.params.id);
 });
 
-server.getRouter().get('/user', async (ctx) => {
+server.getRouter().post('/user', async (ctx) => {
   return ctx.response.noContent();
 });
 
--- a/javascript/adonisjs6-http/app.js
+++ b/javascript/adonisjs6-http/app.js
@@ -1,0 +1,33 @@
+import { createServer } from 'node:http';
+import { Logger } from '@adonisjs/logger';
+import { Emitter } from '@adonisjs/events';
+import { Encryption } from '@adonisjs/encryption';
+import { Application } from '@adonisjs/application';
+import { defineConfig, Server } from '@adonisjs/http-server';
+
+const app = new Application(new URL('./', import.meta.url), {
+  environment: 'web',
+  importer: () => {},
+});
+
+await app.init();
+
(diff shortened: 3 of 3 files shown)


javascript/adonisjs6-http/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: adonisjs.com
+  version: 6.2
```
waghanza, line 3: Who can I ensure that this implementation use v6.2, I mean I can not find any 6.2  version in package.json

waghanza (changes requested):

javascript/adonisjs6-http/config.yaml
RomainLanz: Correct! We are in fact not using the core here 😄 
How could I represent that in the config file?

waghanza: You can't. The idea is to reflect which version of adonijsis used (not depending of module).

If I understand correctly, `http-server` is the main module here. Is their a relation between its version and adonisjs version ?

I mean `http-server` is 7.0.x, does it mean that this minor will use `adonijs` 6.2.x, like `http-server` is 7.1.x will use `adonisjs` 6.3.x

RomainLanz: > Is their a relation between its version and adonisjs version ?

Not really; since it is an internal dependency of `@adonisjs/core,` it is not "displayed" to the final user.
Maybe we can list it under `adonis-http-server-7.0`?

waghanza: Ok for me

waghanza: Just `adonisjs-http` will be enough, and version 7 in config file (will be displayed in UI)

waghanza, line 3:
```suggestion
  version: 7.0
```

javascript/adonisjs6-http/package.json
```diff
@@ -0,0 +1,10 @@
+{
+  "type": "module",
+  "dependencies": {
+    "@adonisjs/application": "^8.0.4",
+    "@adonisjs/encryption": "^6.0.1",
+    "@adonisjs/events": "^9.0.1",
+    "@adonisjs/http-server": "^7.0.3",
+    "@adonisjs/logger": "^6.0.1"
```
waghanza, line 8:
```suggestion
  "dependencies": {
    "@adonisjs/application": "*",
    "@adonisjs/encryption": "*",
    "@adonisjs/events": "*",
    "@adonisjs/http-server": "~7.0.3",
    "@adonisjs/logger": "*"
```

waghanza (changes requested):

javascript/adonisjs6-http/package.json
RomainLanz: Why is that needed?

waghanza: because we have a bot that take care of updates, and having PR each time a minor is update (for events for example) is like a too heavy task for me

also, it is, imho, a good way to do, to see for any breaking change

RomainLanz: > we have a bot that take care of updates

But semver is here for that reason. There is no need for a PR. Any patch and minor release will be installed when updating/installing the dependencies.

With a `*`, if we release a major release, it will be installed and maybe break something.

waghanza: Yes, I understand. Then their is 2 solutions :
+ Leave `*` and ping you when a breaking change is detected (each release)
+ Use semver and ping you on each minor update

I understand the concerns, but I can not maintain all versions myself for each frameworks

waghanza: PS : I use `*` strategy and I have not some many breakings

RomainLanz: > Use semver and ping you on each minor update

If we are using `^` it will be only each **major** update, using `~` in the semver condition will not allow minor update, but it was `^`  before.

waghanza: Yes, feel free to use the carret then, but I'll ping you when updates

waghanza: If you are comfortable with that, I will push it for the next update

RomainLanz: Yes, I am good with that, and I believe it is better! 👍🏻

waghanza (approved):

PR #7237, opened 2024-02-18
feat: enable multithreading on Dart/shelf

Tienisto:
Dart is by default single-threaded (similar to Node.js).
In production, this is not a problem because you can just spin up a docker container for every cpu core.
Here, the new threads are started inside a single program directly.


waghanza (commented):
Could the main function be located in a different file ? I mean like `cluster.js` and `app.js` for `elysia` in javascript

Tienisto:
I have extracted this to `cluster.dart`. `cluster.dart` should be executed instead (I don't know if this is automatically inferred or if something should be configured)

waghanza:
Maybe you should move all dart files in a `bin` directory, and specify 
```yaml
executables:
```
in `pubspec.yaml`

and modify `- "**/*.dart"` for `- "bin/**/*.dart"` in `dart/config.yaml`

Tienisto:
I made some changes so that most cluster is in the extra file while keeping the main file executable.

The problem is that in https://github.com/the-benchmarker/web-frameworks/blob/a0a635b30c48f17cfe699bc25117792151f4f752/dart/Dockerfile#L17
we specifiy the file that should be compiled `RUN dart compile exe server.dart -o server`.

waghanza:
You can change the compilation line then, and use the same technique for the other dart framework

waghanza:
Can you do the same for `spry` ?

cc @medz

Tienisto:
Done

medz:
> Can you do the same for `spry` ?
> 
> cc @medz

Maybe or not, it depends on whether `dart:io` supports it.

Spry doesn't have its own pattern, its purpose is to extend `dart:io` and I focus on the Dart backend, but I've never tried anything like this. (Because my docker only has one CPU core)

It's optional for me, but if shelf is OK, spry is OK too. That depends on @Tienisto. Because both shelf and spry are developed based on `dart:io`'s HttpServer.

medz:
Perhaps, I think this should be tried with other single-process frameworks as well.

cc @waghanza

Tienisto:
`dart:io` supports scaling. That's why there is a `shared` parameter. It was designed specifically for this purpose

waghanza (approved):

waghanza:
Published @Tienisto 🎉 

https://web-frameworks-benchmark.netlify.app/result?l=dart

PR #7272, opened 2024-03-01
Update to Hanami 2.1.0

timriley:
In #7260 and #7261, Renovate attempted to update the hanami and hanami-router gems to 2.1.0 _separately._

However, Hanami 2.1 requires these gems to be updated _together._

In this PR, I'm updating both gems, which should see the Hanami tests continue to pass.

Thanks for giving me a heads up, @waghanza!

### DIFF
--- a/ruby/hanami/Gemfile
+++ b/ruby/hanami/Gemfile
@@ -1,4 +1,4 @@
 source 'https://rubygems.org'
 
-gem 'hanami', '~> 2.0.0'
-gem 'hanami-router', '~> 2.0.0'
+gem 'hanami', '~> 2.1.0'
+gem 'hanami-router', '~> 2.1.0'


waghanza:
Could you update hanami-api also please @timriley ?

waghanza (approved):

waghanza:
I have seen that hanami-api keep its version, 0.3.0

I have another question, should not be hanami-router a dependency of hanami? In which case hanami could be used without a router ?

timriley:
> should not be hanami-router a dependency of hanami? In which case hanami could be used without a router?

It's a good question! With 2.0, one important aspect of Hanami's design is that it's no longer strictly a web framework — all the web-specific parts are now optional, which means you can use Hanami to manage apps of all kinds, such as CLI tools, chatbots, stream consumers, serverless functions and so on. And in doing this, you still get all the power and conveniences of Hanami's containers, components, environments, settings, providers, interactive console and modular slices.

In this way, I see Hanami v2 as an “everything framework.”

I expect web will remain a major use case for Hanami, but we want to make the framework as flexible as possible. This is why we can't put a hard dependency on hanami-router from hanami: because there are valid reasons to use just the "hanami" gem only.

PR #7286, opened 2024-03-06
Adding Julia Oxygen

### LINKED ISSUE #6736: [Julia] Add Oxygen
Oxygen is a newer framework for Julia. Would be interesting to see how fast is it.
Do you mind to add framework to this repo @ndortega?

https://docs.juliahub.com/Oxygen/JtS3f/1.1.2/tutorial/first_steps/
https://github.com/ndortega/Oxygen.jl

AbrJA:
Adding [Oxygen](https://github.com/OxygenFramework/Oxygen.jl) to the benchmark. I let the same structure as [Merly](https://github.com/neomatrixcode/Merly.jl)

------------

Closes #6736

### DIFF
--- a/julia/oxygen/Manifest.toml
+++ b/julia/oxygen/Manifest.toml
@@ -1,0 +1,2 @@
+# This file is machine-generated - editing it directly is not advised
+
--- a/julia/oxygen/Manifest.toml
+++ b/julia/oxygen/Manifest.toml
@@ -1,2 +1,0 @@
-# This file is machine-generated - editing it directly is not advised
-
--- a/julia/oxygen/Project.toml
+++ b/julia/oxygen/Project.toml
@@ -1,5 +1,3 @@
 [deps]
 Oxygen = "df9a0d86-3283-4920-82dc-4555fc0d1d8b"
 HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
-[compat]
-Oxygen = "^1.4.0"
--- a/julia/oxygen/Project.toml
+++ b/julia/oxygen/Project.toml
@@ -1,0 +1,5 @@
+[deps]
+Oxygen = "df9a0d86-3283-4920-82dc-4555fc0d1d8b"
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+[compat]
+Oxygen = "^1.4.0"
--- a/julia/oxygen/config.yaml
+++ b/julia/oxygen/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  github: OxygenFramework/Oxygen.jl
+  version: 1.5.1
--- a/julia/oxygen/config.yaml
+++ b/julia/oxygen/config.yaml
@@ -1,3 +1,3 @@
 framework:
   github: OxygenFramework/Oxygen.jl
-  version: 1.5.1
+  version: 1.5
--- a/julia/oxygen/server.jl
+++ b/julia/oxygen/server.jl
@@ -5,16 +5,15 @@ using Oxygen
 using HTTP
 
 @get "/" function()
-    ""
+    return text("")
 end
 
-@get "/user/{id}" function(request::HTTP.Request, id)
-    id
+@get "/user/{id}" function(request::HTTP.Request, id::String)
+    return text(id)
 end
 
 @post "/user" function ()
-   ""
+   return text("")
 end
 
-# Could be faster setting access_log = nothing and using multithreading with serveparallel
-serve(host = "0.0.0.0", port = 3000, docs = false, metrics = false)
+serveparallel(host = "0.0.0.0", port = 3000, docs = false, metrics = false, access_log = nothing)
--- a/julia/oxygen/server.jl
+++ b/julia/oxygen/server.jl
@@ -1,0 +1,20 @@
+using Pkg
+Pkg.activate(pwd())
+
+using Oxygen
+using HTTP
+
+@get "/" function()
+    ""
+end
+
+@get "/user/{id}" function(request::HTTP.Request, id)
+    id
+end
+
+@post "/user" function ()
+   ""
+end
+
+# Could be faster setting access_log = nothing and using multithreading with serveparallel
+serve(host = "0.0.0.0", port = 3000, docs = false, metrics = false)
--- a/julia/config.yaml
+++ b/julia/config.yaml
@@ -6,7 +6,7 @@ language:
     - Project.toml
   engines:
     default:
-      command: julia server.jl
+      command: julia -t auto server.jl
 
 framework:
   engines:
--- a/julia/oxygen/deps.jl
+++ b/julia/oxygen/deps.jl
@@ -1,0 +1,6 @@
+using Pkg
+
+Pkg.activate(pwd())
+Pkg.Registry.add(RegistrySpec(url = "https://github.com/JuliaRegistries/General"))
+Pkg.update()
+Pkg.resolve()


waghanza:
@ndortega Are you OK to introduce Oxygen on here ?

julia/oxygen/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Oxygen = "df9a0d86-3283-4920-82dc-4555fc0d1d8b"
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
+[compat]
+Oxygen = "^1.4.0"
```
waghanza, line 5:
```suggestion
Oxygen = "^1.5.0"
```
or

```suggestion
Oxygen = "~1.5.0"
```


Not sure which one guarantees to use the 1.5.x version and is more standard in the Julia community

julia/oxygen/Manifest.toml
```diff
@@ -0,0 +1,2 @@
+# This file is machine-generated - editing it directly is not advised
```
waghanza, line 1: Is this file really needed ?

Disclaimer: it is present for merly, but not aware of Julia usages

julia/oxygen/Project.toml
```diff
@@ -0,0 +1,5 @@
+[deps]
+Oxygen = "df9a0d86-3283-4920-82dc-4555fc0d1d8b"
```
waghanza, line 2: Do we need to keep hashes in this file ?
I mean computing the right version to use won't be done at startup by Julia ?

julia/oxygen/server.jl
```diff
@@ -0,0 +1,20 @@
...
+
+@get "/user/{id}" function(request::HTTP.Request, id)
+    id
+end
+
+@post "/user" function ()
+   ""
+end
+
+# Could be faster setting access_log = nothing and using multithreading with serveparallel
```
waghanza, line 19: Maybe you could implement this, and also make it for merly.

I don't know about Julia usage, but multi threading seems to be, in some languages, done in specialized libraries, like gunicorn in python, or in explicit implementation like cluster.ts for bun in javascript

waghanza (changes requested):

ndortega:
@waghanza  Thanks for the ping!

I just opened up a similar PR on the techempower benchmarks repo: https://github.com/TechEmpower/FrameworkBenchmarks/pull/8789

The benchmark here is missing a couple of important performance tweaks: 
- setting `access_log=nothing` will turn off internal logging by HTTP which is a huge speedup
- Explicit type definitions in request handlers
- Using render functions like `text`, `json` to format responses in the request handlers. 
- Launching julia with 2 threads and running it in parallel mode (multithreading). I've noticed that in benchmarking the julia scheduler does much better with 2 threads for small requests. 

I'd like to see changes done before getting merged so that Oxygen can be judged on it's performance at it's best. 

Other notes: 

- We shouldn't include the Manifest.toml (this locks in specific package versions) 
- We can drop the compact section altogether so that it always pulls the latest.

julia/oxygen/Project.toml
ndortega: We do need the UUID's - In Julia this is used to find the right package in the general registry. It's not tied to a specific version, that's what the `compat` section is for. If the `compat` section is dropped, then it will pull the latest version automatically

waghanza:
Thanks for the clarification @ndortega and practices for the Julia community 

Let do it @AbrJA 

For information, the difference between tech empower and this project is that the first is more developer oriented (letting developers tweak the framework to have more performance). 

Here the approach is different, we take care of consistency, code readiness, version update (framework and languages), compilation options ... to reflect a production usage

julia/oxygen/Manifest.toml
ndortega: Yup, we shouldn't need this file

julia/oxygen/Project.toml
ndortega: I'd recommend dropping the compat section or at least changing it to `Oxygen = "^1"` so it grabs the latest 1.x.x version

waghanza:
However, about 

> Launching julia with 2 threads and running it in parallel mode (multithreading). I've noticed that in benchmarking the julia scheduler does much better with 2 threads for small requests.

The idea here is to reflect what Julia could do in production. Not sure that using two threads is good (nor the opposite afaik), maybe using the number of CPU cores is accurate. Just to define a value that you are confident to use in a production environment, not a value driven by the idea of performance in a benchmark environment.

julia/oxygen/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: OxygenFramework/Oxygen.jl
+  version: 1.5.1
```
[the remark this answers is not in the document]
ndortega: Is the `version` property required?  If possible, I'd like to keep things as generic as possible so that it always pulls the latest version. Granted, I don't know if this is best practice in the benchmarking world but It would certainly make my life easier since I wouldn't have to make as many followup PR's for new versions

ndortega:
> However, about
> 
> > Launching julia with 2 threads and running it in parallel mode (multithreading). I've noticed that in benchmarking the julia scheduler does much better with 2 threads for small requests.
> 
> The idea here is to reflect what Julia could do in production. Not sure that using two threads is good (nor the opposite afaik), maybe using the number of CPU cores is accurate. Just to define a value that you are confident to use in a production environment, not a value driven by the idea of performance in a benchmark environment.

In that case running it with `julia -t auto` will let julia decide how many threads to use based on the current hardware. 

But for running in production I'd recommend tweaking this to an actual value based on your workload. If your API has quick operations (under 50ms expected response time) fewer threads will give you better performance due to less overhead in the scheduler. On the other end, if you're expecting to handle heavy traffic with longer durations then have multiple threads would benefit you more.

waghanza:
I think auto will be enough for now, at least because other languages are configured by doing so, for example javascript

julia/oxygen/Project.toml
AbrJA: Done! [5fae3de](https://github.com/the-benchmarker/web-frameworks/pull/7286/commits/5fae3de3237836c57e57c9822b1b1e3b59d46993)

julia/oxygen/Manifest.toml
AbrJA: [5fae3de](https://github.com/the-benchmarker/web-frameworks/pull/7286/commits/5fae3de3237836c57e57c9822b1b1e3b59d46993)

waghanza:
Sorry not to have figured it out previously, maybe, we should keep d'EPS section in Toml file, and remove deps.

At least for me, it appears clearer to understand the version using X.Y.Z instead of a checksum.

Also I think that http module in Julia is a dependency of oxygen no ?

@ndortega @AbrJA

julia/oxygen/Project.toml
```diff
@@ -0,0 +1,3 @@
...
+HTTP = "cd3eb016-35fb-5094-929b-558a96fad6f3"
```
waghanza, line 3: Maybe this should be dropped. This is already defined in Oxygen's project.toml

julia/oxygen/Project.toml
AbrJA: [19cb81c](https://github.com/the-benchmarker/web-frameworks/pull/7286/commits/19cb81c4ae75855de69891e5069c9ed36b8d9a5f)

julia/oxygen/Project.toml
waghanza: @AbrJA

AbrJA: I tried to avoid it but I got an error, let me check another way, I will test it locally to be sure

AbrJA: I was reviewing the implementation of @ndortega and he also include this into the [Project.toml](https://github.com/TechEmpower/FrameworkBenchmarks/pull/8789/files#diff-ad1f5944da489952393483f4aee90304851db60a10fa8dfa9263bfde68677495) Maybe the best option is to keep it

julia/oxygen/server.jl
AbrJA: [78dde22](https://github.com/the-benchmarker/web-frameworks/pull/7286/commits/78dde2266f45d603d35dc862d3f6fd95cd0ab974)

julia/oxygen/Project.toml
waghanza: seems redundant for me, but let @ndortega gives his opinion

cyrusmsk:
Thanks @AbrJA for implementing it. I was curious how will it work some time ago https://github.com/OxygenFramework/Oxygen.jl/issues/131

julia/oxygen/config.yaml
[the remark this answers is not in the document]
waghanza: This is just a descriptive version (to have this info displayed on results) and not the version installed

julia/oxygen/Project.toml
AbrJA: @ndortega What do you think?

waghanza (approved):

PR #7287, opened 2024-03-06
Fixing d tests

trikko:
- Fix serverino code
- Better compile options for all D projects

### DIFF
--- a/d/Dockerfile
+++ b/d/Dockerfile
@@ -10,12 +10,13 @@ RUN apt-get -qy install build-essential ldc dub openssl libssl-dev zlib1g-dev ll
 {{/files}}
 
 ENV DC=ldc2
-RUN dub build -b release --compiler=ldc2
+RUN dub build -b release-nobounds --compiler=ldc2
 
 FROM debian:stable
 
 WORKDIR /opt/web
 
+RUN echo "* hard nofile 16768" >> /etc/limits.conf
 RUN apt-get -qq update && apt-get -qy install openssl ldc
 
 COPY --from=build /usr/src/app/server /opt/web/server
--- a/.tasks/collect.rake
+++ b/.tasks/collect.rake
@@ -32,8 +32,8 @@ task :collect do
   hostname = ENV.fetch('HOSTNAME')
   engine = ENV.fetch('ENGINE')
 
-  `wrk -H 'Connection: keep-alive' -d 5s -c 8 --timeout 8 -t #{threads} http://#{hostname}:3000`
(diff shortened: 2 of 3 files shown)


d/Dockerfile
```diff
@@ -10,12 +10,13 @@ RUN apt-get -qy install build-essential ldc dub openssl libssl-dev zlib1g-dev ll
...
 
 ENV DC=ldc2
-RUN dub build -b release --compiler=ldc2
+RUN dub build -b release-nobounds --compiler=ldc2
 
 FROM debian:stable
 
 WORKDIR /opt/web
 
+RUN echo "* hard nofile 16768" >> /etc/limits.conf
```
waghanza, line 19: Do you mean you want to override distribution limits ?

What is the purpose?

d/Dockerfile
```diff
@@ -10,12 +10,13 @@ RUN apt-get -qy install build-essential ldc dub openssl libssl-dev zlib1g-dev ll
 {{/files}}
 
 ENV DC=ldc2
-RUN dub build -b release --compiler=ldc2
+RUN dub build -b release-nobounds --compiler=ldc2
```
waghanza, line 13: I thought nobounds options could be considered harmful to be used in production

cc @adamdruppe

adamdruppe: Yeah, I would never use it because it disables one of the main protections against buffer overflows.

waghanza: maybe if you had detected this on other languages, we should disable it @cyrusmsk

d/Dockerfile
trikko: Many distros have a very low settings for max number of open descriptors (files, sockets, and so on) and since we're testing a lot of concurrent connections this could lead to errors.

d/Dockerfile
trikko: But we're running a benchmark, not a real life environment. I think we must turn every optimization on.

cyrusmsk: Yes, we discussed this previously.
There are 2 approaches. Some frameworks developers would like to test pure speed of their solution (routing, parsing), without any business logic. For others it is more interesting to see how their stack could work in an environment and scenarios, which are close to production. Of course it would mean that some extra work should be added (I/O, json data), not only "hello world" plain text messages. So it is more depends on @waghanza which direction he wants to go :)

waghanza: I think the opposite. To have values on this projects we must use best-practices, it term of security ... avoid bet practices ... the code should be clean ...

d/Dockerfile
trikko: That's fine then. Usually it is really lower (1024 or 512)

PR #7294, opened 2024-03-07
Upgrade serverino from 0.5.1 to 0.6.0

trikko:


cyrusmsk:
Thanks @trikko for new version update
Btw based on your experience and because we know that each request in the benchmark is really small and fast
How do you think should these parameters be improved?
```
        .setHttpTimeout(10.seconds)
        .enableKeepAlive(180.seconds)
        .addListener("0.0.0.0", 3000)
        .setWorkers(totalCPUs);
```

waghanza (approved):

PR #7310, opened 2024-03-12
Serverino 0.6.0 -> 0.6.2

trikko:


waghanza:
Since we have no lock file, is it not the same (I mean with ~ in dub file, is it not the last patch installed)?

trikko:
Yes, it should. Just for clarity.

waghanza (approved):

PR #7346, opened 2024-03-27
Optimize compilation

monocoder:
Local testing - opt: speed option has the effect of improving running speed.

### DIFF
--- a/nim/happyx/config.yaml
+++ b/nim/happyx/config.yaml
@@ -2,5 +2,5 @@ framework:
   github: HapticX/happyx
   version: 3.8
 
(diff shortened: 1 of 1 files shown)


waghanza:
Is it only for happyx ? Suitable for production?

monocoder:
This is the basic command line switch for nim compilation, which should not only be applicable to HappyX.
[https://nim-lang.org/1.2.0/nimc.html#compiler-usage](url)

waghanza:
If it is suitable for production could you apply it for all nim frameworks, please ?

monocoder:
OK，[#7350 ](https://github.com/the-benchmarker/web-frameworks/pull/7350)

waghanza (approved):

PR #7350, opened 2024-03-28
Update Compilation switch

monocoder:
Optimize compilation to improve program efficiency.


waghanza (changes requested):
Unfortunately, this way it be for all except happyx.

You have two files to modify

monocoder:
Do you mean there is no effect after the update?

waghanza (approved):

PR #7354, opened 2024-03-31
[Release] Update for 1st april 2024

waghanza:


cyrusmsk:
Hi @waghanza
Could you please try to run serverino not in Docker?
To compare results? Because it seems it has some issues when running inside docker

PR #7378, opened 2024-04-10
feature: add python robyn framework

### LINKED ISSUE #7368: Add robyn python framework
https://github.com/sparckles/Robyn

jackcvr:
Closes #7368


python/robyn/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  github: sparckles/robyn
+  version: 0.54.0
+
+  engines:
+    - robyn
+
+language:
+  engines:
+    robyn:
+      command: python /usr/src/app/server.py
```
sansyrox, line 12: Hey @jackcvr 👋 

The author of Robyn here. I am not familiar with benchmark config, but you may also want to specify the workers and processes to get the maximum throughput.

`python /usr/src/app/server.py --workers=N --processes=M`

jackcvr: Hey @sansyrox 
Could you suggest optimal amount of N and M?

waghanza: The more accurate config is a value that we can set for a production ready app. We try not to tweak benchmark for performances but display the more realistic performance like in production

sansyrox: Hey @jackcvr @jackcvr , yes sure :D

So, the exact numbers requires some fine-tuning in a real world scenario but a good estimate to start is

N = `os.cpu_count()`
M = `number of cores`

and there is an additional flag that you would run in production mode is `--log-level=WARN` - this is for internal robyn logs

So 
`python /usr/src/app/server.py --workers=N --processes=M --log-level=WARN`

python/robyn/config.yaml
```diff
@@ -0,0 +1,12 @@
...
+  github: sparckles/robyn
+  version: 0.54.0
+
+  engines:
+    - robyn
+
+language:
+  engines:
+    robyn:
+      command: python /usr/src/app/server.py --workers=$(nproc) --processes=$(nproc) --log-level=WARN
```
waghanza, line 12: I think WARN coul mess up the result @sansyrox @jackcvr. Indeed the purpose here is to reflect a production usage, but IO operations could mess up results.

That's why logging is disabled (or should be) on all frameworks here

waghanza (changes requested):

python/robyn/config.yaml
jackcvr: The easiest way to reduce logs could be to set level to CRITICAL. Will it fit?

waghanza (approved):

waghanza:
results updated https://web-frameworks-benchmark.netlify.app/result?l=python @jackcvr @sansyrox

PR #7404, opened 2024-04-20
[D] Updates

cyrusmsk:
Docker image changed from Debian to Ubuntu.
For all solutions LTO was removed
Bumped several versions for compilers

### DIFF
--- a/d/config.yaml
+++ b/d/config.yaml
@@ -1,5 +1,5 @@
 language:
-  version: "1.30"
+  version: "1.37"
 
   files:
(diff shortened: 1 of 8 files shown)


d/config.yaml
```diff
@@ -1,5 +1,5 @@
 language:
-  version: "1.30"
+  version: "1.37"
```
waghanza, line 2: How can we be sure that latest is d 1.37 ? If you specify version on docker file I can make a PR on each slang release @cyrusmsk

PR #7455, opened 2024-05-13
Update `nim/happyx` version

Ethosa:
`v3.8.14` -> `v3.11.0`


waghanza (commented):
Do you need to change something also for the python version?

Ethosa:
> Do you need to change something also for the python version?

nope :)

PR #7462, opened 2024-05-16
feat: blaze bun

krsbx:


waghanza:
Maybe `blaze` folder should be the implementation for `bun` and `blaze-node` should be for nodejs

As I understand that blaze run on top of hono, and hono is bun native.

PS : The same logic should be used for hono directory

javascript/blaze-bun/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  github: Busy-Hour-Studio/blaze
+  version: 3.1
```
waghanza, line 3:
```suggestion
  version: 3.2
```

waghanza (approved):

waghanza (approved):

PR #7471, opened 2024-05-21
RESTHeart 8.0

ujibang:
we just released RESTHeart 8.0.

note: RESTHeart runs on Java 21. I'm not sure why it is listed as `java (11)` in the results (like any other java framework).


waghanza:
> why it is listed as `java (11)` in the results (like any other java framework).

It is a bug open our end

java/restheart/pom.xml
```diff
@@ -9,7 +9,7 @@
   <properties>
     <maven.compiler.source>21</maven.compiler.source>
     <maven.compiler.target>21</maven.compiler.target>
-    <restheart.version>8.0.0-beta</restheart.version>
+    <restheart.version>[8.0,8.1.0-SNAPSHOT)</restheart.version>
```
waghanza, line 12: I think this is to guarantee to keep in 8.0.x

Side note :
Is there a way to use `[8.0,8.1)` and to exclude all releases containing SNAPSHOT string in its name. I know that some artefact on java world are named 4.1-RC but are 4.0 and not 4.1

java/restheart/pom.xml
```diff
@@ -18,6 +18,21 @@
         <artifactId>restheart</artifactId>
         <version>${restheart.version}</version>
       </dependency>
+
+      <!-- avoids including following artifacts in shaded jar-->
+      <dependency>
+        <groupId>org.graalvm.polyglot</groupId>
```
waghanza, line 24: Not sure why graalvm here ?

The Dockerfile specify temurin.

PS : I plan to make all java framework run on temurin and graalvm to check also memory usage

java/restheart/pom.xml
ujibang: with `<restheart.version>[8.0,8.1)</restheart.version>` the version `8.1.0-SNAPSHOT` would be picked up. mvn is quite wierd with range version and this is the only way I found to exclude  8.1.0-SNAPSHOT

java/restheart/pom.xml
ujibang: setting those transitive dependencies of restheart will avoid including them in the shaded jar. Why this is happening is not yet clear, but we are fixing it in our pom.xml. latest version still requires this (however it would just produce some warnings by the maven shade plugin)

btw those dependencies are used by restheart to execute JavaScript plugins (when running on GraalVM). See https://restheart.org/docs/plugins/core-plugins-js 

For this release, better to exclude them defining their scope as `provided`

java/restheart/pom.xml
waghanza: Ok then, let's use this.

But you prefer to use gradle, be my guest. The package manager depends on the framework choice (I mean I prefer to use what developers of framework recommend)

java/restheart/pom.xml
ujibang: Executing the tests on both Temurin and GraalVM would be absolutely fantastic 🚀! I would love to see this implemented!!

ujibang: for this transitive dependencies, we have now an issue. hope to fix it soon https://github.com/SoftInstigate/restheart/issues/510

ujibang: btw I found a better way to exclude transitive deps...

waghanza (approved):

PR #7494, opened 2024-05-30
Vibed update

cyrusmsk:
Updated version of the library and add some configuration changes.


d/vibed/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   website: vibed.org
-  version: 0.9
+  version: 0.10
```
waghanza, line 3:
```suggestion
  version: '0.10'
```

because of parsing

otherwise it will be displayed 0.1

d/vibed/dub.json
```diff
@@ -1,13 +1,19 @@
...
-    "vibe-d:http": "~>0.9.8"
+    "vibe-d:http": "~>0.10.0",
+    "vibe-d:tls": "*"
   },
   "versions": [
     "VibeManualMemoryManagement",
     "VibeHighEventPriority",
     "VibeDisableCommandLineParsing"
   ],
-  "dflags-ldc": ["-mcpu=native"],
```
waghanza, line 11: Maybe we should have the same flag everywhere. I mean if removed here, it is worth removing it for others

waghanza (changes requested):

d/vibed/dub.json
cyrusmsk: It wasn't removed. It was just below) I put it back. It should stay for everyone.

d/vibed/config.yaml
cyrusmsk: Fixed

waghanza (approved):

PR #7510, opened 2024-06-10
lua: fix turbo framework

jackcvr:


lua/turbo/config.yaml
```diff
@@ -3,4 +3,5 @@ framework:
   version: "2.1.3"
 
   build:
-    - apk add --no-cache lua-turbo
+    - apk add --no-cache luajit lua5.4-dev luarocks5.4 git build-base libressl-dev
+    - luarocks-5.4 install turbo
```
waghanza, line 7: Could you use luarocks to install it from a file ? I mean having a file like Gemfile, requirements.txt, Cargo.toml ...

jackcvr: Unfortunately I don't know lua that much. Let's handle this in separate task.

waghanza (approved):

PR #7554, opened 2024-06-23
[D] Added Hunt and updated versions for serverino and vibe worker pool

cyrusmsk:
This PR is adding new framework for D - Hunt http server.
I've also updated the worker pool size for serverino and vibe. Because larger pool sizes work better for me locally (but I've tested on macOS and M1). Hopefully it will improve the performance in Linux/Docker as well.

### DIFF
--- a/d/vibed/dub.json
+++ b/d/vibed/dub.json
@@ -1,8 +1,8 @@
 {
   "name": "server",
   "dependencies": {
-    "vibe-d:http": "~>0.10.0",
-    "vibe-d:tls": "*"
+      "vibe-http": "~>1.1.0",
+      "vibe-stream:tls": "~>1.1.0"
   },
   "versions": [
     "VibeManualMemoryManagement",
@@ -11,7 +11,7 @@
   ],
   "dflags-ldc": ["-mcpu=native"],
   "subConfigurations": {
-    "vibe-d:tls": "notls"
+    "vibe-stream:tls": "notls"
   },
   "license": "MIT"
 }
--- a/d/hunt/.gitignore
+++ b/d/hunt/.gitignore
@@ -1,0 +1,3 @@
+.dub/
+server
+dub.selections.json
--- a/d/hunt/config.yaml
+++ b/d/hunt/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  github: huntlabs/hunt-http
+  version: 0.8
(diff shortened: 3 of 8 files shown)


waghanza:
@cyrusmsk could you ping accurate people to see if no opposition to introduce hunt here ?

waghanza (changes requested):
Missing config for vibe

cyrusmsk:
> Missing config for vibe

not sure what do you mean?
Vibe folder has config.yaml

cyrusmsk:
Hey @zoujiaqing do you mind to add Hunt http to this benchmark?

waghanza:
> > Missing config for vibe
> 
> not sure what do you mean?
> Vibe folder has config.yaml

Vibe version should not be changed to 1.1 ?

cyrusmsk:
> > > Missing config for vibe
> > 
> > 
> > not sure what do you mean?
> > Vibe folder has config.yaml
> 
> Vibe version should not be changed to 1.1 ?

I don't think so. Because I've used separate package "vibe-http", but it seems it was just separated for convenience. As full framework "vibe" still has 0.10
And it has "vibe-d:http" sub-package, which should be the same (at least for now).
Generally speaking vibe-http - is an http server part of the vibe-d web-framework.
And based on the release notes from the latest major update https://vibed.org/blog/posts/vibe-release-0.10.0 it seems it is recommended way of using the server part.

waghanza:
Ok, understood. Perhaps we need to refacto `dub`manifest then

d/vibed/dub.json
```diff
@@ -1,8 +1,8 @@
 {
   "name": "server",
   "dependencies": {
-    "vibe-d:http": "~>0.10.0",
-    "vibe-d:tls": "*"
+      "vibe-http": "~>1.1.0",
+      "vibe-stream:tls": "~>1.1.0"
```
waghanza, line 5: I think this is enough, no @cyrusmsk 

```suggestion
      "vibe-d": "~>0.10.0"
```

cyrusmsk: No, because by default it is using tls. So to switch off the TLS you need to specify it explicitly. vibe-d has vibe-http in dependency.

waghanza:
@zoujiaqing I'll merge hunt so as to include it here. 

Please ping me if you prefer not to

Issue #7571, opened 2024-07-02
[Go] Add `net/http` package’s router


xuyang2:
https://go.dev/blog/routing-enhancements

PR #7572, opened 2024-07-02
Add go1.22 "net/http" package’s router

xuyang2:
#7571


waghanza (commented):
Good idea. Ths could be a reference for go frameworks that use net/http 🎉

waghanza (approved):

PR #7589, opened 2024-07-08
Upgrade RESTHeart to v8.0.4

ujibang:


java/restheart/pom.xml
```diff
@@ -7,9 +7,8 @@
   <packaging>jar</packaging>
 
   <properties>
-    <maven.compiler.source>21</maven.compiler.source>
-    <maven.compiler.target>21</maven.compiler.target>
-    <restheart.version>[8.0,8.1.0-SNAPSHOT)</restheart.version>
+    <maven.compiler.release>21</maven.compiler.release>
+    <restheart.version>8.0.4</restheart.version>
```
waghanza, line 11: Rather 

```suggestion
    <restheart.version>[8.0,8.1)</restheart.version>
```

waghanza (changes requested):

java/restheart/pom.xml
ujibang: unfortunately your suggested change will pick up `8.1.0-SNAPSHOT`

java/restheart/pom.xml
```diff
@@ -7,8 +7,7 @@
   <packaging>jar</packaging>
 
   <properties>
-    <maven.compiler.source>21</maven.compiler.source>
-    <maven.compiler.target>21</maven.compiler.target>
+    <maven.compiler.release>21</maven.compiler.release>
```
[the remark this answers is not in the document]
ujibang: You can create a parent POM that defines common options (see [introduction to the POM](https://maven.apache.org/guides/introduction/introduction-to-the-pom.html)). 

However, since different frameworks have very distinct dependencies and options, I suspect it may not be very useful.

waghanza (approved):

PR #7595, opened 2024-07-10
[Javascript] Solving issue in Deno

mimiMonads:


waghanza (changes requested):
Merge conflict?

mimiMonads:
Sorry, I was at work and I just wanted to solve the problem quickly, what happened is that I passed this project to `jsr` and my IDE interpreted `app` as `any`, that's why I couldn't see the error at first, but It should be working.

And since hono and vixeny are very similar, I didn't run `deno` , let's hope everything turns out well. I want to see if I can pass axum.

waghanza (approved):

PR #7605, opened 2024-07-13
Add Kotlin "Spring"

mmaryo:


kotlin/spring/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
+plugins {
+	id("org.springframework.boot") version "3.3.1"
+	id("io.spring.dependency-management") version "1.1.5"
+	kotlin("jvm") version "1.9.24"
+	kotlin("plugin.spring") version "1.9.24"
```
waghanza, line 5:
```suggestion
	id("org.springframework.boot") version "3.3.+"
	id("io.spring.dependency-management") version "+"
	kotlin("jvm") version "1.9.+"
	kotlin("plugin.spring") version "1.9.+"
```

to make sure we use the patched version of each

kotlin/spring/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+
+dependencies {
+	implementation("org.springframework.boot:spring-boot-starter-web")
+	implementation("com.fasterxml.jackson.module:jackson-module-kotlin")
+	implementation("org.jetbrains.kotlin:kotlin-reflect")
+}
+
+kotlin {
+	compilerOptions {
+		freeCompilerArgs.addAll("-Xjsr305=strict")
```
waghanza, line 29: is it something specific to kotlin / spring ? I mean this seems to be a jvm option, and then makes the results unfaire between jvm based frameworks

kotlin/spring/build.gradle.kts
```diff
@@ -0,0 +1,39 @@
...
+	}
+}
+
+tasks.withType<Test> {
```
waghanza, line 33: required ?

kotlin/spring/gradle/wrapper/gradle-wrapper.jar
waghanza, line 1: could this file be in .gitignore, no need to version it, imho

kotlin/spring/gradle/wrapper/gradle-wrapper.properties
waghanza, line 1: could this file be in .gitignore, no need to version it, imho

waghanza (changes requested):

kotlin/spring/build.gradle.kts
mmaryo: it's about null safty and by default with spring starter, but not useful in this case https://docs.spring.io/spring-boot/docs/2.0.x/reference/html/boot-features-kotlin.html#boot-features-kotlin-null-safety

kotlin/spring/src/main/resources/application.properties
```diff
@@ -0,0 +1,4 @@
+server.port=3000
+server.error.whitelabel.enabled=false
```
waghanza, line 2: What is this for ?

waghanza (commented):
Quick question

kotlin/spring/src/main/resources/application.properties
waghanza: https://www.baeldung.com/spring-boot-custom-error-page

Maybe the accurate to the same for spring/spring webflux in java

waghanza (commented):
Maybe replicate the same behaviour

waghanza (approved):

kotlin/spring/src/main/resources/application.properties
mmaryo: same as spring java

PR #7633, opened 2024-07-18
feat(elysia): using static resource

SaltyAom:
This PR modifies Elysia code to use static resources by passing a value instead of a callback as recommended in [Elysia handler](https://elysiajs.com/essential/handler.html).

This allows Elysia to know data shape ahead of time allowing Elysia to optimize more performance.

Several plugins use this method eg, static plugins.

A static resource is not a cache as documented in [handler](https://elysiajs.com/essential/handler.html), as it can be mutated for each request or ignored/overridden by other lifecycle events.

### DIFF
--- a/javascript/elysia/app.ts
+++ b/javascript/elysia/app.ts
@@ -1,7 +1,7 @@
 import { Elysia } from "elysia";
 
 new Elysia()
-  .get("/", () => "")
+  .get("/", "")
   .get("/user/:id", (req) => req.params.id)
-  .post("/user", () => "")
+  .post("/user", "")
   .listen(3000);


waghanza (approved):

kravetsone:
> This PR modifies Elysia code to use static resources by passing a value instead of a callback as recommended in [Elysia handler](https://elysiajs.com/essential/handler.html).
> 
> This allows Elysia to know data shape ahead of time allowing Elysia to optimize more performance.
> 
> Several plugins use this method eg, static plugins.
> 
> A static resource is not a cache as documented in [handler](https://elysiajs.com/essential/handler.html), as it can be mutated for each request or ignored/overridden by other lifecycle events.

Can sucrose handle this and optimize it too?

PR #7650, opened 2024-07-27
[Go] Update to 1.23

akyoto:
For the future, this should be set to `GOAMD64=local` once Go adds support for it.
For now, `v3` is supported on many VPS / cloud servers while `v4` is still somewhat rare.
I don't know what the GitHub CI is using, if it complains we might need to downgrade to `v2` or `v1`.
There's also [GOARM64](https://go-review.googlesource.com/c/go/+/591875/5/doc/next/7-ports.md#23) in Go 1.23 in case the benchmarking machine is run on ARM64.


waghanza:
Thanks fot the tips @akyoto . 

Actually, with an M1, I've set 
```dockerfile
ENV CGO_ENABLED=0
ENV GOOS=linux
ENV GOARCH=arm64
``` 
 
since we are on 1.22, we can not change for now

it's easily changeable in the file `Rakefile` when 1.23 will be out

For the second part, `GOAMD64` could you point me to some doc, I am not sure about how go react for that

If you want I can ping you when 1.23 will be out on docker hub (e.g : not in rc)

akyoto:
> For the second part, `GOAMD64` could you point me to some doc, I am not sure about how go react for that

https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels

Essentially there are these microarchitecture levels, v1-v4, that specify which instructions are available. In the case of Go, you can declare that you don't want runtime CPU checks and instead bake the fast path at compile time into the executable. This makes the binary faster (because it avoids runtime checks) at the price of compatibility with older CPUs. However, when you're picking your framework to build your own web service, you're in charge of what machine it will be deployed on. Since both my development machine and my cloud servers support `v3`, that's what I'm using. In general, everyone should be using the highest generation possible for deployments.

The only time you should avoid setting this variable is if you share the binary executables with the public (because you have no control over what machine it will run on).

waghanza:
I see. 

Then, this is not useful here :-) at least for now.

Results are produced on ARM64 (M1), but this could change.

Even if this could not affect our results, it is interesting to have this explicitly set on docker files.

The idea of this project is also to include best-practices of each communities, so good to include this in here

waghanza:
We should set also `GOARM64` to `v8.5` if i understand correctly @akyoto 



https://en.wikipedia.org/wiki/Apple_M1?wprov=sfla1

akyoto:
> We should set also `GOARM64` to `v8.5` if i understand correctly @akyoto
> 
> https://en.wikipedia.org/wiki/Apple_M1?wprov=sfla1

I don't have an Apple M1 to test with, but the simplest way to find out the version on your machine is to just incrementally up the version number until the executable complains:

```
This program can only be run on AMD64 processors with v4 microarchitecture support.
```

^ I get this error message on `GOAMD64=v4 go build` and executing the binary afterwards.

From a quick web search it seems you're correct, v8.5 for the M1 should be the limit.

---

Are you - realistically speaking - the only one who actually runs all the benchmarks? If you are, then I think it's okay to hardcode the values for your machines.

Otherwise, if there are more people running these, we'd have to wait for the `local` value to be supported by Go or at least add an explanation that people with older CPUs need to downgrade these environment variables if the Go binaries aren't starting on their machines.

waghanza:
I'm the only person to run the _official_ ones, but anybody could run that benchmark.

This value could be dynamicly set by ruby, I'll find a way, but ... maye when 1.23 will be out.

Let's work together in this PR to craft anything for this

waghanza:
When does go will support `GOAMD64=local` @akyoto ?

akyoto:
> When does go will support `GOAMD64=local` @akyoto ?

https://github.com/golang/go/issues/60338

The proposal was accepted in June, 2023.
No implementation yet, as far as I can tell.

waghanza:
I was thinking of templetazing `GOARM64` and `GOAMD64` values, so as it is explicit. For example, in my machine (a ryzen), I have a dockerfile like this
```dockerfile
FROM golang:1.23-rc

WORKDIR /go/src/app



ENV CGO_ENABLED=0
ENV GOOS=linux
ENV GOARCH=x86_64

  # This compile with v3 micoarchitecture
  # See https://en.wikipedia.org/wiki/X86-64#Microarchitecture_levels 
  ENV GOAMD64=v3


  COPY 'main.go' 'main.go'
  COPY 'go.mod' 'go.mod'


RUN go get
RUN go build -a -ldflags '-extldflags "-static"'  -o /go/bin/app ./


FROM debian:stable

WORKDIR /go/bin


COPY --from=0 /go/bin/app /go/bin/app


CMD /go/bin/app
````

on the official rounds, it should have `ENV GOARM64=v8.5`

**Comment by waghanza** [comment]
@Duslia `hertz` seems not to compile in **1.23rc2**. any plan to support this version when it will be GA ?

**Comment by akyoto** [comment]
I'm perfectly fine with excluding the variables in the repo code, and instead setting them only when the official benchmark is run. This seems like a reasonable solution for the time being.

File: go/Dockerfile [hunk 1]
```diff
@@ -1,8 +1,9 @@
 {{#language.version}}
-FROM golang:{{{.}}}
+  FROM golang:{{{.}}}
 {{/language.version}}
+
 {{^language.version}}
-FROM golang:1.22
+  FROM golang:1.23rc2
```
waghanza, line 6:
```suggestion
  FROM golang:1.23
```

waghanza (changes requested):

waghanza (approved):

PR #7658, opened 2024-07-30
[OCaml] Add dream

pyx:
Dependencies:

  - opam 2.x
  - OCaml 5.2.0 or newer
  - dune 2.0+
  - dream 1.0.0~alpha6

Only `opam` is needed from system, everything else can be fetched from official repositories and installed by `opam` inside a local switch with `make deps` and follow the instructions. `make serve` to run.

The dream server, by default, runs in a single thread (via Lwt), no optimization applied in code, I even deliberately convert the id parameter into integer and back to string for output, to mimic many other frameworks with auto type-conversion.


pyx:
Sorry that I didn't provide a docker file, since I never use docker, I wrote a Makefile with commands to show how to setup and run instead.

pyx:
Replaced this function
```ocaml
let user_info request =
  match Dream.param request "id" |> int_of_string_opt with
  | None -> Dream.empty `Not_Found
  | Some id -> id |> string_of_int |> Dream.html
```
with this
```ocaml
let user_info request = Dream.param request "id" |> Dream.html
```
If integer `id` (can be string) is not a hard requirement.

waghanza:
@aantron we would like to add dram in this benchmarking project. is this OK with you ?
@pyx I'll push in this PR to create a dockerfile, and also a config.yaml which describe the framework

pyx:
> @aantron we would like to add dram in this benchmarking project. is this OK with you ? @pyx I'll push in this PR to create a dockerfile, and also a config.yaml which describe the framework

@waghanza I cannot represent @aantron (the author of the dream web framework), but IMHO, more exposure is always good for a project, especially I believe dream will perform well in the benchmark, which is the case on my local machine.

waghanza:
> more exposure is always good for a project, especially I believe dream will perform well in the benchmark, which is the case on my local machine.

I'm not totally align with that, sometimes when a toot is not stable is could be counter productive 

Anyway, I have to ask each time 😁

Either I can be mistaken or the author can refuse (even we are in OSS), it happened a few times in the past

waghanza:
I'll try to include dream on next update (Monday) unless it's author is opposed to

ocaml/dream/server.opam
```diff
@@ -0,0 +1,7 @@
+opam-version: "2.0"
+
+depends: [
+  "ocaml" {>= "5.2.0"}
+  "dream"
```
waghanza, line 5: Could you specify a version q@pyx ?

ocaml/dream/Makefile
```diff
@@ -0,0 +1,8 @@
+.PHONY: deps
+deps:
+	opam switch create . 5.2.0 --yes
+	opam install --deps-only --yes .
+
+.PHONY: serve
+serve:
+	dune exec --root . ./server.exe
```
waghanza, line 8: Is this production ready?

aantron:
I very much approve, thank you!

Perhaps I would do a PR afterwards to twiddle some of the details, since this code is public-facing, if that would be welcome.

waghanza:
Very much appreciated @aantron. Perhaps you may make a PR that target this ❤️

waghanza:
I made https://github.com/pyx/web-frameworks/pull/1 to enable ocaml here @pyx

ocaml/dream/Makefile
pyx: Afaik, it is production ready in the sense that the server comes with dream is not designed as a development server, and I use dream in production, as well as the official website of [OCaml](https://ocaml.org/), I am not affiliated with OCaml, but found some public-facing code using dream, as in here https://github.com/ocaml/ocaml.org/blob/b34183bcc1e4966390e2240ff920fc2d64cdde09/src/ocamlorg_web/lib/ocamlorg_web.ml#L15

ocaml/dream/server.opam
pyx: Done, I pinned OCaml to 5.2.0 and dream to 1.0.0~alpha6, both are official (sort of in dream's case) stable releases.

ocaml/dream/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: https://aantron.github.io/dream/
```
waghanza, line 2:
```suggestion
  website: aantron.github.io/dream/
```

waghanza (approved):

waghanza:
Could you please also set port to 3000 @pyx ?

pyx:
> Could you please also set port to 3000 @pyx ?

Done.

pyx:
> Could you please also set port to 3000 @pyx ?

I think we can do better, e.g, in README or something, make the server spec. obvious (I have to figured it out by reading some python and haskell code): 
 - serve on 0.0.0.0:3000
 - serve three end point (or routes, or handlers, or whatever the framework call it)
     - method GET on 0.0.0.0:3000/ to return empty body
     - method GET on 0.0.0.0:3000/user/id in which 'id' is parsed and return as string, and does it have to be parsed as an integer?
     - method POST on 0.0.0.0:3000/user to return empty body
    
  My two cents.

waghanza:
> > Could you please also set port to 3000 @pyx ?
> 
> I think we can do better, e.g, in README or something, make the server spec. obvious (I have to figured it out by reading some python and haskell code):
> 
>     * serve on 0.0.0.0:3000
> 
>     * serve three end point (or routes, or handlers, or whatever the framework calls it)
>       
>       * method GET on 0.0.0.0:3000/ to return empty body
>       * method GET on 0.0.0.0:3000/user/id in which 'id' is parsed and return as string, and does it have to be parsed as an integer?
>       * method POST on 0.0.0.0:3000/user to return empty body
> 
> 
> My two cents.

You write, I'l update the README, since I'll have bandwith it the near future

At least dream is workin https://github.com/the-benchmarker/web-frameworks/actions/runs/10175734496/job/28143750108?pr=7658#step:9:9

waghanza (approved):

pyx:
> > > Could you please also set port to 3000 @pyx ?
> > 
> > 
> > I think we can do better, e.g, in README or something, make the server spec. obvious (I have to figured it out by reading some python and haskell code):
> > ```
> > * serve on 0.0.0.0:3000
> > 
> > * serve three end point (or routes, or handlers, or whatever the framework calls it)
> >   
> >   * method GET on 0.0.0.0:3000/ to return empty body
> >   * method GET on 0.0.0.0:3000/user/id in which 'id' is parsed and return as string, and does it have to be parsed as an integer?
> >   * method POST on 0.0.0.0:3000/user to return empty body
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > My two cents.
> 
> You write, I'l update the README, since I'll have bandwith it the near future
> 
> At least dream is workin https://github.com/the-benchmarker/web-frameworks/actions/runs/10175734496/job/28143750108?pr=7658#step:9:9

My fault, when I read more files in the project, I found out that **you did a great job** in [CONTRIBUTING.md](https://github.com/the-benchmarker/web-frameworks/blob/master/CONTRIBUTING.mdl), maybe adding a line in `README` stating the requirements are in `CONTRIBUTING.md` is enough? Then add a line in `CONTRIBUTING.md` mentioning the test harness will be visiting from outside of LAN and at port 3000, please make the server ready for that or provide a way to configure as such.

Also when I read more code of other languages and frameworks, I figured the id part is treated as string, so I will submit a PR for that.

Thank you.

PR #7696, opened 2024-08-13
Add CodeBehind Framework

elanatframework:


waghanza (changes requested):
CI is failing.   `/user/:id` should return `:id` as plain text

elanatframework:
The CodeBehind Framework supports two types of configuration:
1- Default configuration
2- Configuring the controller in Route

In the default configuration, the name of the Controller class is specified in the View; requests go to the View first, and the View creates a new instance of the Controller class.

Here we first tried to create the default configuration, but in your test build, the contents of the wwwroot directory were being ignored. We were only able to implement the configuration of the controller in Route.

waghanza:
Thanks for clarification about CodeBehind configuration. It's up to you to choose what suits bests.

The idea is to have a production-ready implementation with the latest **LTS**. I mean the idea is that an app could be bootstraped with this code base, so the code should be flexible, secure ....

Probably a one-file implementation is not the best 😛

waghanza (approved):

waghanza:
Results has been updated https://web-frameworks-benchmark.netlify.app/result

elanatframework:
thank you

These tests help us to increase the performance of the CodeBehind framework.
We make changes to the structure of the CodeBehind framework and update the version of the application in your repository.

waghanza:
Sure, feel free. I try my best to  run update each weeks

PR #7727, opened 2024-09-04
chore(rails): update the number of puma threads to 3

nakrovati:
The default value of puma threads has been [updated](https://edgeguides.rubyonrails.org/7_2_release_notes.html#set-a-new-default-for-the-puma-thread-count) in rails 7.2, now there are 3 of them.


waghanza (approved):
Thanks for your contribution :tada:

PR #7730, opened 2024-09-05
Enable YJIT for Ruby

p8:
Rails 7.2 enables YJIT itself after initialization.
This makes sure code that only runs on initializing is not jitted.
So for Rails we set RUBY_YJIT_ENABLE to falsey.

### DIFF
--- a/ruby/config.yaml
+++ b/ruby/config.yaml
@@ -33,6 +33,7 @@ framework:
     - config.ru
 
   environment:
+    RUBY_YJIT_ENABLE: 1
     RACK_ENV: production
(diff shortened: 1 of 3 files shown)


waghanza (commented):
Maybe this could be set in `ruby/config.yaml`

p8:
I've enabled it for all ruby apps, and made an exception for Rails, which sets it itself for improved performance.

p8:
@waghanza  I've updated `ruby/config.yaml`.

waghanza (approved):

waghanza:
https://web-frameworks-benchmark.netlify.app/result 🎉

Issue #7732, opened 2024-09-07
Create new types of tests


mmaryo:
Hello, 

Currently, the web-framework benchmark tests an empty endpoint.

However, web frameworks are about more than just request delay; they are mainly about processing complex tasks like JSON serialization, object composition, DB requests, etc.

Have you ever considered enriching the benchmark with new types of tests? It could provide a more comprehensive evaluation of web frameworks.

Imagine the following endpoints:
/ - the current behavior, we check the response time
/external-service - we call external services (a function that does a sleep) 
/serialization - we send and receive a large JSON body in POST request with few query parameters  
/http-client - we test the standard HTTP client and load a large file hosted in the server itself
/web - we do complete web scenarios with JSON in requests, DB requests, data transformations, and large JSON in response

We also deprecated the endpoints : 
/user/{id}
/user
Because they do not bring value compared to the / endpoint

The current benchmark does not need to be updated. But each new piece of code that is not a framework update must implement that logic.

waghanza:
Hey,

Thanks for this proposal. 

Seems to go in a direction that I'm confortable with

I suggest slight enhancement.

+ /user/0 -> instead of / 
  + the idea is to test routing condition
  + and the response
+ the database used for the /web endpoint should be **in-memory** and **local** and the same across all framework (standard like sqlite)
   + better to avoid network latency and else that could mess the results
   + better to avoid persistency that could mess the reuslts

waghanza:
Not sure to get you @cyrusmsk

cyrusmsk:
As I understood you want to introduce new endpoints for testing? With sleep with json maybe with SQLite.
I’m wondering about more detailed description and the timeline of incorporation of these new tests

waghanza:
The idea is to make new tests, so as this project could be meaningul. But we plan to do this incrementally, no timeline defined at this time

cyrusmsk:
> The idea is to make new tests, so as this project could be meaningul. But we plan to do this incrementally, no timeline defined at this time

In that case maybe new filter field will be helpful. Something like "Old tests" - "New tests". So in the result will be possible to see different implementations without any breaking changes.

waghanza:
You mean add this on UI

If you make somes PRs why not ... Just a joke

I'm more on the line of having a new UI when all tests are ready, and in the meantime we can just make the backend code step by step

waghanza:
We are going to write new test (add some endpoints)

+ /, check the routing part of a framework
[...]

PR #7737, opened 2024-09-08
Adding new tests

mmaryo:


waghanza:
> I don't think your PR is good! First of all, there is no thread about planning and discussing new test proposals before.

The idea is to based a discussion on this 😛 

> Secondly, it is very irresponsible to implement Spring under Java category.

Why ?

> You should turn your PR into a draft and update other frameworks to it within your ability.

The idea is to add some endpoint for a **v2** on results, I mean the whole idea is to avoid being just a hello world

> Your current approach is very irresponsible, and you transfer a huge workload to others.

I do not understand it @medz . Could you explain a bit ?

> From your PR, you are very familiar with Java, and you should at least update other frameworks under Java in a unified way.

Your idea is good, but the first we should have tests to make sure this works, and after make sure we can uniformize java implementations (this is planed as so)

medz:
https://github.com/the-benchmarker/web-frameworks/issues/7732

I just noticed this discussion, so the above statement is wrong.

PR #7751, opened 2024-09-11
add durian.js

mario-huang:
durian.js builds on the strengths of NestJS, running on [Bun](https://github.com/oven-sh/bun), and is designed with speed in mind — both in development and runtime performance.

- Development Speed: durian.js can be up to 16 times faster than NestJS in development.
- Runtime Performance and Latency: durian.js achieves up to 4 times the throughput of NestJS, handling over 60,000 requests per second compared to NestJS’s 15,000, while also reducing average response times from 0.0033 seconds to 0.0008 seconds.


javascript/durian.js/config.yaml
```diff
@@ -0,0 +1,8 @@
+framework:
+  github: mario-huang/durian.js
+  version: 1.0.0
```
waghanza, line 3:
```suggestion
  version: 1.0
```

waghanza (changes requested):

waghanza (approved):

PR #7759, opened 2024-09-15
[ruby] Use Jemalloc for better memory performance

p8:
Using jemalloc typically reduces memory usage of a Ruby application:
https://www.speedshop.co/2017/12/04/malloc-doubles-ruby-memory.html

### DIFF
--- a/ruby/Dockerfile
+++ b/ruby/Dockerfile
@@ -6,6 +6,9 @@ RUN apt-get -qq update
 
 RUN apt-get -qy install --no-install-recommends build-essential
 
+RUN apt-get install -y --no-install-recommends libjemalloc2
+ENV LD_PRELOAD=libjemalloc.so.2
+
 {{#files}}
   COPY '{{source}}' '{{target}}'
 {{/files}}


waghanza:
Thanks @p8 for this clear update.

However, I have a question : is this **production ready** ? I mean I understand that this could be a trics to have more performance from a rubyist perspective, but this seems more like plumbing for me.

I am not aware of any production app behing like that. Am I right ? Perhaps not.

Feel free to give some insight to understand if this is a "best-practice" in ruby communities.

Regards,

PS : This project is more about presenting what a real-world app can do, and then this not about tweaking any piece of code (either ruby or else) to perform better

p8:
@waghanza

There is a [long thread](url) (with a lot of prominent Rubist) on the Ruby mailinglist to enable it by default.
It was eventually rejected because of cross-platform compatibility.
It's pretty common the enable it:
1. The "dev.to" team enabled it:
    https://dev.to/devteam/how-we-decreased-our-memory-usage-with-jemalloc-4d5n
2. It's enabled for discourse:
    https://github.com/discourse/discourse/blob/edd221fbf92ce8e1a2f5ba4193b4479c3ce14e02/bin/docker/exec#L8
3. It also is enabled by default in Rails 7.2 Dockerfile:
    https://edgeguides.rubyonrails.org/7_2_release_notes.html#setup-jemalloc-in-default-dockerfile-to-optimize-memory-allocation

waghanza:
Thinks I see is that this is the default for rails 7.2 new apps, and used in some widely spreded apps. 

As I read the only caveats is that this is not compatible with windows, but I'm not aware of any ruby based app using windows.

As far as I'm concerned, it could be enabled

waghanza (approved):

PR #7781, opened 2024-09-17
add durian.js config to use multi cores

mario-huang:


waghanza (commented):
Nice, thanks for this.

Could you then remove cluster.ts from all frameworks with bun then ?

mario-huang:
> Nice, thanks for this.
> 
> Could you then remove cluster.ts from all frameworks with bun then ?

OK, I will try.

mario-huang:
@waghanza It seems that bunicorn has bug in itself, what should I do.

javascript/0http-bun/config.yaml
```diff
@@ -2,4 +2,4 @@ framework:
   github: BackendStack21/0http-bun
   version: 1.0
   engines:
-    - bun
+    - durian.js
```
waghanza, line 5: I do not understand this, the engine is still bun

javascript/config.yaml
```diff
@@ -42,6 +42,13 @@ language:
       bootstrap:
         - npm install
       command: node app.js
+    durian.js:
```
waghanza, line 45: You can drop the old config for bun

waghanza (changes requested):

javascript/config.yaml
mario-huang: OK, got it.

mario-huang:
When I test, I found there is still bug with pm2 and bun. So no rush, I will keep eyes on it.
https://github.com/Unitech/pm2/issues/5893

waghanza (approved):

waghanza:
Thanks, I'll merge and fix it

mario-huang:
@waghanza Hi.
I think perhaps it's better not to merge for now. During my search over the past few days, I found that PM2 still doesn't support Bun running in cluster mode at the moment. If it's run in fork mode, the program will throw an error due to port conflicts.
This made me consider another issue: as a benchmark, wouldn't it be better not to introduce additional load balancing like pm2? After all, different load balancing implementations can also affect the results. If we let all frameworks run on a single CPU core, we can maintain the simplicity of the frameworks and better reflect the performance of the frameworks themselves.

waghanza:
Understood @mario-huang. We actually only display one engine, per frameworks, I'll be too heavy to maintain to have multiple load balancesrs. Seems like pm2 is the most used in javascript world, but what load balance do you think about ?

mario-huang:
Yes, pm2 is the most used in javascript world, it let nodejs apps use multi CPU cores. 
With the widespread adoption of Kubernetes (k8s), PM2 is no longer essential. We can directly utilize k8s' load balancing features to achieve the same results. Alright, back to the JavaScript situation in this project. Currently, we have Node.js, Deno, and Bun. Both Node.js and Deno use PM2 to take advantage of multi-core CPUs and share port 3000. The Bun project directly runs Bun.serve({reusePort: true}), achieving the same goal. However, Durian.js does not run based on Bun.serve(), which prevents it from having multiple instances and sharing ports like other Bun projects. I tried using PM2 to solve this problem, but unfortunately, it doesn't seem feasible at the moment. So, it looks like I'll need to wait for PM2 to support Bun, or add cpu: "1.0" in all the Dockerfiles to ensure that all frameworks use only one core. This way, we can compare performance on a single core, and none of the frameworks would need any load balancing.

waghanza:
Ok. Understood. So I'll restore the old situation, and let you make a PR when this feature will be available on pm2, it is not doable to use k8s here with multiple nodes, at least with only one machine because this runs on my own infrastructure (not cloud).

Thanks for your work on this btw :heart:

PR #7807, opened 2024-09-24
try to use cluster in bun

mario-huang:


mario-huang:
I find a new way to use cluster without pm2, let's try it.

PR #7809, opened 2024-09-24
Refactor controller to use functional routing

nathphon:
Replaces the annotation-based controller with a functional routing approach. Introduces a `RouterFunction` bean to handle HTTP requests, eliminates redundant response creation, and improves efficiency by directly parsing path variables.

### DIFF
--- a/java/spring-webflux/src/main/java/framework/benchmark/controller/BenchmarkController.java
+++ b/java/spring-webflux/src/main/java/framework/benchmark/controller/BenchmarkController.java
@@ -1,4 +1,4 @@
-package framework.benchmark.router;
+package framework.benchmark.controller;
 
 import org.springframework.context.annotation.Bean;
 import org.springframework.http.MediaType;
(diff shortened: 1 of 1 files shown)


waghanza:
Thanks @nathphon for this suggestion.

What is the best approach for a server side app (typically an API) ? This implementation, or this old one ?

Please, give me some insights, not in java world 😀

nathphon:
in very high-throughput systems, the functional style (RouterFunction) may have a slight edge due to its more lightweight design, bypassing some of the overhead involved in handling annotations.

and i re write code to be the same style with ActiveJ

waghanza:
cc @mmaryo

whiplash:
It all makes sense to me. It appears that using Spring these days you have a choice between going MVC of WebFlux. WebFlux seems to have pushed more on the functional pedal (which I tend to prefer). So It makes sense to have both.
As far as I can tell, the code here is the idiomatic way using WebFlux.

waghanza:
Ok, so using webfux, like functional way seems to be the way to go.

But  the same impl is no the way to go with the MVC approach.

Ok ❤️

waghanza:
@nathphon maven packaging is crying https://github.com/the-benchmarker/web-frameworks/actions/runs/11012642272/job/30593587492?pr=7809#step:7:2884

waghanza (approved):

PR #7824, opened 2024-10-01
Optimize Netty server configuration 

nathphon:
Enhanced Netty server settings for performance, including connection and idle timeouts, header validation, buffer size, and response compression.


waghanza (commented):
Could you explain a bit your PR ?

Why 15 seconds ?

nathphon:
@waghanza  

I've been trying to configure various things based on other Java frameworks like Vertx, ActiveJ, Jooby, Quarkus, and others to observe the test results against WebFlux. 

The 15-second setting is because I noticed that the benchmarking only lasts 15 seconds, so I set it as needed. I've seen that some languages and some frameworks also use cache through additional configuration.

waghanza:
Ok. Understood. My point is that those configurations should not be used.

I mean the idea of this project is to show how frameworks could react with a production ready configuration, generally the default one.

If you spot some frameworks with caching configuration, could you make some PRs to remove those configurations ?

nathphon:
@waghanza  I already remove cache control from Webflux , so please review and merge please , just want to see the performance next round

waghanza (approved):

PR #7836, opened 2024-10-06
use os.availableParallelism()

mario-huang:


waghanza (changes requested):
Thanks :tada:

Seems that using this function is the way to go https://nodejs.org/api/os.html#oscpus

Could you please replace it on all cluster.ts files ? I'll update results with it for tomorrow

waghanza:
Btw, thanks for your contribution.

I'll do it

waghanza (approved):

mario-huang:
Sure, I opened a PR

PR #7882, opened 2024-10-22
Express bun added

zahidcakici:
`express-bun` added to the project as cluster.

### DIFF
--- a/javascript/express-bun/cluster.ts
+++ b/javascript/express-bun/cluster.ts
@@ -7,10 +7,6 @@ if (cluster.isMaster) {
   for (let i = 0; i < cpus; i++) {
     cluster.fork();
   }
-
-  cluster.on("exit", (worker, code, signal) => {
-    console.log(`worker ${worker.process.pid} died`);
-  });
 } else {
   await import("./app.ts");
 }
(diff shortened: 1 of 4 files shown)


javascript/express-bun/cluster.ts
waghanza, line 1: Could you homogenize all cluster.ts, I mean for all bun based ?

I am not working with js/ts, I let you decide the most production-ready solution

waghanza (changes requested):

javascript/express-bun/cluster.ts
zahidcakici: It is because express is a npm package and we can't use the bun's build in cluster structure unlike other native bun frameworks. So I think a common structure is not possible right now. At least we can keep the same folder structure and Dockerfile. 

I also stuied the clustering deno case, tried many solutions but couldn't find a proper way. As you know the deno frameworks in repository is running on single instance since even if we run it with pm2 '-i max' command the same problem with npm packages. 

At least bun supports node cluster package unlike deno.

waghanza: Could you add some comments to specify that. I mean the idea is to use some wiki, or other things to have documentation.

Maybe we can specify it in a README in javascript folder.

waghanza (approved):

PR #7908, opened 2024-10-30
Rust/may_minihttp improve update

Xudong-Huang:
improve performance

### DIFF
--- a/rust/may_minihttp/Cargo.toml
+++ b/rust/may_minihttp/Cargo.toml
@@ -4,4 +4,13 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-may_minihttp = "0.1"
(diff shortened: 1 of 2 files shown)


rust/may_minihttp/Cargo.toml
```diff
@@ -4,4 +4,13 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-may_minihttp = "0.1"
\ No newline at end of file
+may_minihttp = { version = "0.1", default-features = false }
+
+[profile.release]
+opt-level = 3
```
waghanza, line 10: can we use this as command line arguments ?

Xudong-Huang: other rust frame work also use this value

waghanza (approved):

PR #7960, opened 2024-11-17
feat(JavaScript/ditsmod): upgrade ditsmod and added support for bun.

KostyaTretyak:


javascript/ditsmod/package.json
```diff
@@ -16,11 +16,11 @@
...
   "dependencies": {
     "@ditsmod/body-parser": "~2.17.1",
-    "@ditsmod/core": "~2.61.0",
+    "@ditsmod/core": "~2.61.1",
     "@ditsmod/routing": "~2.7.0"
   },
   "devDependencies": {
-    "@types/node": "22",
-    "typescript": "5.6.3"
+    "@types/node": "^20.6.1",
```
waghanza, line 23: We are using the last LTS of node here, 22.

KostyaTretyak: It doesn't affect the runtime though, but I'll update to version 22 now.

waghanza: sure, just a matter of providing an clean approach, for people without node / tech background

PR #7961, opened 2024-11-17
Remove binary stripping

Kayden1412:
@waghanza can you try compiling it again with binary stripping removed?

### DIFF
--- a/zig/httpz/build.zig
+++ b/zig/httpz/build.zig
@@ -18,7 +18,6 @@ pub fn build(b: *std.Build) void {
         .version = version,
         .optimize = optimize,
         .root_source_file = root_source_file,
-        .strip = true,
     });
     exe.root_module.addImport("httpz", httpz_mod);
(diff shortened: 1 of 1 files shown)


Kayden1412:
Can you send the error message

waghanza:
+ `zig build`
```
thread 26 panic: reached unreachable code
Panicked during a panic. Aborting.
error: the following build command crashed:
/app/.zig-cache/o/0569e2d413424966c9e4f413dd073d7e/build /usr/bin/zig /app /app/.zig-cache /root/.cache/zig --seed 0x855d6f8d -Z614e92966cd544db
```

waghanza:
yes, I see compilation stage on the console, but it is rewritten each time 

is their a way to create a compilation log ?

waghanza:
Ok, then I can follow this issue, and as I understand this will be fixed on the next release of zig

PR #7981, opened 2024-11-23
fix(javascript/fastify): fixed body parser.

KostyaTretyak:
As you can see from the benchmarks, fastify throws errors:

image

This is because POST requests are not actually parsed because there is no parser for this type:

https://github.com/the-benchmarker/web-frameworks/blob/19027d3c8e8a580b26e5c45c3cf2486313a9e22b/pipeline_post.lua#L3

### DIFF
--- a/javascript/fastify-bun/app.ts
+++ b/javascript/fastify-bun/app.ts
@@ -1,5 +1,6 @@
 import fastify from 'fastify';
 const app = fastify();
+app.register(await import('@fastify/formbody'));
 
 app.get('/', function (request, reply) {
   reply.send();
--- a/javascript/fastify-bun/package.json
+++ b/javascript/fastify-bun/package.json
@@ -1,6 +1,7 @@
 {
   "type": "module",
   "dependencies": {
+    "@fastify/formbody": "^8.0.1",
     "fastify": "~5.1.0"
   }
 }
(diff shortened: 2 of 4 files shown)


waghanza:
Thanks for this @KostyaTretyak. Glad to see anyone check the implementations health.

However, I'm in favour of changing the line you found in lua script.

I mean form encoded is not relevant here. In a server side context, data are sent in plain text of serialised in JSON, msgpack or else.

Better, remove the lua line, what do you think ?

PS : I'm thinking of changing the stress tool, maybe oha, which does not have lua support.

KostyaTretyak:
First, thank you for inviting me to join @the-benchmarker organization. But to be honest, I don't see when I will be able to devote time to it.

Regarding the replacement of the `Content-Type` header, I agree with you that much more often POST requests are formed in the `application/json` format. And in that case, really, this PR can be close.

waghanza:
@KostyaTretyak this invitation is kindly to formalize your contribution, but no rate is expected.

I mean even if you have no time to devote for this, I can understand.

I however use team members to ask some questions, when I have a doubt of something, or when I want to gather various opinion.

Let me know it this is ok for you (or not).

And I'll close this PR for now

KostyaTretyak:
I have extensive experience with the TypeScript stack, so if there are any questions regarding it, I’m happy to help.

waghanza:
typescript only or the whole javascript inclduing bun and other things ?

KostyaTretyak:
TypeScript/JavaScript stack.

PR #7993, opened 2024-11-27
fix(javascript/ditsmod-bun): fixed cluster mode for ditsmod-bun.

KostyaTretyak:


KostyaTretyak:
> As I understand this could replace pm2 then, to spread on all cores ? and then use bun run app.ts, correct ?

Yes

> If it such a thing could you replicate this on all javascript based framework ?

For bun?

waghanza:
> > As I understand this could replace pm2 then, to spread on all cores ? and then use bun run app.ts, correct ?
> 
> Yes
Could you change the command for `bun` in `javascript/config.yaml` ? this is the command used to create `Dockerfile` 😛 


> For bun?

Yes

KostyaTretyak:
Not all JavaScript-based frameworks require this because there are frameworks that support Bun's native cluster mode (this is what is currently specified in files with `Bun.spawn`). I am not familiar with all these frameworks to make such changes.

ragokan:
I can follow what others are doing, thanks for asking to us @waghanza

KostyaTretyak:
> Could you however change command in config.yaml file

I think you mean to change this code:

https://github.com/the-benchmarker/web-frameworks/blob/8cba03e6258ebbb8e13fbfc6254cb51cef00ab72/javascript/config.yaml#L43-L46

Please tell me when the local `config.yaml` for each framework is run. After `bootstrap` and before `command` in the `javascript/config.yaml` file?

waghanza:
You point correctly @KostyaTretyak . You need to replace bin run cluster.ts

KostyaTretyak:
I'm sorry, but I feel that I don't have the relevant experience to understand exactly what needs to be changed there. Not all JavaScript-based frameworks have a `cluster.ts` file. And if some frameworks have it, there is different content.

PR #7994, opened 2024-11-27
[Scala]: Adding Apache Pekko.

whiplash:
Akka (backed by Lightbend, who recently changed name and is actually _itself_ now called [Akka](https://akka.io)) [changed its licence](https://akka.io/blog/why-we-are-changing-the-license-for-akka).
Quickly after that, an open source fork was born under the Apache foundation with the name [Pekko](https://pekko.apache.org/).

Pekko is essentially Akka 2.6.

### DIFF
--- a/scala/pekkohttp/build.sbt
+++ b/scala/pekkohttp/build.sbt
@@ -1,0 +1,11 @@
+name := "server"
+scalaVersion := "3.5.2"
+
+val PekkoVersion = "1.1.2"
+val PekkoHttpVersion = "1.1.0"
+libraryDependencies ++= Seq(
+  "org.apache.pekko" %% "pekko-actor-typed" % PekkoVersion,
+  "org.apache.pekko" %% "pekko-stream" % PekkoVersion,
+  "org.apache.pekko" %% "pekko-http" % PekkoHttpVersion
+)
+enablePlugins(JavaAppPackaging)
--- a/scala/pekkohttp/config.yaml
+++ b/scala/pekkohttp/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: pekko.apache.org
+  version: 1.1.0
--- a/scala/pekkohttp/project/build.properties
+++ b/scala/pekkohttp/project/build.properties
@@ -1,0 +1,1 @@
+sbt.version=1.9.9
--- a/scala/pekkohttp/.gitignore
+++ b/scala/pekkohttp/.gitignore
@@ -1,0 +1,3 @@
+project/project
+project/target
+target/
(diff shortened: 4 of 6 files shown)


scala/pekkohttp/build.sbt
```diff
@@ -0,0 +1,11 @@
+name := "server"
+scalaVersion := "3.5.2"
+
+val PekkoVersion = "1.1.2"
```
waghanza, line 4:
```suggestion
val PekkoVersion = "[1.1,1.2)"
```

scala/pekkohttp/build.sbt
```diff
@@ -0,0 +1,11 @@
...
+val PekkoHttpVersion = "1.1.0"
```
waghanza, line 5:
```suggestion
val PekkoHttpVersion = "[1.1,1.2]"
```

scala/pekkohttp/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: pekko.apache.org
+  version: 1.1.0
```
waghanza, line 3:
```suggestion
  version: 1.1
```

scala/pekkohttp/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.9.9
```
waghanza, line 1: this is driven by container provided version, could you remove this file ?

waghanza (changes requested):

scala/pekkohttp/project/build.properties
whiplash: Done

scala/pekkohttp/build.sbt
```diff
@@ -0,0 +1,11 @@
+name := "server"
+scalaVersion := "3.5.2"
+
+val PekkoVersion = "1.1.0"
```
waghanza, line 4: could you use the same version for PekkoVersion and PekkoHttpVersion ?

whiplash: It looks like it's not possible => https://pekko.apache.org/docs/pekko-http/current/introduction.html#using-apache-pekko-http

btw, it's the same with Akka and Akka HTTP.

waghanza: maybe we should ping any core contributor for this, it appears not clear in the docs (at least for me), since version of both are pined

whiplash: Trust me, I know and have been used Akka. Akka is the actor framework. Later on they built Akka HTTP which is the HTTP framework on-top-of Akka. So Akka HTTP uses _Akka_ so it's quite natural that versions are different.

PR #8003, opened 2024-11-29
Javascript: improvements

enghitalo:
- versions upgrade
- version manager
- clusters
- Content-Length fixes
- gitignore


waghanza (changes requested):
no need to use cluster explicitly, imho, at least no for all javascript frameworks

`pm2` make it for you 😛

waghanza:
However, your content length bullet point, makes me say that we need to check it (in specs) to ensure all implementations are returning this header

waghanza (approved):

PR #8030, opened 2024-12-08
updating vixeny to 0.1.53

mimiMonads:
Sorry for that, skill issues. It should be working now, also I locked the version so it should give any problems in the future!


javascript/vixeny-deno/app.ts
```diff
@@ -1,22 +1,21 @@
-import { wrap } from "jsr:@vixeny/core@0.1.42";
+import { wrap } from "jsr:@vixeny/core@0.1.53";
```
waghanza, line 1: You need to specify the version here ?

PR #8077, opened 2024-12-25
Upgrade to ruby-3.4

p8:


waghanza (changes requested):
Could you please also change ruby/config.yaml ?

PR #8089, opened 2024-12-28
chore(deps): update viz to v0.10

fundon:


rust/viz/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   website: viz.rs
-  version: 0.9
+  version: 0.10
```
waghanza, line 3:
```suggestion
  version: '0.10'
```


Otherwise 0.1 will be printed

waghanza (changes requested):

rust/viz/config.yaml
fundon: thanks

rust/viz/config.yaml
```diff
@@ -1,3 +1,3 @@
 framework:
   website: viz.rs
-  version: 0.9
+  version: "0.10.0"
```
waghanza, line 3:
```suggestion
  version: "0.10"
```

waghanza (changes requested):

rust/viz/config.yaml
fundon: does not work

waghanza (approved):

PR #8119, opened 2025-01-04
feat: new framework 'hyper'

tomoriny:
A protective and efficient HTTP library for all.

https://hyper.rs/

### DIFF
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -4,11 +4,11 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-http = "1.2.0"
-http-body-util = "0.1.2"
-hyper = { version = "1.5.2", features = ["full"] }
-hyper-util = { version = "0.1.10", features = ["full"] }
-tokio = { version = "1", features = ["full"] }
+http = "1.2"
+http-body-util = "0.1"
+hyper = { version = "1.5", features = ["http1", "server"] }
+hyper-util = { version = "0.1", features = ["http1", "server", "tokio"] }
+tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
 
 [profile.release]
 opt-level = 3
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -4,7 +4,6 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-http = "*"
 http-body-util = "*"
 hyper = { version = "1.5", features = ["http1", "server"] }
 hyper-util = { version = "*", features = ["http1", "server", "tokio"] }
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -4,7 +4,7 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-http = "1.2"
+http = "1"
 http-body-util = "0.1"
 hyper = { version = "1.5", features = ["http1", "server"] }
 hyper-util = { version = "0.1", features = ["http1", "server", "tokio"] }
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -6,7 +6,7 @@ edition = "2021"
 [dependencies]
 http-body-util = "0.1"
 hyper = { version = "1.5", features = ["http1", "server"] }
-hyper-util = { version = "1", features = ["http1", "server", "tokio"] }
+hyper-util = { version = "0.1", features = ["http1", "server", "tokio"] }
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
 
 [profile.release]
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -4,10 +4,10 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-http = "1"
-http-body-util = "0.1"
+http = "*"
+http-body-util = "*"
 hyper = { version = "1.5", features = ["http1", "server"] }
-hyper-util = { version = "0.1", features = ["http1", "server", "tokio"] }
+hyper-util = { version = "*", features = ["http1", "server", "tokio"] }
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
 
 [profile.release]
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -4,9 +4,9 @@ version = "0.0.0"
 edition = "2021"
 
 [dependencies]
-http-body-util = "*"
+http-body-util = "0.1"
 hyper = { version = "1.5", features = ["http1", "server"] }
-hyper-util = { version = "*", features = ["http1", "server", "tokio"] }
+hyper-util = { version = "1", features = ["http1", "server", "tokio"] }
 tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
 
 [profile.release]
--- a/rust/hyper/Cargo.toml
+++ b/rust/hyper/Cargo.toml
@@ -1,0 +1,22 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http = "1.2.0"
+http-body-util = "0.1.2"
+hyper = { version = "1.5.2", features = ["full"] }
+hyper-util = { version = "0.1.10", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
+
+[profile.release]
+opt-level = 3
+debug = false
+debug-assertions = false
+lto = true
+panic = "abort"
+incremental = false
+codegen-units = 1
+rpath = false
+strip = false
--- a/rust/hyper/config.yaml
+++ b/rust/hyper/config.yaml
@@ -1,3 +1,3 @@
 framework:
   website: hyper.rs
-  version: 1.5.2
+  version: 1.5
--- a/rust/hyper/config.yaml
+++ b/rust/hyper/config.yaml
@@ -1,0 +1,3 @@
+framework:
+  website: hyper.rs
+  version: 1.5.2
--- a/rust/axum/src/main.rs
+++ b/rust/axum/src/main.rs
@@ -7,7 +7,7 @@ async fn main() {
     let router: Router = Router::new()
         .route("/", get(|| async {}))
         .route("/user", post(|| async {}))
-        .route("/user/{id}", get(|Path(id): Path<String>| async move { id }));
+        .route("/user/:id", get(|Path(id): Path<String>| async move { id }));
 
     let listener = tokio::net::TcpListener::bind("0.0.0.0:3000").await.unwrap();
     axum::serve(listener, router.into_make_service())
--- a/rust/axum/src/main.rs
+++ b/rust/axum/src/main.rs
@@ -7,7 +7,7 @@ async fn main() {
     let router: Router = Router::new()
         .route("/", get(|| async {}))
         .route("/user", post(|| async {}))
-        .route("/user/:id", get(|Path(id): Path<String>| async move { id }));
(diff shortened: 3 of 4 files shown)


waghanza:
We are going to include `hyper` in here @davidpdrsn @LucioFranco @seanmonstar, is this something you are not against ?

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http = "1.2.0"
```
waghanza, line 7: not needed imho, cause of https://github.com/hyperium/hyper/blob/30f2961e89eb306780d856e6e2c1ee10ffbbafd2/Cargo.toml#L24

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+http-body-util = "0.1.2"
+hyper = { version = "1.5.2", features = ["full"] }
+hyper-util = { version = "0.1.10", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
```
waghanza, line 11: not needed imho, cause of https://github.com/hyperium/hyper/blob/30f2961e89eb306780d856e6e2c1ee10ffbbafd2/Cargo.toml#L26

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http = "1.2.0"
+http-body-util = "0.1.2"
+hyper = { version = "1.5.2", features = ["full"] }
+hyper-util = { version = "0.1.10", features = ["full"] }
```
waghanza, line 10: is there a way to use the last stable version ? I mean avoid to specify version of this component, but only `hyper`

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http = "1.2.0"
+http-body-util = "0.1.2"
```
waghanza, line 8: what is this for ?

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
...
+hyper = { version = "1.5.2", features = ["full"] }
+hyper-util = { version = "0.1.10", features = ["full"] }
+tokio = { version = "1", features = ["full"] }
+
+[profile.release]
```
waghanza, line 13: is this the same options that for other rust based frameworks ?

I could be convenient for maintainance if this coul be defined as cli option in `Dockerfile`

rust/hyper/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: hyper.rs
+  version: 1.5.2
```
waghanza, line 3:
```suggestion
  version: 1.5
```

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http = "1.2.0"
+http-body-util = "0.1.2"
+hyper = { version = "1.5.2", features = ["full"] }
```
waghanza, line 9:
```suggestion
hyper = { version = "1.5", features = ["full"] }
```

what's the default valud of _features_ ?

waghanza (changes requested):

rust/hyper/Cargo.toml
tomoriny: Yes, See https://doc.rust-lang.org/cargo/reference/config.html#command-line-overrides

waghanza: Then could you change `rust/Dockerfile` and remove this section in all rust based frameworks ?

rust/hyper/Cargo.toml
tomoriny: These crates are used by hyper. https://github.com/hyperium/http

rust/hyper/Cargo.toml
tomoriny: Yes, we should create a new pull request to refactor.

waghanza: You could do it inside ;-) it will be clearer imo and trigger CI for all rust based

rust/hyper/Cargo.toml
waghanza: a better solution, imho, is to say that we want the latest version ? is this something doable ?

or maybe use features in hyper declaration

tomoriny: I replaced 'hyper::http' with 'http', but 'http-body-util' could not be replaced.

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,21 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+
+[dependencies]
+http-body-util = "*"
```
waghanza, line 7: this package is in dev-dependencies in https://github.com/hyperium/hyper-util/blob/5f055e36d6f945b51b4d4b023290f551ae7eb4b6/Cargo.toml#L36. any idea why ?

rust/hyper/Cargo.toml
```diff
@@ -0,0 +1,21 @@
...
+hyper = { version = "1.5", features = ["http1", "server"] }
+hyper-util = { version = "*", features = ["http1", "server", "tokio"] }
+tokio = { version = "1", features = ["rt", "rt-multi-thread", "net", "macros"] }
```
waghanza, line 10: is version required => https://github.com/hyperium/hyper/blob/30f2961e89eb306780d856e6e2c1ee10ffbbafd2/Cargo.toml#L26 ?

waghanza:
cc @chrislearn @msrd0

rust/hyper/Cargo.toml
msrd0: I really don't think that using `*` as a dependency version is ever a good, future-proof solution. If this is really what we want to do, I recommend you add a comment that explains why exactly this is a good idea here

rust/hyper/Cargo.toml
msrd0: if you mean _can we replace "1" with "*"_ - no we cannot, cargo will happily add two versions of the same crate to the dependency graph if they are _not_ semver compatible

rust/hyper/Cargo.toml
msrd0: I think besides `lto = true` and `panic = "abort"`, these are default values to begin with. Enabling `lto` is something I usually do and have seen other ppl do so probably a good default. Setting `panic = "abort"` might increase performance and probably decreases binary size but is something I haven't seen a lot being used in the wild (but then again I don't look at all rust crates ...)

rust/hyper/Cargo.toml
msrd0: Regarding why it is in the dev dependencies - because they use it in their example code, see https://github.com/search?q=repo%3Ahyperium%2Fhyper-util%20http_body_util&type=code

msrd0:
The Rust source code seems to not have been formatted using `rustfmt`. I don't know if this project has a policy on doing that or not, but I think that for future contributors it could come as a surprise if they run `cargo fmt` out of habbit and suddenly have modified every single line in the project.

waghanza:
> The Rust source code seems to not have been formatted using `rustfmt`.

You mean the implementation here, or `hyper` @msrd0 ?

msrd0:
I mean the .rs file part of this PR

tomoriny:
> The Rust source code seems to not have been formatted using `rustfmt`. I don't know if this project has a policy on doing that or not, but I think that for future contributors it could come as a surprise if they run `cargo fmt` out of habbit and suddenly have modified every single line in the project.

Actually, I ran rustfmt but with my personal config.

My rustfmt.toml.
```toml
indent_style = "Block"
reorder_imports = true
tab_spaces = 2
```

waghanza:
Maybe it is time to include a `rustfmt.toml` in here then, if accurated ;-)

msrd0:
... and that's why indentation with spaces is stupid and we should use tabs and have everyone set the tab size to their subjective preference.

Anyways, I'd either use the default config or add your config to the repo.

rust/hyper/Cargo.toml
waghanza: If you say that we should not use `panic = "abort"` since it is not very used (I mean we should target production grade applications), you are probably right @msrd0 , Im not in rust community to have my proper opinion about that, I'll follow ...

This is howver an other argument to use cli based compilation options in `rust/Dockerfile`

waghanza: If you are OK @tomoriny, I can push that in your PR

tomoriny: > You could do it inside ;-) it will be clearer imo and trigger CI for all rust based

I think it would be better and clearer if the changes were included in another pull request

msrd0: Btw it looks like those flags are already part of every rust framework, so it should probably be applied to all rust frameworks and not just hyper

tomoriny:
> Maybe it is time to include a `rustfmt.toml` in here then, if accurated ;-)

It's not necessary. Using the default config is ok.

tomoriny:
Ready to merge. @waghanza

waghanza:
I'll then force formatting on each release @msrd0

waghanza (approved):

rust/hyper/Cargo.toml
msrd0: I might have to take the `panic = "abort"` back, at least AlpineLinux sets `CARGO_PROFILE_RELEASE_PANIC="abort"` as a default  for all their packages.

waghanza: Not on alpine, results are computed on a linux (but on a mac mini) -> https://asahilinux.org/

@tomoriny I'll publish results for monday, I can do this stuff while releasing

waghanza: Then 
```sh
cargo build --release  \
  --config  profile.release.opt-level = 3 \
  --config  profile.release.debug = false \
  --config  profile.release.debug-assertions = false \
  --config  profile.release.lto = true \
  --config  profile.release.panic = "abort" \
  --config  profile.release.incremental = false \
  --config  profile.release.codegen-units = 1 \
  --config  profile.release.rpath = false \
  --config  profile.release.strip = false
```
should do the trick @msrd0 @tomoriny 

but not sure why we have a strip command in https://github.com/the-benchmarker/web-frameworks/blob/09da963a9423f77bf6eef1cd2fadc96d59979a7a/rust/Dockerfile#L9, perhaps better to use ` --config  profile.release.strip = true` imho

msrd0: The default values don't need to be specified manually (see https://doc.rust-lang.org/cargo/reference/profiles.html#release), so this should be enough:

```
cargo build --release  \
  --config  'profile.release.lto=true' \
  --config  'profile.release.panic="abort"' \
  --config  'profile.release.codegen-units=1'
```

> but not sure why we have a strip command

it's probably older than the strip option

msrd0: Alternatively you can take the approach AlpineLinux takes and set

```bash
export CARGO_PROFILE_RELEASE_PANIC="abort"
export CARGO_PROFILE_RELEASE_CODEGEN_UNITS=1
export CARGO_PROFILE_RELEASE_LTO="true"

Comment (waghanza) [reply rust/hyper/Cargo.toml:12]: @Xudong-Huang I have those options for may_minihttp
```toml
opt-level = 3
codegen-units = 1
panic = 'abort'
lto = 'thin'
debug = false
incremental = false
overflow-checks = false
```

and I have failures with options in https://github.com/the-benchmarker/web-frameworks/pull/8119#discussion_r1903130563

could you please explain the lto value here ?

waghanza: with lto = true I have server timeouts

PR #8138, opened 2025-01-09
Update csharp to .NET9

wuuer:


waghanza:
@ShreyasJejurkar do you have inputs for me ? I wonder if their is best practice / recommandation in C# world ? I mean if the wide majority of companies are using LTS or STS

ShreyasJejurkar:
@waghanza benchmarks should always run on the latest/greatest version so running on .NET 9 is ok here.

waghanza:
I understand but the idea is not "just" to compare raw performance in research and development mode, but more to mimic real life usage, then to use the same approach that could be recommended.

waghanza:
The idea is the same in the JavaScript world, I'm not sure it is business prone to use the latest (not LTS) nodejs in production.

Maybe for .NET the approach should not be the same then

PR #8179, opened 2025-01-25
Javascript: add koa with bun engine

vladDotH:


waghanza:
coul you check this PR @KostyaTretyak, specially the clustering part

if could be better if we use the same clusting technique for all bun base framework, imho

KostyaTretyak:
You won't be able to use the same code for all frameworks that run on Bun, because some of them use native Bun methods, and some (as implemented in this PR) use standard Node.js code.

That is, you need to ask the maintainers (or PR authors) whether the framework supports native Bun methods or not. If it does, then ask that the PR contain code with native Bun methods.

vladDotH:
Bun and express dont use native bun api, but according to the bun docs it implements node http modules, that koa or express relies on
(https://bun.sh/guides/ecosystem/express)
> Express and other major Node.js HTTP libraries should work out of the box. Bun implements the node:http and node:https modules that these libraries rely on.

About clustering part: I copied it from express-bun and fastify-bun. But I dont know if it is a good idea to use clustering with one engine and dont use with others (e.g. there is no clustering at express and express-deno benchmarks).

KostyaTretyak:
By the way, it would also be very good if all JavaScript frameworks also set `keepAliveTimeout: 0`, because this changes the benchmark results very significantly.

waghanza:
> About clustering part: I copied it from express-bun and fastify-bun. But I dont know if it is a good idea to use clustering with one engine and dont use with others (e.g. there is no clustering at express and express-deno benchmarks).

I'm working on it ... I'll add cluster.js / cluster.ts and ping some authors / contributors to check if it is a good implementation or not.

> By the way, it would also be very good if all JavaScript frameworks also set keepAliveTimeout: 0, because this changes the benchmark results very significantly.

Yes, planned to remove this for next round (after this one on monday)

KostyaTretyak:
> Yes, planned to remove this for next round (after this one on monday)

No, simply removing `keepAliveTimeout: 0` will not put everyone on the same footing, because for example, fastify has a non-standard setting (72 seconds):

https://github.com/fastify/fastify/blob/68c204697e8dbd0254a2cba551ba2e5c93f4eede/lib/configValidator.js#L19

KostyaTretyak:
> Maybe we could ask @mcollina for the rationales behind

In my opinion, it doesn't even matter why each framework sets a default value for `keepAliveTimeout` that is different from the default value for Node.js. It is more important that all frameworks have the same `keepAliveTimeout` value, as this significantly affects benchmarks. So, I suggest asking each framework maintainer to set `keepAliveTimeout: 0` (or setting it yourself, if you know how).

waghanza:
I understand this, but I http keep alive is not an HTTP sent by the client ?

I mean the keep alive param is sent by `wrk` when sieging. This will be remove next round
https://github.com/the-benchmarker/web-frameworks/blob/9af74ce398f4c4418ee356709004d59bb23004eb/.tasks/collect.rake#L68

vladDotH:
> but http keep alive is not an HTTP sent by the client ?

I suppose node http set `Connection: keep-alive` in some cases
https://github.com/nodejs/node/blob/ce6a62872081d720734d82bf975653a322795288/lib/_http_outgoing.js#L525C1-L526C78

KostyaTretyak:
> I mean the keep alive param is sent by `wrk` when sieging. This will be remove next round

Instead of `Connection: keep-alive` header, send `Connection: close` header. In this case, all web servers should close the connection after processing each request. And this will also set the same benchmark conditions for all web servers, and you don't need to ask maintainers to set `keepAliveTimeout: 0`.

javascript/koa-bun/app.ts
```diff
@@ -0,0 +1,22 @@
...
+    })
+    .post('/user', (ctx, next) => {
+      ctx.body = '';
+    });
+
+const server = app
+    .use(router.routes())
+    .listen(3000);
+
+server.keepAliveTimeout = 0
```
waghanza, line 22: so we can remove this ;-)

next round (not this one for monday, the header will be change)

waghanza (approved):

PR #8180, opened 2025-01-25
rust: bump ohkami to v0.21

kanarus:


rust/ohkami/Cargo.toml
```diff
@@ -16,5 +16,5 @@ rpath = false
 strip = false
 
 [dependencies]
-ohkami = { version = "0.20", features = ["rt_tokio"] }
-tokio  = { version = "1", features = ["macros", "rt-multi-thread"] }
+ohkami = { version = "0.21", features = ["rt_tokio"] }
+nio    = { version = "0.0" }
```
waghanza, line 20: what is the idea of this 0.0 ?

waghanza (changes requested):

rust/ohkami/Cargo.toml
kanarus: Sorry, I forgot to chage `features` to `rt_nio` from `rt_tokio`.
I'm trying https://crates.io/crates/nio runtime instead of tokio here.

waghanza: could you then create okhami-nio and okhami-tokio ?

as we have for javascript with -bun and -deno

waghanza (approved):

waghanza (changes requested):
better to expose result for both runtimes, `tokio` and `nio`

rust/ohkami-nio/Cargo.toml
```diff
@@ -5,15 +5,9 @@ edition = "2021"
...
 [profile.release]
-opt-level = 3
-debug = false
-debug-assertions = false
-lto = true
-panic = "abort"
-incremental = false
+lto           = true
+panic         = "abort"
 codegen-units = 1
```
waghanza: not needed 😛 

compilation options are in `rust/Dockerfile`

what is codegen-units for ?

rust/ohkami-tokio/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2021"
+publish = false
+
+[profile.release]
+lto           = true
+panic         = "abort"
+codegen-units = 1
```
waghanza, line 10:
```suggestion
```

waghanza (changes requested):
🎉

rust/ohkami-tokio/Cargo.toml
kanarus: They're quite standard optimizations for `release` profile ( other rust frameworks' do the same in their `Cargo.toml`s, too )

kanarus: almost all other `rust/*` are using

```toml
[profile.release]
opt-level = 3
debug = false
debug-assertions = false
lto = true
panic = "abort"
incremental = false
codegen-units = 1
rpath = false
strip = false
```
( see their `Cargo.toml`s )

but actually other than `lto = true`, `panic = "abort"` and `codegen-units = 1` are [default configuration of `release` profile](https://doc.rust-lang.org/cargo/reference/profiles.html#release), so I just set them

waghanza: I removed them in current branch, from advice of @msrd0. The compilation step is 
```
cargo build --release  \
  --config  'profile.release.lto=true' \
  --config  'profile.release.panic="abort"' \
  --config  'profile.release.codegen-units=1'
```

waghanza (approved):

PR #8221, opened 2025-02-07
Add native library

jaguililla:


kotlin/hexagon-netty-epoll/build.gradle
```diff
@@ -28,6 +28,7 @@ application {
 dependencies {
     implementation("com.hexagonkt:http_server_netty_epoll:$hexagonVersion")
     implementation("io.netty:netty-transport-native-epoll:$nettyVersion:linux-aarch_64")
+    implementation("io.netty:netty-transport-native-epoll:$nettyVersion:linux-x86_64")
```
waghanza, line 31: Not sure to understand @jaguililla. Results are currently computed on aarch64 (asahi on m1)

waghanza: I have not tested yet, but I doubt ...

Is their a way to "guess" the architecture in a gradle build file, maybe with https://github.com/gradle/gradle/blob/8b40d5b18372f6cc2158ff508c920fc05196b212/platforms/native/platform-native/src/main/java//org/gradle/nativeplatform/platform/NativePlatform.java#L45

waghanza: no change @jaguililla. here is the command I use to compute results 
```
wrk -H 'Connection: close' --connections 512 --threads 8 --duration 15 --timeout 1 --script /home/waghanza/web-frameworks/pipeline.lua http://172.17.0.2:3000/" 
```

maybe you can open a PR with debug enabled, so we can discover the glitch

PR #8113, opened 2025-01-03
New release

waghanza:
Hi,

This PR will update release.

I have however an issue with `axum`, with version **0.8**

```
root@699c7e31e802:/# /usr/src/app/target/release/server
thread 'main' panicked at src/main.rs:10:10:
Path segments must not start with `:`. For capture groups, use `{capture}`. If you meant to literall@khry match a segment starting with a colon, call `without_v07_checks` on the router.
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
Aborted (core dumped)
```

cc @wyatt-herkamp @krishnaTORQUE @davidpdrsn @jplatte

PR #8225, opened 2025-02-09
fix(hono-deno): use npm library

ryuapp:
Hono is no longer published on `deno.land/x`. Change to refer to the package on npm to apply the latest version.
https://github.com/honojs/hono/blob/main/docs/MIGRATION.md


ryuapp:
#8113 seems to fix this problem, so I'll close it.

PR #8228, opened 2025-02-10
Configure PHP / Spiral app

roxblnfk:
- removed unused modules
- configure web server for production


php/spiral/.rr.yaml
waghanza, line 1: Is this file suitable for all frameworks using road runner ?

waghanza (commented):
Thanks for this PR, I  am thrilled to have production ready implementations here.

May I ping you for any (minor or spec change) update on spiral here ? Renovatebot updates each minor

php/spiral/.rr.yaml
roxblnfk: > May I ping you for any (minor or spec change) update on spiral here ? Renovatebot updates each minor

Of course. I don't think there will be any problems with updating minor versions, but if there are, feel free to reach out.
Also, feel free to contact me if you are expanding test cases for FullStack frameworks that involve data handling.

> Is this file suitable for all frameworks using road runner ?

Yes. If you don't want to test the console's throughput, it's better not to log every request.

roxblnfk: Also it's important to enable `ext-protobuf` in PHP. Without that a slow PHP-implementation will be used.

roxblnfk: This is PHP-side extension. Roadrunner knows nothing about PHP. It just runs a worker process and sends requests / receives responses

PR #8269, opened 2025-02-25
update dependency go/web

akyoto:
Hi,

is `data.json` updated automatically?
I can add it to the commit if needed.


waghanza (approved):
No data.json and data.min.json are automatically updated on the release process

PR #8287, opened 2025-03-10
Simplify Workerman code

joanhey:
And with better performance

ping @walkor

### DIFF
--- a/php/workerman/start.php
+++ b/php/workerman/start.php
@@ -10,7 +10,7 @@ $worker->onMessage = static function ($connection, $request) {
     $path = $request->path();
     match ($path) {
         '/', '/user' => $connection->send(''),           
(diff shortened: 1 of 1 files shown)


waghanza (changes requested):
is is not a matter of (only) performance ...

I mean the idea is to compare fraeworks (perf, overhead, latency, readability ...).

Maybe best to introduce routing feature 

cc @walkor

joanhey:
I only simplify the code for better readability.
And with the change `match()` is faster than `switch()`.

I can't add a router, as Workerman don't have any official router, any dev can use any router.

Some framworks in this bench use Workerman with a router:
* Mark:  Workerman with fastroute
* Webman
* Comet
* MixPHP
* Imi
* ...

waghanza (approved):
Ok, for now ...

I mean the endpoints will change so this code will need to be change, but we will do this later ...

PS : Please do not merge, I'll do it

PR #8290, opened 2025-03-16
feat(fsharp): add Oxpecker sample

panesofglass:
Adding Oxpecker to the F# options.

### DIFF
--- a/fsharp/oxpecker/Program.fs
+++ b/fsharp/oxpecker/Program.fs
@@ -13,9 +13,15 @@ let userIdHandler: EndpointHandler =
         let id = ctx.Request.RouteValues["id"] :?> string
         ctx.WriteText(id)
 
-let endpoints =
-    [ GET [ route "/" emptyHandler; route "/user/{id}" userIdHandler ]
-      POST [ route "/user" emptyHandler ] ]
+let endpoints = [
+    GET [
+        route "/" emptyHandler
+        route "/user/{id}" userIdHandler
+    ]
+    POST [
+        route "/user" emptyHandler
+    ]
+]
 
 let configureLogging (log: ILoggingBuilder) = log.ClearProviders() |> ignore
 
--- a/fsharp/oxpecker/Program.fs
+++ b/fsharp/oxpecker/Program.fs
@@ -1,0 +1,34 @@
+open Microsoft.AspNetCore.Builder
+open Microsoft.AspNetCore.Hosting
+open Microsoft.Extensions.DependencyInjection
+open Microsoft.Extensions.Hosting
+open Microsoft.Extensions.Logging
+open Oxpecker
+
+let userNameHandler (name: string) : EndpointHandler = _.WriteText(name)
+
+let endpoints =
+    [ GET [ route "/" <| text ""; routef "/user/{%s}" text ]
+      POST [ route "/user" <| text "" ] ]
+
+let configureLogging (log: ILoggingBuilder) = log.ClearProviders() |> ignore
+
+let configureServices (services: IServiceCollection) =
+    services.AddRouting().AddOxpecker() |> ignore
+
+let configureApp (app: IApplicationBuilder) =
+    app.UseRouting().UseOxpecker(endpoints) |> ignore
+
+let args = System.Environment.GetCommandLineArgs()
+
+Host
+    .CreateDefaultBuilder(args)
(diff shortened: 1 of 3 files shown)


waghanza:
@Lanayx we want to introude Oxpecker in here, this projet is a benchmark comparing frameworks. Is such an introduction ok for you ?

Lanayx:
Hi! I have no objections of adding Oxpecker here. Will add several comments to the PR

fsharp/oxpecker/Program.fs
```diff
@@ -0,0 +1,34 @@
...
+open Microsoft.AspNetCore.Hosting
+open Microsoft.Extensions.DependencyInjection
+open Microsoft.Extensions.Hosting
+open Microsoft.Extensions.Logging
+open Oxpecker
+
+let userNameHandler (name: string) : EndpointHandler = _.WriteText(name)
+
+let endpoints =
+    [ GET [ route "/" <| text ""; routef "/user/{%s}" text ]
```
Lanayx, line 11: For maximum performance `routef` should be substituted to manual route value retrieval ([example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/FSharp/oxpecker/src/App/Program.fs#L58) )

fsharp/oxpecker/Program.fs
```diff
@@ -0,0 +1,34 @@
+open Microsoft.AspNetCore.Builder
+open Microsoft.AspNetCore.Hosting
+open Microsoft.Extensions.DependencyInjection
+open Microsoft.Extensions.Hosting
+open Microsoft.Extensions.Logging
+open Oxpecker
+
+let userNameHandler (name: string) : EndpointHandler = _.WriteText(name)
```
Lanayx, line 8: This handler is unused.

fsharp/oxpecker/Program.fs
Lanayx: Also, for max perf, first GET handler could just return `Task.CompletedTask` or write content type header if it's required ([example](https://github.com/TechEmpower/FrameworkBenchmarks/blob/master/frameworks/FSharp/oxpecker/src/App/Program.fs#L84))

panesofglass:
Thank you for the feedback, @Lanayx. I'm made the requested changes. How does that look?

fsharp/oxpecker/Program.fs
```diff
@@ -0,0 +1,41 @@
...
+open Microsoft.Extensions.DependencyInjection
+open Microsoft.Extensions.Hosting
+open Microsoft.Extensions.Logging
+open Oxpecker
+
+let emptyHandler: EndpointHandler = fun _ -> Task.CompletedTask
+
+let userIdHandler: EndpointHandler =
+    fun ctx ->
+        match ctx.TryGetRouteValue<string>("id") with
```
Lanayx, line 13: I think id is not optional in this benchmark, so it can be simplified to
```fsharp
let id = ctx.Request.RouteValues["id"] :?> string
ctx.WriteText(id)
```

fsharp/oxpecker/Program.fs
```diff
@@ -0,0 +1,41 @@
...
+        | Some id -> ctx.WriteText(id)
+        | None -> Task.CompletedTask
+
+let endpoints =
+    [ GET [ route "/" emptyHandler; route "/user/{id?}" userIdHandler ]
```
Lanayx, line 18: Two comments
1. I think id is not optional in the benchmark so it can be rewritten to `route "/user/{id}" userIdHandler`
2. To make it overall endpoints list look nicer I'd rewrite it into
```fsharp
let endpoints = [ 
    GET [ 
        route "/" emptyHandler
        route "/user/{id}" userIdHandler 
    ]      
    POST [ 
        route "/user" emptyHandler 
    ] 
]
```

panesofglass: I can tackle 1. 2. is a bit more challenging with fantomas installed. :)

Lanayx (approved):

fsharp/oxpecker/Program.fs
Lanayx: Thank you, approved!

waghanza (approved):
Thans both of you.

I'll publish results on monday.

Please do not merge, I'll handle

PR #8292, opened 2025-03-18
feat: support Bun built-in http server

medz:
Bun is not just a runtime, it is also a unified PM and development framework.

The unified Runtime and PM have been completed in Bun, and the Framework API has taken shape.

Therefore, we added Bun's server to the test❤️

> Also, I found that all the clustering code under javascript was implemented incorrectly.

### DIFF
--- a/javascript/bun/bun.lock
+++ b/javascript/bun/bun.lock
@@ -1,0 +1,21 @@
+{
+  "lockfileVersion": 1,
+  "workspaces": {
+    "": {
+      "devDependencies": {
+        "@types/bun": "^1.2.5",
+      },
+    },
+  },
+  "packages": {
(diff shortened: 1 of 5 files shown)


medz:
Completed for review

CC @waghanza

javascript/bun/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  github: oven-sh/bun
+  version: 1.2.5
```
waghanza, line 3:
```suggestion
  version: 1.2
```

javascript/bun/bun.lock
waghanza, line 1: Not needed ;-)

waghanza (changes requested):
Thanks it coul be useful to compare bun framework to this reference

PR #8411, opened 2025-04-26
chore(version): update rama to latest edition and framework version

GlenDC:
- rust edition 2024
- rama version 0.2.0-alpha.13

### DIFF
--- a/rust/rama/Cargo.toml
+++ b/rust/rama/Cargo.toml
@@ -1,10 +1,11 @@
 [package]
 name = "server"
 version = "0.0.0"
-edition = "2021"
+edition = "2024"
 publish = false
+rust-version = "1.85"
 
 [dependencies]
-rama = { version = "0.2.0-alpha.6", features = ["http-full"] }
+rama = { version = "0.2.0-alpha.13", features = ["http-full"] }
 serde = "1.0"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
--- a/rust/rama/config.yaml
+++ b/rust/rama/config.yaml
@@ -1,3 +1,3 @@
 framework:
   github: plabayo/rama
-  version: 0.2.0-alpha.6
+  version: 0.2.0-alpha.13
--- a/rust/rama/src/main.rs
+++ b/rust/rama/src/main.rs
@@ -23,7 +23,7 @@ async fn main() {
                 .get("/", StatusCode::OK)
                 .post("/user", StatusCode::OK)
                 .get(
-                    "/user/:id",
+                    "/user/{id}",
                     async |Path(GetUserParams { id }): Path<GetUserParams>| id,
                 ),
         )
--- a/rust/rama/src/main.rs
+++ b/rust/rama/src/main.rs
@@ -1,10 +1,10 @@
 use rama::{
     http::{
-        matcher::HttpMatcher,
-        server::HttpServer,
-        service::web::{extract::Path, match_service},
         StatusCode,
+        server::HttpServer,
+        service::web::{Router, extract::Path},
     },
+    net::address::SocketAddress,
     rt::Executor,
 };
 use serde::Deserialize;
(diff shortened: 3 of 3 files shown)


rust/rama/Cargo.toml
```diff
@@ -1,10 +1,11 @@
 [package]
 name = "server"
 version = "0.0.0"
-edition = "2021"
+edition = "2024"
```
waghanza, line 4: what does it change ?

I mean perhaps, we should replicate this for all rust

rust/rama/Cargo.toml
```diff
@@ -1,10 +1,11 @@
...
 publish = false
+rust-version = "1.85"
```
waghanza, line 6: I prefer to stick to version provided in dockerfile (since their is a bot taking care of updates)

waghanza (changes requested):

rust/rama/Cargo.toml
waghanza: Sorry, after I have read the documentation it appears that is the minimal version

GlenDC: Yes it's the MSRV, more important for libs than binaries, but it still gives a nice error warning. It doesn't define the version you actually run it with (which is 1.86 in your docker last i checked).

Rama is already on edition 2024 (which is only available since rust 1.85).
As such it won't be able to run on edition 2021.

A rust project on edition X can use a lib on edition Y if X >= Y. So a rust project on edition 2024 can us a lib on edition 2021, but a project on edition 2021 cannot use a lib on edition 2024.

In case you want we can turn the rust directory into a rust workspace and enforce the same MSRV and edition on _all_ rust benchmarks, but shouldn't really matter too much as they already run on the same rust version anyway (defined by your dockerfile) and the edition is more about unlocking (language) features, they shouldn't really make a difference in performance AFAIK.

---

So if you agree with all that I think we can resolve these comments and merge, but up to you of course. Open to any feedback.

rust/rama/Cargo.toml
GlenDC: https://github.com/the-benchmarker/web-frameworks/pull/8411#discussion_r2062562302

waghanza: Clear enough.

The it should not affect benchmark.

As I understand, it's more about features for `rama`. The should not this be driven be `rama` itself, not consumer app. I mean, when `rama` will use features from edition **2025**, will it be modified here ?

GlenDC: Editions are only every 3 years, so that would be 2027.

It should not affect the benchmark no.

But because rama makes use of edition 2024 since a couple of versions ago, there is no way to update rama in a project as long as it's not also enabling edition 2024.

Editions are a mechanism in Rust to introduce breaking changes into the language without affecting existing libraries. E.g. there are still libraries out there on edition 2018 (the first edition). The key here is backwards compatible. Meaning that if a library is on edition 2024, you cannot use it as long as you are also not on edition 2024 (or later, which would be 2027 or 2030 or ...).

Read more: https://doc.rust-lang.org/edition-guide/editions/

waghanza: Ok, so it is a kind on minimal edition key.

Meaning that librairies in 2018 could be working in 2024, if no neww features of 2024 is used

GlenDC: Yes and it is guaranteed those libraries on edition 2018 do not use features from edition 2021 / 2024 because they cannot do so, guaranteed by their defined edition. If you do not define edition it assumes edition 2018 (the original stable rust version basically).

However rama does already declare itself as edition 2024 since a couple of versions ago, so as a project (e.g. this benchmark rama binary project) you cannot use one of these latest rama versions if you do not also declare that you opt-in to edition 2024.

waghanza (approved):
Thanks for clear explanaitions @GlenDC

PR #8499, opened 2025-07-13
chores(lihil): Update website url in config.yaml

raceychan:


python/lihil/config.yaml
```diff
@@ -1,7 +1,7 @@
 framework:
-  website: www.lihil.cc/lihil/
+  website: www.lihil.cc
   github: raceychan/lihil
-  version: 0.2
+  version: 0.2.22
```
waghanza, line 4: no need to be so precise 😛 

```suggestion
  version: 0.2
```

It is aligned with sematics on versionning here, using `X.Y` using `Z` will be time consuming (more than now) for maintenance

raceychan: @waghanza would you be able to directly make changes on my commit? I do agree that we should keep it in the x.y format

waghanza (approved):

PR #8551, opened 2025-07-27
jooby: 4.0

jknack:
- upgrade to latest

### DIFF
--- a/java/jooby/config.yaml
+++ b/java/jooby/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: jooby.io
-  version: 3.11
+  version: 4.0
 
   build:
     - mvn clean package
--- a/java/jooby/pom.xml
+++ b/java/jooby/pom.xml
@@ -15,7 +15,7 @@
     <!-- Startup class -->
     <application.class>benchmark.App</application.class>
 
-    <jooby.version>[3.11,3.12)</jooby.version>
+    <jooby.version>[4.0.3,4.1]</jooby.version>
 
     <maven.compiler.release>21</maven.compiler.release>
     <maven.compiler.parameters>true</maven.compiler.parameters>
--- a/kotlin/jooby/build.gradle
+++ b/kotlin/jooby/build.gradle
@@ -6,7 +6,7 @@ buildscript {
 
     ext {
         kotlinVersion = "2.2.+"
-        joobyVersion = "3.11.+"
+        joobyVersion = "4.0.+"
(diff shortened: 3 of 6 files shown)


waghanza (changes requested):
Could you also update the kotlin version @jknack ?

java/jooby/config.yaml
```diff
@@ -9,4 +9,4 @@ framework:
     - target/benchmark.jar
 
   command: >
-    /usr/bin/java -jar target/benchmark.jar benchmark.App
+    /usr/bin/java -Dio.netty.disableHttpHeadersValidation=true -Dio.netty.buffer.checkBounds=false -Dio.netty.buffer.checkAccessible=false -jar target/benchmark.jar benchmark.App
```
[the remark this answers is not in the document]
jknack: not really, but let me revert

jknack:
@waghanza forgot about kotlin. Done

java/jooby/pom.xml
```diff
@@ -15,7 +15,7 @@
     <!-- Startup class -->
     <application.class>benchmark.App</application.class>
 
-    <jooby.version>[3.11,3.12)</jooby.version>
+    <jooby.version>4.0.3</jooby.version>
```
waghanza, line 18: Is it safe to keep range @jknack ?

I mean specifying a range could be good (to decrease maintainance), but maybe we good use a non-production ready version (like a rc, or milestone)

kotlin/jooby/build.gradle
```diff
@@ -6,7 +6,7 @@ buildscript {
 
     ext {
         kotlinVersion = "2.2.+"
-        joobyVersion = "3.11.+"
+        joobyVersion = "4.0.3"
```
waghanza, line 9:
```suggestion
        joobyVersion = "4.0.+"
```

If it safe 😃

kotlin/jooby/src/main/kotlin/benchmark/App.kt
```diff
@@ -1,18 +1,22 @@
 package benchmark
 
 import io.jooby.ExecutionMode.EVENT_LOOP
+import io.jooby.ServerOptions
 import io.jooby.StatusCode
+import io.jooby.kt.Kooby
```
waghanza, line 6: I'm confused, is it named jooby or kooby on kotlin ?

Seems that we have two frameworks here, if so

waghanza (approved):

PR #8608, opened 2025-08-29
rust: bump ohkami to v0.24 with adding new flavors

kanarus:
- updating `ohkami` to v0.24
- adding `ohkami-glommio` bench which runs `ohkami` on [`glommio`](https://github.com/DataDog/glommio) runtime
- adding `ohkami-smol` bench which runs `ohkami` on [`smol`](https://github.com/smol-rs/smol) runtime


kanarus:
removed `ohkami-glommio` because of the CI's failure.

PR #8639, opened 2025-09-04
khttp: use default serving strategy

karlivory:
Changes the serving strategy for `khttp`.

`serve_epoll` was purely optimized for the use-case of only keep-alive connections (e.g. benchmarks), `serve` should have much better throughput if tcp connections are not reused.

### DIFF
--- a/rust/khttp/src/main.rs
+++ b/rust/khttp/src/main.rs
@@ -19,5 +19,5 @@ fn main() {
     });
 
     // serve
-    app.build().serve_epoll().unwrap();
+    app.build().serve().unwrap();
 }


waghanza:
Thanks @karlivory . The main idea is to stick closest to what could be made for a production ready application

karlivory:
> Thanks @karlivory . The main idea is to stick closest to what could be made for a production ready application

Yup, this change should reflect that. I designed `serve_epoll` purely on the assumption that connections are reused so it has about 15% better performance in that niche use case but sacrifices a lot of performance in the general case.

`serve` is more general, it handles both reused and closed connections similarly.

PR #8653, opened 2025-09-10
Ultimate Express: node 24

cesco69:


cesco69:
I thought it was used to set the node js version in the benchmark

waghanza:
No the version is pinned on dockerfile

PR #8684, opened 2025-09-17
Add MoroJS web framework implementation

M-Chris:
- Add app.mjs with basic server setup
- Add cluster.mjs for clustering support
- Add config.yaml with framework configuration
- Add package.json with dependencies

Thanks!

### DIFF
--- a/javascript/morojs/config.yaml
+++ b/javascript/morojs/config.yaml
@@ -1,3 +1,3 @@
 framework:
   website: morojs.com
-  version: 1.5.7
+  version: 1.5


javascript/morojs/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  website: morojs.com
+  version: 1.5.2
```
waghanza, line 3:
```suggestion
  version: 1.5
```

waghanza (changes requested):
Thanks for your contribution 🎉

javascript/morojs/config.yaml
M-Chris: Thanks for making the testing grounds 💯  just bumped the version hopefully that bump is ok before your next build

M-Chris: Oh also made the patch version go away, which I think was the change request didnt see that till I had pushed the bump

M-Chris:
Something is odd in the results @waghanza can you provide some insight? very surprised to see the results be what they are when we've tested in multiple environments and methods and have never come anywhere close to what your reporting. We've used work and autocannon and we are quite confident our results should be far better.

M-Chris:
Talking like 25x orders of magnitude different

PR #8702, opened 2025-09-23
Fix brahma firelight v1.0.6 clean

Shyam20001:
---
### **fix(brahma-firelight): pin to v1.0.6 and correct repo URL**
---

### **Description**

```
- Pin brahma-firelight to v1.0.6 (intended baseline)
- Fix malformed GitHub repo URL in config
- Update handler to support /user/:id

```

---

### DIFF
--- a/javascript/brahma-firelight/app.ts
+++ b/javascript/brahma-firelight/app.ts
@@ -1,45 +1,20 @@
-import { useBrahma, startServer } from "brahma-firelight";
+import { createApp } from 'brahma-firelight';
 
-useBrahma((req) => {
-    const { path } = req;
+const app = createApp();
 
-    // GET --> /
-    if (path === "/") {
-        return {
-            headers: { "Content-Type": "text/plain" },
-            status: 200,
-            body: "",
-        };
-    }
 
-    // POST --> /user
-    if (path === "/user") {
-        return {
-            headers: { "Content-Type": "text/plain" },
-            status: 200,
-            body: "",
-        };
-    }
-
-    // GET --> /user/:id
-    let id = null;
-    const m = path.match(/^\/user\/([^/]+)$/);
-    if (m) id = decodeURIComponent(m[1]);
-
-    if (id !== null) {
-        return {
-            headers: { "Content-Type": "text/plain" },
-            status: 200,
-            body: `${id}`,
-        };
-    }
-    // Fallback to 404 - Not Found
-    return {
-        status: 404,
-        body: "Route not found",
-    };
+// GET
+app.get('/', (req, res) => {
+    res.text('');
 });
 
-const port = 3000;
-const host = "0.0.0.0";
-startServer(host, +port);
+app.get('/user/:id', (req, res) => {
+    res.text(req.params.id)
+})
+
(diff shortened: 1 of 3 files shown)


javascript/brahma-firelight/app.ts
```diff
@@ -1,19 +1,45 @@
...
-const app = createApp();
+useBrahma((req) => {
+    const { path } = req;
 
-// GET
-app.get('/', (req, res) => {
-  res.text('');
-});
+    // GET --> /
+    if (path === "/") {
```
waghanza, line 7: this seems to be an implementation tweaked for benchmark. the idea is to have a production-ready implementation to have the fairest comparison beetween each framework

javascript/brahma-firelight/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   website: https://github.com/Shyam20001/rsjs
```
waghanza, line 2: no need 😛 

the second parameter will be the only show on UI

javascript/brahma-firelight/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "module",
   "dependencies": {
-    "brahma-firelight": "^1.5.2"
+    "brahma-firelight": "1.0.6"
```
waghanza, line 4: I don't understand the last stable release is 1.5
https://github.com/Shyam20001/rsjs/releases/tag/1.5.0

waghanza (changes requested):

javascript/brahma-firelight/config.yaml
Shyam20001: Yeah but on clicking the brama-firelight it shows invalid url check here // https://https//github.com/Shyam20001/rsjs // MISSING : 
https://web-frameworks-benchmark.netlify.app/result

javascript/brahma-firelight/app.ts
waghanza: Sure, not saying that Brahma is not production ready, but using if to define routes (rather than express like API for routing) is not something expected to be written when we have an application to be delivered

javascript/brahma-firelight/config.yaml
waghanza: Yes, intended... Having website key means that https will be added, but github key means that https://github.com will be added.

There is only one info to show and website has precedence over github

Shyam20001:
The requested changes are done successfully. Now brahma-firelight is using more stable version 1.5.2 (stable & prod ready) If any suggestions just let me know thanks.

javascript/brahma-firelight/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "type": "module",
   "dependencies": {
-    "brahma-firelight": "^1.5.2"
+    "brahma-firelight": "1.5.2"
```
waghanza, line 4:
```suggestion
    "brahma-firelight": "^1.5.2"
```

better to avoind sticky version 😛

waghanza (changes requested):

PR #8732, opened 2025-10-04
Helidon SE pushed to 4.3

whiplash:


waghanza:
@whiplash Thie PR take version 4.3.0-M4, I tin it should be change to `[4.2,4.3.0-M1)`, will try in the next release

whiplash:
Strange... I See 4.3.0 and 4.3.1 [here](https://mvnrepository.com/artifact/io.helidon.applications/helidon-se)

waghanza:
In deed there is 

- 4.3.1
- 4.3.0 <- maven consider that `[4.0,4.3)` stops here and take all release before either stable or unstable
- 4.3.0-M4
- 4.3.0-M3
- 4.3.0-M2
- 4.3.0-M1

as observe with restheart (I think, the interval is evaluated by maven to get all milestones, because it consider only the stable one for interval)

maybe better to strict to strict version then, I mean only specifying 4.3.1

PR #8746, opened 2025-10-08
[D] CGI version bump and new multicore VibeD code

cyrusmsk:
For CGI new version was used.
For Serverino just reoder.
For Vibe.D - new multicore code and more idiomatic "router" usage.


cyrusmsk:
Changed configs.

PR #8872, opened 2025-11-22
disabled cluster mode

Shyam20001:
Well the cluster mode is disabled for now. Also i noticed that cluster setup js frameworks in your current monorepo doesn't pay off. If you take a look at express, hono and fastify obviously they perfom low in single core mode but on cluster they possess the ability to hit `3x` their current req/s. Unfortunately here its not happening.

Speaking about uws-js and ultimate-express they are c++ bindings also support so_reuseport. I guess they can even hit more req/s without your cluster setup which is killing them. I guess this runs where not done on server grade processor but a consumer grade pc which ultimately turns the table.

### DIFF
--- a/javascript/brahma-firelight/app.ts
+++ b/javascript/brahma-firelight/app.ts
@@ -4,7 +4,7 @@ const app = createApp();
 
 // GET
 app.get('/', (req, res) => {
-  res.text('');
+  res.send(200);
 });
 
 app.get('/user/:id', (req, res) => {
@@ -13,7 +13,7 @@ app.get('/user/:id', (req, res) => {
 
 // POST
 app.post('/user', (req, res) => {
-  res.text('');
+  res.send(200);
 });
 
 app.listen('0.0.0.0', 3000);
--- a/javascript/brahma-firelight/cluster.ts
+++ b/javascript/brahma-firelight/cluster.ts
@@ -1,12 +1,15 @@
-// brahma-firelight v1.5.18 now supports true multi-core ⚡⚡
+// brahma-firelight v1.5.18+ now supports true multi-core ⚡⚡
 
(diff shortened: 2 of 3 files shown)


waghanza:
HI @Shyam20001 ,

Maybe ... Having cluster mode or not is a  hard question for javascript/ typescript.

Sine node.js (not sur about bun and deno) are using a single thread, maybe we should NOT use multi-cores for javascript based frameworks.

What do you think ?

Regards,

Shyam20001:
We can use cluster but i am not sure about your bench-mark environment. Well Techem-benchmarks often use 56core totally server grade processors with 100+cpus. here clustering gives u op perf undoubtebly. 

All though i disabled cluster from my side. I have an optional switch called Rust's Tokio multi core mode. I will test on the upcoming runs. For now i am reverting back to original changes here.  

kindly do ping me if any changes required. waiting for approval.

waghanza:
As you whish for you framework...

But, my opinion is ... not to do this.

The idea here is to benchmark for production backend system (like API), not heavy CPU workload (where I'll advice to use single thread in node).

Actually, benchmark are running only personal server, but the goal is to change and use real vms, with real network... this is in the roadmap

waghanza (approved):

Shyam20001:
> As you whish for you framework...
> 
> But, my opinion is ... not to do this.
> 
> The idea here is to benchmark for production backend system (like API), not heavy CPU workload (where I'll advice to use single thread in node).
> 
> Actually, benchmark are running only personal server, but the goal is to change and use real vms, with real network... this is in the roadmap

Cool that's good idea and now this framework is actually using single core only undoubtebly. Looking forward on large VMs.
thanks..

PR #8874, opened 2025-11-23
chore: update elixir and its frameworks

nakrovati:


waghanza:
Thanks @nakrovati . I wonder why we have use this image (I mean hexpm). Is there any difference between https://hub.docker.com/r/hexpm/elixir and https://hub.docker.com/_/elixir ?

nakrovati:
The difference is that hexpm images are immutable and allow you to specify exact versions of Erlang and OS, while the official Elixir image does not. Otherwise, there seem to be no differences.> Thanks @nakrovati . I wonder why we have use this image (I mean hexpm). Is there any difference between https://hub.docker.com/r/hexpm/elixir and https://hub.docker.com/_/elixir ?

PR #9068, opened 2026-02-23
update fiber framework to version 3.0

ReneWerner87:
https://github.com/gofiber/fiber/releases/tag/v3.0.0

### DIFF
--- a/go/fiber/config.yaml
+++ b/go/fiber/config.yaml
@@ -1,6 +1,6 @@
 framework:
   website: gofiber.io
-  version: 2.52
+  version: 3.0
 
   engines:
     - fasthttp
--- a/go/fiber/go.mod
+++ b/go/fiber/go.mod
@@ -1,3 +1,3 @@
 module main
 
-require github.com/gofiber/fiber/v2 v2.52.10
+require github.com/gofiber/fiber/v3 v3.0.0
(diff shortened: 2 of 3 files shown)


ReneWerner87:
@waghanza My initial idea was to include both versions simultaneously, as in the Techempower benchmark.

image

[https://www.techempower.com/benchmarks](https://www.techempower.com/benchmarks/#section=test&runid=1d5bfc8a-5c4a-4fb2-a792-ad967f1eb138&test=json&l=zijocf-pa7)

But then I would have had to adapt your scripts

because then my structure would have been “go/fiber/v2” and “go/fiber/v3.”

waghanza:
What will be the impact having two versions of fiber in here @ReneWerner87  ?

I mean,  if this is for internal use (comparing speed of v2 and v3) it could be done just using git, comparing develop and this branch.

However, the main idea here is to reveal the power of fiber, no matter what version it is.

ReneWerner87:
Exactly, that's why we updated to version v3, so it should also work without comparing the versions.

waghanza:
Do you want to me to compare both version @ReneWerner87 ?

ReneWerner87:
No all good
We can just merge the update for the fiber bench.

PR #9152, opened 2026-03-07
refactor(js): process SIGINT and SIGTERM in clusters

nakrovati:
Rewrite the Bun cluster code using the Bun API and add `SIGINT `and `SIGTERM` handling to clusters - now stopping Docker containers with `ctrl+c`, via `docker stop {container_id}`, or the Docker GUI will work.


waghanza (approved):
Awesome 🎉 

Thanks for your multiple contributions @nakrovati

PR #9155, opened 2026-03-09
Update hyperlane

eastspire:


waghanza (approved):
Good catch 🎉 

Seems that hyperlane is in 18.2 in develop

PR #9216, opened 2026-03-26
[Python] Add FastPySGI v0.4

remittor:
**FastPySGI** - very fast WSGI/ASGI engine for Python WEB-frameworks

**FastPySGI** is a fork of [FastWSGI](https://github.com/jamesroberts/fastwsgi), which has 500 stars and hasn't been updated in a long time. So, I decided to create my own fork of this project, which I've made a lot of changes to.

Project homepage: https://github.com/remittor/fastpysgi

**FastPySGI** - C-based WHL module with using `libuv` and `llhttp`.

### DIFF
--- a/python/fastpysgi-asgi/server.py
+++ b/python/fastpysgi-asgi/server.py
@@ -35,7 +35,6 @@ if __name__ == "__main__":
     parser.add_option("-h", "--host", dest="host", default='0.0.0.0', type="string")
     parser.add_option("-p", "--port", dest="port", default=3000, type="int")
     parser.add_option("-w", "--workers", dest="workers", default=0, type="int")
-    parser.add_option("-v", "--verbose", dest="verbose", default=0, type="int")
     (opt, args) = parser.parse_args() 
 
     workers = opt.workers
@@ -43,6 +42,4 @@ if __name__ == "__main__":
         import multiprocessing
         workers = multiprocessing.cpu_count()
 
-    fastpysgi.server.backlog = 4096
-    fastpysgi.server.loop_timeout = 1
-    fastpysgi.run(app, opt.host, opt.port, loglevel = opt.verbose, workers = workers)
+    fastpysgi.run(app, opt.host, opt.port, workers = workers)
--- a/python/fastpysgi-asgi/server.py
+++ b/python/fastpysgi-asgi/server.py
@@ -1,0 +1,48 @@
+import fastpysgi
+
+async def app(scope, receive, send):
+    if scope['type'] != 'http':
+        return
+
+    path = scope['path']
+    req_method = scope.get('method', '')
+    headers = [ ]
+
+    if req_method == "GET":
+        if path == "/":
+            await send( { 'type': 'http.response.start', 'status': 200, 'headers': [ ] } )
+            await send( { 'type': 'http.response.body', 'body': b"", 'more_body': False } )
+            return
+
+        if path.startswith("/user/"):
+            value = path[6:]
+            await send( { 'type': 'http.response.start', 'status': 200, 'headers': [[ b'Content-Type', b'text/plain; charset=utf-8' ]] } )
+            await send( { 'type': 'http.response.body', 'body': value.encode(), 'more_body': False } )
+            return
+
+    if req_method == "POST":
+        if path == "/user":
+            await send( { 'type': 'http.response.start', 'status': 200, 'headers': [ ] } )
+            await send( { 'type': 'http.response.body', 'body': b"", 'more_body': False } )
+            return
+
+    await send( { 'type': 'http.response.start', 'status': 404, 'headers': [ ] } )
+    await send( { 'type': 'http.response.body', 'body': b"", 'more_body': False } )
+
+if __name__ == "__main__":
+    import optparse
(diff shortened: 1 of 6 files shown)


python/fastpysgi-asgi/server.py
```diff
@@ -0,0 +1,48 @@
...
+    parser.add_option("-w", "--workers", dest="workers", default=0, type="int")
+    parser.add_option("-v", "--verbose", dest="verbose", default=0, type="int")
+    (opt, args) = parser.parse_args() 
+
+    workers = opt.workers
+    if workers <= 0:
+        import multiprocessing
+        workers = multiprocessing.cpu_count()
+
+    fastpysgi.server.backlog = 4096
```
waghanza: why 4096  ?

python/fastpysgi-asgi/server.py
```diff
@@ -0,0 +1,48 @@
...
+    fastpysgi.server.loop_timeout = 1
```
waghanza: why 1 ?

python/fastpysgi-asgi/server.py
```diff
@@ -0,0 +1,48 @@
...
+    fastpysgi.run(app, opt.host, opt.port, loglevel = opt.verbose, workers = workers)
```
waghanza: are log disabled ?

I mean it could mess results

python/fastpysgi-wsgi/pyproject.toml
```diff
@@ -0,0 +1,9 @@
+[build-system]
+build-backend = "flit_core.buildapi"
+requires = ["flit_core"]
+
+[project]
+name='server'
+version = '1.0.0'
+description = "Simple benchmark implementation"
+dependencies = [ "fastpysgi>=0.4,<1.0" ]
```
waghanza:
```suggestion
dependencies = [ "fastpysgi>=0.4,<0.5" ]
```

there is a bot that make update PR

python/fastpysgi-wsgi/server.py
```diff
@@ -0,0 +1,41 @@
+import fastpysgi
+
+def app(env, start_response):
+    path = env["PATH_INFO"]
+    req_method = env.get('REQUEST_METHOD', '')
+    headers = [ ]
+
+    if req_method == "GET":
```
waghanza, line 8: no routing feature yet ?

python/fastpysgi-asgi/server.py
remittor: Recommended formula for calculation:
`backlog = min(workers * connections_per_worker * 2, somaxconn)`
I just took the average value that many servers use by default.

python/fastpysgi-asgi/server.py
remittor: The value 1 is used to prevent the CPU from slacking when there's no network load (1 ms latency).
If the default settings are important (they're set to 3 ms), I can delete this line.

python/fastpysgi-asgi/server.py
remittor: Well, we're doing performance testing here, so logging is disabled.
If you'd like, I can change it to level `LL_CRIT_ERROR` = 2.

python/fastpysgi-wsgi/server.py
remittor: There is no routing, since this is an engine for use by high-level WEB frameworks (Flask, Falcon, FastAPI, etc).
Example:
https://github.com/the-benchmarker/web-frameworks/blob/develop/python/asgineer/server.py

python/fastpysgi-asgi/server.py
remittor: Starting with Linux kernel 5.4, the default value of `net.core.somaxconn` was increased from 128 to 4096.

waghanza: could you make this more dynamic ? I mean having a clear formula in python code base

python/fastpysgi-asgi/server.py
waghanza: maybe worth to have command line param, this let us homogenize timeout between each frameworks

python/fastpysgi-asgi/server.py
waghanza: No worries, just a question

The idea is to have the same conf as the other frameworks

python/fastpysgi-asgi/server.py
remittor: This value should be configured by the server administrator depending on the environment and the tasks being performed.
Okay. It'll be easier for me to just delete this line of code.

remittor: Fixed (this line deleted)
Default value: 1024

python/fastpysgi-asgi/server.py
remittor: Fixed (this line deleted)
Default value: 3 ms

python/fastpysgi-asgi/server.py
remittor: Fixed (this option deleted)
Default value: LL_ERROR = 3

python/fastpysgi-wsgi/pyproject.toml
remittor: Fixed

waghanza (approved):

PR #9255, opened 2026-04-10
Implement memory sampling

Joannis:
I tested the benchmarks on a few projects, Hummingbird, Express and Bun. All report their memory with this PR and the numbers align with the expected.


waghanza:
Nicee work @Joannis .

Can you provide a simple (a json file or else) that contians some figures ?

PR #9330, opened 2026-05-10
Add MicroPie Python framework

patx:


python/micropie/pyproject.toml
```diff
@@ -0,0 +1,10 @@
+
+[build-system]
+build-backend = "flit_core.buildapi"
+requires = ["flit_core"]
+
+[project]
+name='server'
+version = '1.0.0'
+description = "Simple benchmark implementation"
+dependencies = ["micropie>=0.28"]
```
waghanza: We should make sure that we stick on 0.28

```suggestion
dependencies = ["micropie>=0.28,<0.29"]
```

:information_source: Their is a bot here making update on major when needed.

May I ping you when having trouble on micropie ?

python/micropie/config.yaml
waghanza, line 1: May e also gunicorn (could use asgi since 25.3)

waghanza (changes requested):

python/micropie/config.yaml
patx: @waghanza I added granian and gunicorn to the config under engines. Not sure if I need to add a language section as well for gunicorn, I see aiohttp has it, but most others do not...

waghanza: I'll do it

patx: I believe I added the needed override in `confing.yaml`.

waghanza: Yes, but actually only one engine (the first is displayed). In our case uvicorn will be used

patx: @waghanza MicroPie recommends `uvicorn` or `granian` so this is most likely not an issue. Unless you plan on changing the benchmark to include multiple engines per run? Granian sees better performance with MicroPie so I have made it first. Thanks alot!

waghanza (commented):
great job @patx

python/micropie/pyproject.toml
patx: @waghanza ping away!

PR #9390, opened 2026-05-30
[add] gogo benchmark

7elven:
## What changed

- Add a Go/gogo benchmark implementation for the required route contract:
  - `GET /` returns 200 with an empty body
  - `GET /user/:id` returns the route parameter
  - `POST /user` returns 200 with an empty body
- Register a `gogo` Go engine so the benchmark generator can select a framework-specific Dockerfile.
- Add a `go/gogo.Dockerfile` that enables cgo, installs native zlib build headers, builds with the `gogo` tag, and installs runtime libraries needed by the native binary.

## Validation

- `gofmt -w go/gogo/main.go`
- `git diff --cached --check`
- `ruby -ryaml -e 'ARGV.each { |f| YAML.safe_load(File.read(f)); puts "ok #{f}" }' go/config.yaml go/gogo/config.yaml`
- `CGO_ENABLED=1 go test -tags gogo ./...` in the gogo source checkout
- `CGO_ENABLED=1 go run -mod=mod -tags gogo .` from `go/gogo`, then checked:
  - `GET /` -> 200, empty body
  - `GET /user/0` -> 200, body `0`
  - `POST /user` -> 200, empty body

## Notes

I could not run `bundle exec rake config` or a Docker build locally because this machine only has macOS system Ruby 2.6 for the repo's Ruby 3+ rake code and does not have Docker installed. The PR is opened as a draft so CI can validate the generated Docker path before review.

### DIFF
--- a/go/config.yaml
+++ b/go/config.yaml
@@ -6,6 +6,8 @@ language:
       command: /go/bin/app
     fasthttp:
       command: /go/bin/app
+    gogo:
+      command: /go/bin/app
 
 framework:
   engines:
--- a/go/gogo.Dockerfile
+++ b/go/gogo.Dockerfile
@@ -1,0 +1,73 @@
+{{#language.version}}
+  FROM golang:{{{.}}} AS build
+{{/language.version}}
+{{^language.version}}
+  FROM golang:1.26 AS build
+{{/language.version}}
+
+WORKDIR /go/src/app
+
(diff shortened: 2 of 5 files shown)


7elven:
url: https://github.com/Snocko-main/gogo

image

waghanza (approved):

PR #9416, opened 2026-06-03
Add trillium.rs

jbr:
I wasn't sure if "tuned" profiles are appropriate so I used two stock configurations. Let me know if a `trillium-tuned` would be an appropriate third configuration, using reuseport and appropriate configuration parameters for the workload

More info on the framework: https://trillium.rs

### DIFF
--- a/rust/trillium-smol/Cargo.toml
+++ b/rust/trillium-smol/Cargo.toml
@@ -1,0 +1,21 @@
+[package]
+name = "server"
+version = "0.0.0"
+edition = "2024"
+publish = false
+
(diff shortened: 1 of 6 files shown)


waghanza:
Thanks @jbr for this. The idea of this project is to expose how framework will react is **production**, so any implementation here should be **production-ready**, I mean with all concerns (security, reliability ...).

Maybe `-tuned` is not accurate for this purpose.

I mean the question is : should I take X or Y to create / migrate my app ? 

Not how can I optimize an existing app

PR #9484, opened 2026-06-21
Breeze Framework Added

nelthaarion:


go/breeze/config.yaml
```diff
@@ -0,0 +1,6 @@
+framework:
+  website: nelthaarion.github.io/breeze
+  version: 0.1.3
```
waghanza:
```suggestion
  version: 0.1
```

go/breeze/config.yaml
```diff
@@ -0,0 +1,6 @@
...
+
+  engines:
+    - GNET
```
waghanza, line 6: need to install gnet on container

waghanza (changes requested):

go/breeze/config.yaml
nelthaarion: could you please guide me what should I do ?

nelthaarion: Im pretty sure it will install along side go get github.com/nelthaarion/breeze

PR #9507, opened 2026-06-25
chore(v): pin V and vanilla commits

enghitalo:


v/fasthttp/config.yaml
```diff
@@ -1,5 +1,6 @@
 framework:
   github: vlang/v
+  version: 24120e7a80d798a1625a08d5cc02b96e17148bdb
```
waghanza, line 3: why pinning @ https://github.com/vlang/v/commit/24120e7a80d798a1625a08d5cc02b96e17148bdb @enghitalo ?

enghitalo: GC fixes. Have one more fix actualy. I'll turn it draft for now

waghanza: I'll merge to update results @enghitalo

PR #9580, opened 2026-07-15
Fix FastAPI benchmark: match baseline routing to fastapi-startkit

tmgbedu:
## Problem

The raw-FastAPI baseline (`python/fastapi/server.py`) registered routes with
direct `@app.get`/`@app.post` decorators, while the `fastapi-startkit` app
registers the same three routes via `APIRouter` + `app.include_router(...)`.
On FastAPI 0.139, `include_router` keeps a nested `_IncludedRouter` node on
`app.router.routes`, adding a per-request resolution layer that flat
`add_api_route` registration doesn't pay. So the benchmark was measuring
`include_router`'s routing cost, not `fastapi-startkit`'s framework overhead.

## Fix

Register the baseline's three routes on an `APIRouter`, mounted via
`app.include_router(router)` — the only remaining difference between the two
apps is now the framework itself.

Route behavior is unchanged and verified:
- `GET /` → `200`, empty body
- `GET /user/123` → `200`, body `123`
- `POST /user` → `200`, empty body

## Benchmark: before vs after

Method: Docker `python:3.14-slim`, uvicorn `--workers=$(nproc)` (11 workers),
`oha` 1.14 with keepalive + `--latency-correction`, 8s duration × 4 reps,
concurrency 64 and 256 — run twice per cell (once target-sequential, once
target-interleaved per rep, to average out host-load drift), 8 reps/cell
total. Matched stack both sides: FastAPI 0.139.0 / Starlette 1.3.1
(fastapi-startkit pinned to 0.47.0).

Relative delta (fastapi-startkit vs fastapi baseline; negative = startkit slower):

| Route | Δ @ c64 before → after | Δ @ c256 before → after |
|---|---|---|
| `GET /` | −5.8% → **−0.4%** | −9.3% → **+0.3%** |
| `GET /user/0` | −11.7% → **−1.0%** | −6.3% → **+8.4%** |
| `POST /user` | −9.8% → **+1.4%** | −10.9% → **−1.8%** |

**Average delta across all 6 cells: −8.9% before → +1.1% after.** The
framework-level overhead collapses to noise once both sides register routes
identically — confirming the prior gap was routing-mechanism cost
(`include_router`), not `fastapi-startkit` overhead.

Caveat: same-host macOS/OrbStack comparison, not official/authoritative
numbers (those require the Linux `run.sh` harness); host was under mixed
background load during measurement, hence the double run per cell.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,8 +16,10 @@ FRAMEWORKS.yaml
 */*/.Dockerfile*
 */*/.Makefile
 */*/ip.txt
+*/*/ip-*.txt
 venv
 */*/cid.txt
+*/*/cid-*.txt
 */.Makefile
 
 /lib/**/*.cr
@@ -62,3 +64,7 @@ __pycache__/
 .vscode
 .idea
 .zed
+
+# Claude Code local config
+.claude/
+.mcp.json
--- a/README.md
+++ b/README.md
@@ -666,16 +666,17 @@ artifacts):
 
 | | **fastapi** | **fastapi-startkit** |
 |---|---|---|
-| Declared dependency | `fastapi>=0.139,<0.140` | `fastapi-startkit[fastapi]==0.46.0` (pinned in PR #4) |
-| Underlying FastAPI | direct | `fastapi[standard]>=0.124.4,<0.125.0` (transitive) |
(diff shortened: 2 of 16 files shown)


tmgbedu:
Opened against the wrong repo by mistake (gh defaulted to upstream instead of our fork). Closing — correct PR is against fastapi-startkit/web-frameworks.

PR #9583, opened 2026-07-15
Fix bundler version mismatch blocking rspec on old system Ruby

tmgbedu:
## Problem

Running the local benchmark harness:

```
docker run -td python.fastapi.uvicorn
ENGINE=uvicorn LANGUAGE=python FRAMEWORK=fastapi bundle exec rspec .spec
```

fails before a single spec executes:

```
Could not find 'bundler' (4.0.3) required by your Gemfile.lock. (Gem::GemNotFoundException)
```

## Root cause

`Gemfile.lock` (gitignored, generated locally) pins `BUNDLED WITH 4.0.3`,
and its locked gems (e.g. `activesupport 8.1.3`) require **Ruby >= 3.2**.
This matches exactly what CI installs — `.github/workflows/ci.yml` sets up
`ruby-version: "4.0.5"` via `ruby/setup-ruby`. CI is not affected by this at all.

The failure only reproduces on machines whose only available Ruby is an old
system Ruby (e.g. macOS's bundled Ruby 2.6.10). Confirmed bundler 4.0.3
genuinely cannot install on Ruby 2.6 regardless of permissions:

```
bundler requires Ruby version >= 3.2.0. The current ruby version is 2.6.10.210.
```

and none of the exact locked gem versions install on Ruby 2.6 either — so
downgrading `Gemfile.lock`'s pin would diverge from what CI actually runs and
mask real incompatibilities, not fix anything.

## Fix

Since the correct fix is "use a Ruby that satisfies what's already pinned",
not "change the pins to fit an outdated Ruby":

- Add `ruby '>= 3.2'` to the `Gemfile`, so bundler surfaces the real
  requirement directly (a clear, actionable message) instead of the
  confusing "could not find bundler" error.
- Add `.ruby-version` pinned to `4.0.5`, matching CI, so Ruby version
  managers (rbenv/rvm/asdf/chruby) pick the right Ruby automatically.

No change to `Gemfile.lock` itself (it's gitignored/locally generated and
already correct for CI).

## Verification

Installed Ruby 4.0.0 locally via Homebrew (ships bundler 4.0.3 as its
default gem — satisfies the new constraint), then ran the full harness
end-to-end:

```
$ ENGINE=uvicorn LANGUAGE=python FRAMEWORK=fastapi bundle exec rspec .spec
......
6 examples, 0 failures
```

Also sanity-checked bundler resolution against a second framework
(python/flask + gunicorn) to confirm the fix isn't fastapi-specific:

```
$ ENGINE=gunicorn LANGUAGE=python FRAMEWORK=flask bundle exec rspec .spec
......
6 examples, 0 failures
```

Both docker containers used for verification were stopped/removed after
the run.


tmgbedu:
Opened against the wrong repo by mistake — closing. Correct PR is against the fastapi-startkit fork.

PR #9600, opened 2026-07-26
Fix README.md

sheosi:
During commit 200ba3284d4643697b0eda5929dbfab09f10c7a4 README.md got overrode by the breeze README. Let's fix that


waghanza:
Thanks, the REAMDE will be rewritten next time (apart of it was overwritten, it very outdated)

PR #9601, opened 2026-07-26
Update to zig 0.16

sheosi:
Updates zig to version 0.16 and to latest version of HTTPZ.

### DIFF
--- a/zig/httpz/src/main.zig
+++ b/zig/httpz/src/main.zig
@@ -2,8 +2,8 @@ const std = @import("std");
 const build_mode = @import("builtin").mode;
 const httpz = @import("httpz");
 
-pub fn main() !void {
-    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+pub fn main(init: std.process.Init) !void {
+    var gpa = std.heap.DebugAllocator(.{}){};
 
     defer {
         if (build_mode == .Debug) _ = gpa.deinit();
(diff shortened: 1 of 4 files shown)


zig/httpz/src/main.zig
```diff
@@ -2,8 +2,8 @@ const std = @import("std");
 const build_mode = @import("builtin").mode;
 const httpz = @import("httpz");
 
-pub fn main() !void {
-    var gpa = std.heap.GeneralPurposeAllocator(.{}){};
+pub fn main(init: std.process.Init) !void {
+    var gpa = std.heap.DebugAllocator(.{}){};
```
waghanza, line 6: DebugAllocator ? Not sur this is suitable for production

waghanza (changes requested):
missing config.yaml

zig/httpz/src/main.zig
sheosi: In 0.16 allocators have changed, GeneralPurposeAllocator is now DebugAllocator. Regardless, it wasn't being used in production, instead, smp_allocator was, and that hasn't changed:

```
    const allocator = if (build_mode == .Debug) gpa.allocator() else std.heap.smp_allocator;
```

sheosi: I'm updating config.yaml's zig version. Is that enough or there's more that needs to be added? httpz had already a config.yaml (since it already existed), it's not showing in files changed since it needed no change, but let me know if a change was needed.

waghanza (approved):

waghanza (approved):

Issue #9467, opened 2026-06-16
vanilla_io_uring fails CI: io_uring is blocked by Docker's default seccomp on GitHub Actions


enghitalo:
## Summary

The V `vanilla_io_uring` framework cannot pass the GitHub Actions CI smoke-test because **io_uring is blocked by Docker's default seccomp profile** on the runners. It builds and starts, but `io_uring_setup`/`io_uring_enter` are denied, so it never opens a ring and never becomes HTTP-ready.

Failing run: https://github.com/the-benchmarker/web-frameworks/actions/runs/27570918190/job/81506991936 (PR #9466)

```
Container did not become ready in 120 seconds!
Failed to initialize io_uring for worker 0:
  io_uring_queue_init_params failed for all ring entry/flag combinations
```

In the same run, **`vanilla_epoll`, `pico`, `fasthttp` and `veb` all pass** with the same V toolchain — only the io_uring variant fails, at the `docker run` readiness step (not at build). Since ~Docker 20.10 the default seccomp profile blocks the `io_uring_*` syscalls (io_uring submissions can bypass per-syscall seccomp filtering, so it's denied by default).

## Action taken

**Removed the `vanilla_io_uring` framework** from this repo (PR #9466). Rather than carry a framework that can't pass the smoke-test (or paper over it with a per-framework CI skip), it's dropped here. `vanilla_epoll` stays, and the io_uring engine continues to be benchmarked in [HttpArena](https://github.com/MDA2AV/HttpArena), whose harness permits io_uring.

## If io_uring support is wanted here later — the *fair* way

Running just the io_uring container with `--security-opt seccomp=unconfined` would make it pass, but it's unfair: that container would run **less sandboxed** than its peers (seccomp adds a small per-syscall cost), so its numbers wouldn't be comparable, and `unconfined` drops the whole sandbox rather than just io_uring.

The fair approach is a **single custom seccomp profile applied uniformly to every framework**: take Docker's default profile, add `io_uring_setup`, `io_uring_enter`, `io_uring_register` to the allow-list, and pass `--security-opt seccomp=<profile>.json` on *every* `docker run`.

- Frameworks that don't use io_uring are **bit-for-bit unaffected** (the added entries are never hit) — no perf change, no re-baseline.
- io_uring frameworks can initialize their rings.
- Everyone runs under the identical filter, so comparability is preserved and only the legitimate I/O-model difference remains.

[...]

PR #9609, opened 2026-07-31
Add zzz a Zig based framework

sheosi:
This adds zzz (a zig web server ) to the testing.


waghanza:
@mookums @bernardassan we want to introduce zzz here, is it something you are OK with ?

bernardassan:
@sheosi any reason why use are using the `epoll` backend instead of `io_uring`? Also, using `init.gpa` and compiling in release fast mode `.optimize = .ReleaseFast`, or directly setting up the release allocator `std.heap.smp_allocator`, should be preferred over using `std.heap.c_allocator` now after https://ziglang.org/devlog/2025/#2025-02-07

Edit: also, any particular reason for the `signal_handler` as it just panics, and even if it's fixed it doesn't seem that useful?
```zig
fn signal_handler(sig: std.os.linux.SIG) callconv(.c) void {
    const io = std.Options.debug_io;
    var buf: [128]u8 = undefined;
    var stderr = std.Io.File.stderr().writer(io, &buf);
    const stderr_w = &stderr.interface;
    stderr_w.print("signal {t}!", .{sig}) catch unreachable;
    stderr_w.flush() catch unreachable;
    std.process.exit(1);
}
```

bernardassan:
@waghanza, that's okay if @sheosi or someone else would be willing to update/maintain this, as Zzz is rapidly improving and there would be some breaking changes before 1.0.

waghanza:
Unfortunatly @bernardassan `io_uring` could not be tested because of security issues on docker, the same issue was detcted with ` v`

edit : see https://github.com/the-benchmarker/web-frameworks/actions/runs/27570918190/job/81506991936#step:8:34

bernardassan:
If that's a hard requirement then no problem but `--security-opt seccomp=unconfined` can be passed to allow it https://docs.docker.com/engine/security/seccomp/#run-without-the-default-seccomp-profile

waghanza:
Maybe not see easy to take, see https://github.com/the-benchmarker/web-frameworks/issues/9467

@sheosi is it ok for you to maintain? (I'll notify you on update of zzz)

data.json
waghanza, line 1: This file should not be modified.

I'll be auto-update on next round+ it trigger all CI jobs (all frameworks)

zig/zzz/config.yaml
```diff
@@ -0,0 +1,3 @@
+framework:
+  github: tardy-org/zzz
+  version: 0.3.2
```
waghanza, line 3:
```suggestion
  version: 0.3
```

waghanza (changes requested):

sheosi:
@bernardassan, I didn't quite get what you mean about init.gpa and .optimize = releaseFast. About c_allocator, ok I'll change it to smp_allocator. Unfortunately I can't make the changes right now, it'll have to be in a couple days.

And yes, I don't mind mainting this if @waghanza notifies me.

bernardassan:
Zig uses different allocators in `Debug` and `Release` mode, so `init.gpa` in `Debug` mode is `std.heap.SafeAllocator` while `Release` mode is `std.heap.SmpAllocator` (fast multi-threaded allocator). setting `.optimize = .ReleaseFast` will compile in `ReleaseFast` mode, leading to the release allocator `std.heap.smp_allocator` been used

sheosi:
@bernardassan Thank you, I couldn't find init.gpa anywhere (it turns out it exist but it's a little bit hidden). I'll make use of the init.gpa with releasefast as I think it's the most standard way.

Btw, the signal handler there was for containers. In my experience when running compiled programs (in my case Rust-based ones) inside Podman (and I believe Docker has the same behaviour) when receiving a sigkill no one catches that and the program keeps on running until being forcibly shutdown by the container manager a few seconds later. Some containers use some intermediaries but I like having a minimal ammount of executables. If this is a problem in any way I can test whether the same happens under Docker and Zig ( though Zig should make no difference ) and remove it.

zig/zzz/config.yaml
sheosi: Here, I think it would be ok to have the patch version, as 0.3.0 doesn't exist per se (it's 0.3.0-rc1), and 0.3.2 targets Zig 0.16 while 0.3.0-rc1 targets 0.13 and at this early point in the language those jumps introduce a lot of changes ( Zig 0.15 brought in a new async interface).

sheosi:
Btw, hard coding `.optimize = .ReleaseFast` makes the compiler not recognize `-Doptimize`, and that is used for `httpz`, I'm thinking about leaving it as it is as it isn't an important change. We can modify both of them later in another PR.

bernardassan:
No other change would be required since there is https://github.com/sheosi/web-frameworks/blob/86fe7f8414743d0b1e501b00a0f5237fb9772996/zig/Dockerfile#L35 so `init.gpa` works fine.

zig/zzz/config.yaml
waghanza: This information is just shown on UI, the idea is to be alogned with other framework)

Zig specifics is not handled here

waghanza (approved):

PR #9614, opened 2026-08-03
Add fulmine (JavaScript)

nigrosimone:
Adds **fulmine** ([fulmine.js](https://www.npmjs.com/package/fulmine.js)), an Express 5 compatible framework built on µWebSockets.js

The entry mirrors the `javascript/express` one: same three routes, `etag` off, `cluster.mjs` for multi-core, engine declared as `uwebsockets` in `config.yaml`. The dependency is pinned `^5.0.0` so rebuilds track the latest 5.x.

Verified locally against the published package: `GET /` answers 200 with an empty body, `GET /user/42` answers `42`, `POST /user` answers 200 with an empty body.

### DIFF
--- a/javascript/fulmine/app.mjs
+++ b/javascript/fulmine/app.mjs
@@ -1,0 +1,19 @@
+import express from 'fulmine.js';
+
+const app = express();
+
+app.set('etag', false);
+
+app.get('/', function (req, res) {
+  res.send('');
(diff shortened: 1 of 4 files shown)


waghanza (approved):
Good job 🎉 

I'll merge it soon to update results

PR #9627, opened 2026-08-06
Seems moro engine wasnt used in last run

M-Chris:
We have the fastest http engine available its a custom c++ based into a pure ABI for node. Seems it didn't get in based on the results last posted. This should resolve it I hope, although would be good to know how things are being ran or if we need to adjust something in the package build.

### DIFF
--- a/javascript/morojs/app.mjs
+++ b/javascript/morojs/app.mjs
@@ -20,7 +20,7 @@ const app = await createApp({
   // Minimal middleware for fair comparison
   performance: {
     clustering: {
-      enabled: true, // unleash the power of clustering to really see the power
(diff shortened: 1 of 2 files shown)


M-Chris:
@waghanza lmk if there is something particular keeping the native engine from working in the environment. 
I don't expect there is but also curious if the install of packages maybe pins instead of pulls the latest per the ^1.8.0 
Thanks! Quite confident we have the fastest node engine available per wide benchmark testing on our own repo https://github.com/Moro-JS/benchmark

Would be great to see the right numbers posted sooner than later too idk if your setup allow for one offs per package update or whatnot.. haven't dug too deep into the repo tbh  :)

waghanza (approved):

PR #9644, opened 2026-08-13
fulmine: use the framework's own cluster option

nigrosimone:
fulmine 5.9.0 has a `cluster` option, so the entry does not need `cluster.mjs` any more:
`express({ cluster: 'auto' })` forks one worker per usable core and each worker binds port 3000
with uWebSockets.js's shared flag, SO_REUSEPORT, which is what the shared `cluster.mjs` was doing
with `cluster.fork()`. The worker count reads the cgroup quota before the machine, so a container
gets its own cores.

The engine's command is overridden to `node app.mjs`, since there is no wrapper to run.

Ran here against 5.9.0: 8 workers on an 8-core machine, and the three routes answer as before.


waghanza:
You need to change the command in `config.yaml` @nigrosimone

nigrosimone:
Mhh what Is the right command?

waghanza:
I made a corrective commit @nigrosimone

nigrosimone:
Thank you!

PR #9677, opened 2026-08-25
Add cps-http

gabearro:
Adds [cps-http](https://github.com/gabearro/cps-http), an HTTP/1.1, HTTP/2, and HTTP/3 implementation built on the CPS Nim runtime.

The benchmark adapter exercises the full HTTP/1.1 parser and router. Since each CPS event loop is deliberately single-threaded, the container runs one `SO_REUSEPORT` worker per visible CPU, with `CPS_WORKERS` available for CPU-quota-aware deployments.

### DIFF
--- a/nim/cps-http/config.yaml
+++ b/nim/cps-http/config.yaml
@@ -1,0 +1,31 @@
+framework:
+  github: gabearro/cps-http
+  version: 1.0.0
+  engines:
+    - multiprocess
+
+# The CPS event loop is deliberately single-threaded. Run one independent,
+# SO_REUSEPORT-backed worker per visible CPU, as production process managers do.
+# CPS_WORKERS remains available for CPU-quota-aware benchmark environments.
+language:
+  engines:
+    multiprocess:
+      command:
+        - /bin/sh
+        - -c
+        - 'workers="${CPS_WORKERS:-$(getconf _NPROCESSORS_ONLN)}"; pids=""; trap ''kill $pids 2>/dev/null; wait'' TERM INT; i=0; while [ "$i" -lt "$workers" ]; do /usr/src/app/server & pids="$pids $!"; i=$((i + 1)); done; wait'
+
+# The suite's Nim image already adds release mode, speed optimization, and LTO.
+# Danger mode removes runtime checks from the hot path, ARC avoids tracing-GC
+# work, and native CPU tuning lets the compiler use the benchmark host ISA.
+build_opts:
+  - "-d:danger"
+  - "--mm:arc"
+  - "--passC:-march=native"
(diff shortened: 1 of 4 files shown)


gabearro:
Lmk if anything needs to change for this to get merged, thanks!

nim/cps-http/config.yaml
```diff
@@ -0,0 +1,31 @@
+framework:
+  github: gabearro/cps-http
+  version: 1.0.0
```
waghanza:
```suggestion
  version: 1.0
```

nim/cps-http/config.yaml
```diff
@@ -0,0 +1,31 @@
...
+  engines:
+    - multiprocess
```
waghanza: do you have some documentation around this engine ?

nim/cps-http/config.yaml
```diff
@@ -0,0 +1,31 @@
...
+# The suite's Nim image already adds release mode, speed optimization, and LTO.
+# Danger mode removes runtime checks from the hot path, ARC avoids tracing-GC
+# work, and native CPU tuning lets the compiler use the benchmark host ISA.
+build_opts:
+  - "-d:danger"
+  - "--mm:arc"
+  - "--passC:-march=native"
+  - "--passC:-mtune=native"
+
+build_with:
```
waghanza, line 25: please you a versionning system to include this project here

waghanza (changes requested):

nim/cps-http/config.yaml
gabearro: Got it, will make the appropriate changes!

nim/cps-http/config.yaml
gabearro: Updated the framework version to 1.0 in 2a02ccd3.

nim/cps-http/config.yaml
gabearro: Added nim/cps-http/README.md documenting the engine and changed the adapter to the shard-local multithreaded reactor: one owned event loop, router, listener, and connection set per thread, with SO_REUSEPORT accept distribution.

nim/cps-http/config.yaml
gabearro: Done. I published SemVer release tags: cps-runtime v1.1.0; cps-tls, cps-quic, and cps-http v1.0.1. The benchmark now installs those immutable tags instead of commit SHAs.

waghanza: better to set that in  `nim/cps-http/server.nimble` @gabearro

waghanza (approved):

PR #9716, opened 2026-08-30
fulmine: drop the connection headers

nigrosimone:
fulmine answers every response with `Connection: keep-alive` and `Keep-Alive: timeout=10` because Express does. On HTTP/1.1 keep-alive is implicit, so the two headers add about 50 bytes and two header writes per response without telling the client anything new. The framework has a setting to turn them off, this PR sets it, like `etag` already is.

`Connection: close` is still sent when the client asks for it, verified on the three routes.

### DIFF
--- a/javascript/fulmine/app.mjs
+++ b/javascript/fulmine/app.mjs
@@ -7,6 +7,9 @@ import express from 'fulmine.js';
 const app = express({ cluster: 'auto' });
 
 app.set('etag', false);
+// keep-alive is implicit on HTTP/1.1, so the Connection and Keep-Alive headers carry no
+// information: off, which is also what the engine itself answers.
+app.set('connection headers', false);
 
 app.get('/', function (req, res) {
   res.send('');


waghanza:
Thanks for notice about this behavior in express.  Maybe we should disable this behavior is express also, and alike

However if we disable keepalive, in load test, not sur this has a impact

nigrosimone:
To be precise, this does not disable keep-alive: the connection stays persistent, only the two advertisement headers go, and HTTP/1.1 is persistent by default without them. So it is bytes per response, not behavior.
On express you cannot fully match it: node always writes Connection: keep-alive, and Keep-Alive: timeout=5 only goes away with keepAliveTimeout = 0, which also removes the idle timeout, so it is a behavior change and not the same thing.
I have no measurement on your machine, and with the old oha --disable-keepalive run it would have changed nothing at all since every connection closed anyway. With zrk --closed the headers do go out on every response. If you prefer all entries to look the same, close this and I am fine with it.

waghanza (approved):

waghanza (approved):

Issue #9715, opened 2026-08-30
The links to the frameworks on the website are all incorrect


nigrosimone:
see https://web-frameworks-benchmark.vercel.app/result?l=javascript&f=

some example:
- fulmine - https://foxify.js.org/
- uwebsockets - https://totaljs.com/
- ultimate-express - https://totaljs.com/

waghanza:
yep, I have to make something, it's in my roadmap

feel free to investigate thos

PR #9718, opened 2026-08-30
Fix the website and version fields carrying over from the framework merged before

nigrosimone:
`Hash#recursive_merge` in `.tasks/helpers.rb` uses `merge!`, so it modifies the receiver. `.tasks/db.rake` loads `main_config` once and merges it with every framework in the loop, so `main_config` keeps every key it has ever seen and a framework whose `config.yaml` does not set a key gets the value of the framework merged before it, in alphabetical order.

In the published `data.min.json` this makes **230 of 358 `website` fields wrong**: 64 point at another framework's site (`clojure/donkey` at coast.swlkr.com, `cpp/oatpp` at luminusweb.com, `javascript/fulmine` at foxify.js.org), and 203 more are `http://` instead of `https://` because one `unsecure:` framework poisons all the rest. `version` leaks too: `zig/httpz` and `v/vanilla_epoll` publish `24120e7a...`, which is the V compiler pinned by `v/veb`.

The fix is `merge` instead of `merge!`. Two frameworks then had no version of their own, so I added the one each of them already pins: `52eb187c...` from `zig/httpz/build.zig.zon`, `1822cbb4...` from the `download` step of the two `v/vanilla_*`. Change them if you would rather have something else there.

Checked with ruby 3.4 in a container: `rake config` writes the same 847 Dockerfiles and Makefiles before and after, so nothing about the builds moves, and the three added `version:` keys do not change a manifest either. `.spec/helpers_spec.rb` is new, it fails on two of three examples without the fix.

fix https://github.com/the-benchmarker/web-frameworks/issues/9715

### DIFF
--- a/.spec/helpers_spec.rb
+++ b/.spec/helpers_spec.rb
@@ -1,0 +1,30 @@
+require_relative '../.tasks/helpers'
+
+RSpec.describe Hash do
+  describe '#recursive_merge' do
+    it 'leaves the receiver alone' do
+      main = { 'framework' => { 'version' => 1 } }
+      main.recursive_merge('framework' => { 'website' => 'first.example' })
+
+      expect(main).to eq('framework' => { 'version' => 1 })
+    end
+
+    # .tasks/db.rake merges into the same main config once per framework, so a
+    # value read here has to come from the hash passed in and from nothing else.
+    it 'does not carry a value over to the next merge' do
+      main = {}
+      first = main.recursive_merge('framework' => { 'website' => 'first.example' })
+      second = main.recursive_merge('framework' => { 'version' => 2 })
+
+      expect(first.dig('framework', 'website')).to eq('first.example')
+      expect(second.dig('framework', 'website')).to be_nil
+    end
+
+    it 'merges nested hashes' do
+      config = { 'framework' => { 'version' => 1, 'engines' => { 'node' => 'a' } } }
+
+      expect(config.recursive_merge('framework' => { 'engines' => { 'node' => 'b', 'bun' => 'c' } }))
+        .to eq('framework' => { 'version' => 1, 'engines' => { 'node' => 'b', 'bun' => 'c' } })
+    end
+  end
+end
--- a/.tasks/helpers.rb
+++ b/.tasks/helpers.rb
@@ -8,7 +8,11 @@ def normalize_shell(shell)
 end
 
 class Hash
+  # Returns a new hash and leaves the receiver alone. With `merge!` the receiver
+  # kept every key it was ever merged with, so in .tasks/db.rake, where the same
+  # main config is merged once per framework, a framework whose config.yaml does
+  # not set a key got the value of the framework merged before it.
   def recursive_merge(hash)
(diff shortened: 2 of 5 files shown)


.spec/helpers_spec.rb
waghanza, line 1: 👍🏻

.tasks/helpers.rb
```diff
@@ -8,7 +8,11 @@ def normalize_shell(shell)
 end
 
 class Hash
+  # Returns a new hash and leaves the receiver alone. With `merge!` the receiver
+  # kept every key it was ever merged with, so in .tasks/db.rake, where the same
+  # main config is merged once per framework, a framework whose config.yaml does
+  # not set a key got the value of the framework merged before it.
   def recursive_merge(hash)
-    merge!(hash) { |_, old, new| old.instance_of?(Hash) ? old.recursive_merge(new) : new }
+    merge(hash) { |_, old, new| old.instance_of?(Hash) ? old.recursive_merge(new) : new }
```
waghanza, line 16: 🎉

v/vanilla_epoll/config.yaml
```diff
@@ -1,5 +1,6 @@
 framework:
   github: vlang/v
+  version: 1822cbb4b5316f433df4805bb68560fadd4df385
```
waghanza, line 3: why ?

v/vanilla_io_uring/config.yaml
```diff
@@ -1,5 +1,6 @@
 framework:
   github: vlang/v
+  version: 1822cbb4b5316f433df4805bb68560fadd4df385
```
waghanza, line 3: why ?

zig/httpz/config.yaml
```diff
@@ -1,2 +1,3 @@
 framework:
   github: karlseguin/http.zig
+  version: 52eb187cff75e168316b44480cd9f0fae26a1c5e
```
waghanza, line 3: why ?

nigrosimone: Because this one has no version of its own. Before the fix it was showing a version carried over from the framework merged before it, and with the carry-over gone the field would be empty. So I put here the commit http.zig already pins in zig/httpz/build.zig.zon, 52eb187cff75e168316b44480cd9f0fae26a1c5e. Happy to drop the line if you prefer no version at all.

v/vanilla_epoll/config.yaml
nigrosimone: Same reason as zig/httpz: these two had no version of their own and were showing one carried over from another framework. 1822cbb4b5316f433df4805bb68560fadd4df385 is the V commit the download step of this framework already pins, so it is the same thing that gets built.

v/vanilla_io_uring/config.yaml
nigrosimone: Same as v/vanilla_epoll: no version of its own, and 1822cbb4b5316f433df4805bb68560fadd4df385 is the V commit its own download step pins.

waghanza (approved):

PR #9723, opened 2026-08-30
Add jet_server (Dart)

Wdestroier:


dart/jet_server/pubspec.yaml
```diff
@@ -0,0 +1,9 @@
+name: jet_server_api
+
+environment:
+  sdk: "^3.13.0"
+
+dependencies:
+  jet_server:
+    git:
+      url: https://github.com/Wdestroier/jet_server.git
```
waghanza: Jet server could be installed via a public package ?

Wdestroier: I didn't publish it on pub.dev, that's why I added a git reference.

Wdestroier: I published the package and adjusted the dependency.

waghanza (approved):

PR #9745, opened 2026-09-05
Add cwist

gg582:
Adds [cwist](https://github.com/Religiya-Serdtsa/CWIST), a C17 web framework and application server with HTTP/1.1, HTTP/2, HTTP/3 (QUIC), and WebSocket support. I maintain the framework.

The implementation is the usual three routes from the contract (`GET /`, `GET /user/:id`, `POST /user`) on port 3000. The build clones the framework at tag v3.3 with its vendored dependencies (BoringSSL, lsquic, libttak, SQLite) and links statically, so the resulting container has no runtime requirements beyond libc.

One thing worth flagging: the vendored build (BoringSSL in particular) makes the image build slower than most C entries here. If that turns out to be a problem for CI, I can look at trimming the dependency set for this target.

### DIFF
--- a/c/cwist/config.yaml
+++ b/c/cwist/config.yaml
@@ -1,0 +1,39 @@
+# the-benchmarker/web-frameworks implementation config.
+# This file is maintained here and mirrored into the benchmark repository as
+# `c/cwist/config.yaml`. The matching application source is `main.c`.
+
+framework:
+  github: Religiya-Serdtsa/CWIST
+  version: 3.2
+
+  build_deps:
+    - git
+    - cmake
+    - pkg-config
+    - libzstd-dev
+    - libbrotli-dev
+    - libcurl4-openssl-dev
+    - libnghttp2-dev
+
+  download:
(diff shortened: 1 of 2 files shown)


gg582:
CI passed. Thanks

waghanza:
Weird I have a compilation issue locally
```
0.187 In file included from cwist/include/cwist/net/http/http.h:12,
0.187                  from cwist/include/cwist/sys/app/app.h:9,
0.187                  from cwist/include/cwist/app.h:14,
0.187                  from main.c:10:
0.187 cwist/include/cwist/core/db/sql.h:9:10: fatal error: sqlite3.h: No such file or directory
0.187     9 | #include <sqlite3.h>
0.187       |          ^~~~~~~~~~~
0.187 compilation terminated.
``` 
@gg582

gg582:
Got it. I'll check for this and force push it into v3.3 if it's reproduced

gg582:
Yes that's reproduced. idk how I didn't find this problem before. I'll be right back with a new pull request.

Sorry for your inconvenience(if it bothered you)

PR #9781, opened 2026-09-14
chores(cwist): Update CWIST version to v3.4.1

gg582:
- Bump CWIST version to v3.4.1


c/cwist/config.yaml
```diff
@@ -9,7 +9,7 @@ language:
 
 framework:
   github: c4punks/CWIST
-  version: 3.4
+  version: 3.4.1
```
waghanza:
```suggestion
  version: 3.4
```

Better to stock to X.Y format on UI

waghanza (approved):

c/cwist/config.yaml
gg582: I agree with your idea. In my project, X is a major release, Y is a minor release. Z is for minor refactor tasks and CVE. Basically it's 3.4

gg582:
Good, I hope current CWIST v.3.4.1 patches enhance tail latency. Theorically it should be, but it's always interesting to see that theories go weird

waghanza (approved):

waghanza:
Let's see

PR #9789, opened 2026-09-15
Perf improvements to the zzz benchmark

sheosi:
It was found that the ZZZ had some configuration problems, the first is that the number of connnections is too high (1024) when the maximum of the benchmark is 512 and second and most important is that it leaves log by default, outputting a lots of logs, now we are only showing errors.

This bumps it's performance to be on pair with the likes of rust axum.


waghanza:
well done @sheosi 

image

sheosi:
Thanks, those were really suggestions of @bernardassan that I implemented.

PR #9810, opened 2026-09-18
fulmine.js: answer with res.end

nigrosimone:
Switches the three handlers from `res.send` to `res.end`.

`res.send('')` writes `Content-Type: text/html; charset=utf-8` on an empty body, as Express does. `res.end` is node's and sends no Content-Type, so each response is about 40 bytes shorter, the same headers the raw uWebSockets.js entry sends.


waghanza:
Thanks for the clear explanation.

Could you check if any other JavaScript framework use send instead of end ?

nigrosimone:
> Thanks for the clear explanation.
> 
> Could you check if any other JavaScript framework use send instead of end ?

I checked all the javascript entries on master. send is used by express, express-bun, express-deno, fastify, fastify-bun, feathersjs, foxify, hyper-express, iotjs-express, restana, ultimate-express and tinyhttp. end is used by uwebsockets, sifrr, polka, restify, rayo, muneem, morojs, routejs and a few more. The rest return the body from the handler.

I would rather not touch the other entries. They are not mine, and their maintainers could say I changed their implementation to make it slower. Since there is always a way to do things better, that complaint is always valid.

One thing I know for sure, because fulmine is a wrapper on it: the uwebsockets entry could be much faster. It does not use DeclarativeResponse, which removes the javascript layer and lets the C++ answer. That is what fulmine does automatically behind the scenes, with uWebSockets.js it has to be written by hand. If you want I can tag its author here and he can decide.

Side note: I can't be certain that using `end` is faster for everyone, but it should be.

waghanza:
Ok, then thanks for the investiguation @nigrosimone .

I'll change the implementations of each framework and submit the maintainer for approval

waghanza:
for uwebsocket, you can optimise the implementation here @nigrosimone 

the author does not want to be involve in this project

nigrosimone:
> the author does not want to be involve in this project

Which is exactly the reason I would rather not touch his framework 😂

Joking aside, I'll pass on opening that PR myself. fulmine is a wrapper on uWebSockets.js, so submitting a change to his entry that we already know would not be welcome is the best way to spoil a relationship I depend on.

If you or anyone else wants to take it, this is what the implementation would look like with `DeclarativeResponse`. The routes are compiled at `listen()` and answered by uWS itself, no javascript runs per request:

```js
import { App, DeclarativeResponse } from 'uWebSockets.js';

// Create a new instance of uWebSockets.js HTTP
const port = 3000;
const app = App();

// GET "/" => 200 with empty body
app.get('/', new DeclarativeResponse().end(''));

// GET "/user/:id" => 200 with "id" as body
app.get('/user/:id', new DeclarativeResponse().writeParameterValue('id').end());

// POST "/user" => 200 with empty body
app.post('/user', new DeclarativeResponse().end(''));

// Start the server on port 3000
app.listen(port, () => {});
```

One caveat though. I compared the bytes on the wire with the current entry (20.71.0):

- `/` and `POST /user` come out byte identical, `Content-Length: 0`, same headers.
- `GET /user/:id` does not. The parameter has no known length when the instructions are built, so uWS frames it chunked: `Transfer-Encoding: chunked` and `2\r\n42\r\n0\r\n\r\n`, 113 bytes against the 94 of `res.end(req.getParameter(0))`.

So it removes the javascript layer on all three routes, but on `/user/:id` it pays chunked framing and 19 extra bytes for it. Keep-alive and pipelining behave the same, I checked with four requests on one connection. Whether that is a net win is for the CI to say, it may well be that only the two empty body routes are worth compiling.

PR #9813, opened 2026-09-20
python: peregrine 1.1.7

grepjava:
Bumps the `peregrine` engine to 1.1.7, released today and on PyPI as
[`peregrine-server` 1.1.7](https://pypi.org/project/peregrine-server/1.1.7/).

TLS moves to a BoringSSL copy vendored in the engine's protocol layer, rather
than the system OpenSSL. Measured on one worker serving static files, against
the same server built on OpenSSL, over seven rotated rounds:

| | throughput | CPU per GiB |
|---|---|---|
| HTTPS/1.1 | +23 to +31% | −19 to −23% |
| HTTP/2 | +35 to +45% | −26 to −31% |

A connection that handshakes once per request gains 75% at 44.6% less CPU a
request, which is the case this suite's `-c 64` keep-alive load does not
exercise, so the effect here will be smaller than those numbers.

Three lines:

- `python/config.yaml` — the engine pin, `>=1.1.5,<1.2` to `>=1.1.7,<1.2`
- `python/peregrine-asgi/config.yaml`, `python/peregrine-wsgi/config.yaml` —
  `version`, which is peregrine's own

`peregrine-django`, `-fastapi`, `-flask` and `-blacksheep` report the version
of the framework they serve rather than the server's, so they are unchanged.

No new entries and no Dockerfile change. Wheels cover cp311 to cp314 including
free-threaded 3.14t, on `manylinux_2_35`, so the `python:*-slim` image installs
without building.


waghanza (commented):
Not sure this PR is usefull 😛 

When the results are updated, peregrine is taken in its last version

python/peregrine-asgi/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: grepjava/peregrine
-  version: 1.1.5
+  version: 1.1.7
```
waghanza:
```suggestion
  version: 1.1
```

python/peregrine-wsgi/config.yaml
```diff
@@ -1,6 +1,6 @@
 framework:
   github: grepjava/peregrine
-  version: 1.1.5
+  version: 1.1.7
```
waghanza:
```suggestion
  version: 1.1
```

python/config.yaml
```diff
@@ -95,7 +95,7 @@ language:
           ${PYTHON_APP}
     peregrine:
       bootstrap:
-        - pip install 'peregrine-server>=1.1.5,<1.2'
+        - pip install 'peregrine-server>=1.1.7,<1.2'
```
waghanza:
```suggestion
        - pip install 'peregrine-server>=1.1,<1.2'
```

waghanza (commented):
Maybe be less restrictive, will allows less maintainance.

What do you think @grepjava  ?

PS : Version are always use at latest on each round (ofc in encapsed range)

PR #9814, opened 2026-09-20
Add Garuda (Swift)

grepjava:
Adds [Garuda](https://github.com/grepjava/garuda) 1.0.1 as a Swift entry.

Garuda is a Swift web framework with its own HTTP engine — no SwiftNIO, no Foundation. MIT licensed.

### The entry

`swift/garuda/` holds the three files the other Swift entries hold:

- `config.yaml` — repository, version, and `SERVER_HOSTNAME`/`SERVER_PORT`, the same shape as `swift/hummingbird-framework/config.yaml`.
- `Package.swift` — a `server` package with one executable target, depending on a published tag.
- `Sources/server/main.swift` — the three routes, with the environment read into the same argument parser the `garuda` command line uses, since the image starts the binary with no arguments. One worker per core.

`reviewers.yaml` gains `garuda: [grepjava]`.

Nothing else changes: `swift/Dockerfile` and `swift/config.yaml` already do what this entry needs, and the generated `.Dockerfile*`/`.Makefile` are gitignored.

### Checked

Built through the generated Dockerfile (`swift:6.3` build stage, `swift:6.3-slim` runtime) on Linux and driven with curl:

| | |
|---|---|
| `GET /` | 200, empty body |
| `GET /user/42` | 200, body `42` |
| `POST /user` | 200, empty body |

Image 442 MB against the 403 MB `swift:6.3-slim` base.

One thing I checked rather than assumed, in case it matters for the shared Swift image: Garuda links OpenSSL, where the other Swift entries reach TLS through a vendored BoringSSL and so would not prove it is there. `swift:6.3-slim` does carry `libssl.so.3`, `libcrypto.so.3` and `libz.so.1`, so the runtime stage needs no change. The build stage already installs `libssl-dev` and `zlib1g-dev`, which are its only build dependencies.

No benchmark figures are claimed here — CI builds and route-checks, and the numbers are yours to measure.


swift/garuda/config.yaml
```diff
@@ -0,0 +1,8 @@
+framework:
+  github: grepjava/garuda
+  version: 1.0.1
```
waghanza:
```suggestion
  version: 1.0
```

waghanza (changes requested):

waghanza (approved):

